PR #17, opened 2016-01-29
Improved SplitPackages

javiercn:
Addressed feedback from https://github.com/aspnet/DnxTools/pull/16

### DIFF
--- a/src/SplitPackages/Program.cs
+++ b/src/SplitPackages/Program.cs
@@ -5,26 +5,86 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Microsoft.Extensions.CommandLineUtils;
+using Microsoft.Extensions.Logging;
 
 namespace SplitPackages
 {
     class Program
     {
-        static void Main(string[] args)
+        private const int Ok = 0;
+        private const int Error = 1;
+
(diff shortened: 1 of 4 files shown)


src/SplitPackages/Program.cs
```diff
@@ -5,26 +5,86 @@
...
+                CommandOptionType.SingleValue);
+
+            var whatIfOption = app.Option(
+                "--whatif",
+                "Performs a dry run of the command with the given arguments without copying the packages",
+                CommandOptionType.NoValue);
+
+            var program = new Program(app, sourceOption, csvOption, destinationOption, whatIfOption);
+
+            app.OnExecute(new Func<int>(program.Execute));
```
Eilon: Wondering - can you just pass in the delegate and let the compiler do the work?

```c#
app.OnExecute(program.Execute);
```

Eilon:
Looks good :shipit:

src/SplitPackages/Program.cs
javiercn: The compiler is not smart enough to determine the right overload to use, so I need to rely on the new Func<...>

Eilon: Ah ok no worries.

PR #21, opened 2016-02-10
Improvements to split packages

javiercn:
* Fix bug in whatif option not being honored.
* Added support for generating project.json files that allow restoring from
  the destination folders.
* Added an option to treat warnings as errors and fail early.
* Added an option to print only warning and error messages only.

### DIFF
--- a/src/SplitPackages/Program.cs
+++ b/src/SplitPackages/Program.cs
@@ -5,11 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text.RegularExpressions;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using NuGet.Packaging;
 
 namespace SplitPackages
 {
@@ -17,19 +17,6 @@ namespace SplitPackages
     {
         private const int Ok = 0;
         private const int Error = 1;
-
-        private static readonly Regex DependencyRegex;
-
-        static Program()
-        {
-            //var moniker = @"\.v4\.(5|6)(.1)?";
-            var moniker = @"\.v\d(.\d){1,2}";
-            var nameExpr = @"(?<name>[a-z]+((\.|-)[a-z0-9]+)*?(" + moniker + ")?){1}";
-            var versionExpr = @"(?<version>\d+(\.\d+){2,3}(\-[a-z0-9]+)*)";
-
-            DependencyRegex = new Regex($"({nameExpr})\\.({versionExpr})(.nupkg)?",
-                RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
-        }
 
         private ILogger _logger;
 
@@ -169,10 +156,7 @@ namespace SplitPackages
 
             foreach (var path in packagesByFolder)
             {
-                var dependenciesList = path
-                    .Select(p => p.DestinationPath)
-                    .Select(Path.GetFileNameWithoutExtension)
-                    .Select(CreateDependency);
+                var dependenciesList = path.Select(CreateDependency);
 
                 var jsonFileBuilder = new ProjectJsonFileBuilder(
                     Path.Combine(path.Key, "project.json"),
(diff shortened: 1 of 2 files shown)


javiercn:
@pranavkm @muratg

muratg:
You need to add a frameworks section to the generated project.json files, otherwise dotnet restore will fail.

javiercn:
@muratg What failure will it produce? I've tried this and successfully ran dotnet restore against it. The only thing required is a dependencies node. The only failure that I observed was that it was unable to resolve some packages, but I don't think it's related to that.

muratg:
```
murats-mbp% dotnet restore
error: The project 'ship' does not specify any target frameworks in '/Users/mg/Pkgs/split/ship/project.json'.
log  : Committing restore...
log  : Restore failed in 89ms.

Errors in /Users/mg/Pkgs/split/ship/project.json
    The project 'ship' does not specify any target frameworks in '/Users/mg/Pkgs/split/ship/project.json'.
```

muratg:
You may be using an older dotnet cli build. My build is from last Friday.

src/SplitPackages/Program.cs
```diff
@@ -66,7 +86,24 @@ static int Main(string[] args)
                 "Performs a dry run of the command with the given arguments without copying the packages",
                 CommandOptionType.NoValue);
 
-            var program = new Program(app, sourceOption, csvOption, destinationOption, whatIfOption);
+            var warningsOnly = app.Option(
+                "--warningsonly",
```
pranavkm: `--quiet`?

src/SplitPackages/Program.cs
```diff
@@ -206,9 +303,69 @@ private IList<PackageItem> ExpandPackages(IList<PackageItem> packagesFromCsv, st
...
+        {
+            return packages.Any(p => p.Identity.Equals(packageName, StringComparison.OrdinalIgnoreCase));
+        }
+
+        private static IList<PackageItem> ExpandPackagesFromLiteral(
+            IList<PackageItem> packagesFromSource,
+            string packageName,
+            string destinationFolderName)
+        {
+            var packageFromSource = packagesFromSource
```
pranavkm: Would there ever be the case where `packageName` doesn't exist in the source? If so, we should probably log if it can't be found.

src/SplitPackages/Program.cs
```diff
@@ -206,9 +303,69 @@ private IList<PackageItem> ExpandPackages(IList<PackageItem> packagesFromCsv, st
...
+                .Single(p => p.Identity.Equals(packageName, StringComparison.OrdinalIgnoreCase));
+
+            return new List<PackageItem>
+            {
+                new PackageItem {
```
pranavkm: Formatting (brace on next line)

src/SplitPackages/Program.cs
```diff
@@ -206,9 +303,69 @@ private IList<PackageItem> ExpandPackages(IList<PackageItem> packagesFromCsv, st
...
+                    DestinationFolderName = destinationFolderName,
+                    Identity = packageFromSource.Identity,
+                    Version = packageFromSource.Version
+                }
+            };
+        }
+
+        private static IList<PackageItem> ExpandPackagesFromPattern(
+            string sourceFolder,
+            IList<PackageItem> packagesFromSource, string pattern, string destinationFolderName)
```
pranavkm: Params on separate lines.

src/SplitPackages/Program.cs
```diff
@@ -331,5 +505,98 @@ private class Arguments
...
+
+            public void AddFramework(Framework fx)
+            {
+                switch (fx)
+                {
+                    case Framework.Net451:
+                        _frameworks.Add("net451");
+                        break;
+                    default:
+                        break;
```
pranavkm: Is the plan to have more than one tfm here? Couldn't this be `AddFramework(string frameworkName)` instead?

pranavkm:
:watch: for questions.

src/SplitPackages/Program.cs
javiercn: good idea

src/SplitPackages/Program.cs
javiercn: We log it later on, when this method is called we have already checked that the package exists

src/SplitPackages/Program.cs
javiercn: Probably, I'm not sure if adding more TFMs will affect restore using this project.json so I would just leave it like this for now, it's not crazy complex.

javiercn:
@pranavkm updated

pranavkm:
:shipit:

PR #25, opened 2016-02-24
Add dnxcore50 support for SplitPackages

shirhatti:
@javiercn Can you have a look?

### DIFF
--- a/src/SplitPackages/project.json
+++ b/src/SplitPackages/project.json
@@ -16,6 +16,11 @@
         "NuGet.Client": "3.4.0-*"
     },
     "frameworks": {
+        "dnxcore50": {
+            "dependencies": {
+                "NETStandard.Library": "1.0.0-rc2-*"
+            }
+        },
         "net451": {
             "frameworkAssemblies": {
                 "Microsoft.CSharp": "",


javiercn:
@shirhatti Were you able to run the tool successfully on CoreCLR?

src/SplitPackages/project.json
```diff
@@ -16,6 +16,11 @@
         "NuGet.Client": "3.4.0-*"
     },
     "frameworks": {
+        "dnxcore50": {
+            "dependencies": {
+                "NETStandard.Library": "1.0.0-rc2-*"
```
javiercn: Just 1.0.0-*

javiercn:
:shipit: After that

src/SplitPackages/project.json
shirhatti: @troydai ?

javiercn: @pranavkm :smiley_cat:

troydai: We mostly spent time on `dev` and `rel/1.0.0` branch where we use `rc2` packages only. More tests are done on `rc2` packages.

shirhatti: So what's the consensus? `1.0.0-*` or `1.0.0-rc2-*`?

javiercn: @shirhatti Whatever you prefer, we filter our packages to RC2, so the end result is going to be the same

shirhatti: :+1:  I'm going to leave it at `1.0.0-rc2-*`

PR #41, opened 2016-03-29
Added a rule to check if packages are built in release configuration

ajaybhargavb:
@Eilon @pranavkm

### DIFF
--- a/src/NuGetPackageVerifier/PackageIssueLevel.cs
+++ b/src/NuGetPackageVerifier/PackageIssueLevel.cs
@@ -3,7 +3,7 @@
 
 namespace NuGetPackageVerifier
 {
-    public enum MyPackageIssueLevel
+    public enum PackageIssueLevel
     {
         Warning,
         Error
--- a/src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
+++ b/src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Mono.Cecil;
+using Mono.Collections.Generic;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class AssemblyHasCorrectBuildConfigurationRule : AssemblyHasAttributeRuleBase
+    {
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath, Collection<CustomAttribute> assemblyAttributes)
+        {
+            if (!HasReleaseConfiguration(assemblyAttributes))
+            {
+                yield return PackageIssueFactory.AssemblyHasIncorrectBuildConfiguration(currentFilePath);
+            }
+        }
+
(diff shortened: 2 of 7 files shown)


src/NuGetPackageVerifier/PackageIssueLevel.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace NuGetPackageVerifier
 {
-    public enum MyPackageIssueLevel
+    public enum PackageIssueLevel
```
Eilon: :cry: :panda_face:

src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (!HasReleaseConfiguration(assemblyAttributes))
+            {
+                yield return PackageIssueFactory.AssemblyHasIncorrectBuildConfiguration(currentFilePath);
+            }
+        }
+
+        private static bool HasReleaseConfiguration(Collection<CustomAttribute> assemblyAttributes)
+        {
+            var foundAttr = assemblyAttributes.SingleOrDefault(attr => attr.AttributeType.FullName == typeof(DebuggableAttribute).FullName);
```
pranavkm: Should we look for all instances of `DebuggableAttribute`? It's compiler generated, so they could possibly change how this is generated.

Eilon:
Take a look at what @pranavkm said, but aside from that, please make sure it actually passes, then :shipit:

src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
ajaybhargavb: I think we can just update the rule if it breaks later because I am not sure if checking for all instances guarantees that they will not break us.

src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
```diff
@@ -0,0 +1,40 @@
...
+            if (foundAttr == null)
+            {
+                return false;
+            }
+
+            var foundAttrArg = foundAttr.ConstructorArguments.SingleOrDefault();
+            var attrValue = (DebuggableAttribute.DebuggingModes)foundAttrArg.Value;
+            if (attrValue != DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)
```
pranavkm: This is a flags enum. Think you need to check for existence of the flag rather than equality

pranavkm: Maybe like this: http://stackoverflow.com/a/654461

src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
pranavkm: :ok:

ajaybhargavb:
:up: :date:

Eilon:
:shipit:

src/NuGetPackageVerifier/Rules/AssemblyHasCorrectBuildConfigurationRule.cs
pranavkm: `foundAttribute`

pranavkm:
:shipit:

PR #43, opened 2016-04-05
Default to moving files instead of copying to reduce Coherence drop size

pranavkm:


pranavkm:
cc @javiercn

javiercn:
:shipit:

Eilon:
@pranavkm still needed?

pranavkm:
We do, but need to change some stuff for the test runs before we can use this.

Eilon:
Ok.

PR #46, opened 2016-04-11
[Fixes #44][Fixes #45] Generate multiple project.json files and take into account the frameworks supported by our packages.

### LINKED ISSUE #44: Split packages should generate project.json based on the frameworks supported by the packages
Just putting all the packages in a dependencies node is not good enough. Packages need to be classified in three categories. 'All' for packages that support netstandardapp1.5 and net451, netstandardapp1.5 for packages that support netstandard or a compatible framework and net451 for packages that are exclusive for net451

### LINKED ISSUE #45: Split packages should generate multiple project.json files with different imports on specific target frameworks
Ideally, all of our packages should be able to be restored without requiring any imports in the project.json file. Imports are required right now due to third party dependencies and obsolete project.json definitions in our packages.

We want to generate a project.json file without imports so that we can run dotnet restore on it and determine which packages are responsible for requiring imports in the project.json file.

javiercn:


src/SplitPackages/Frameworks.cs
```diff
@@ -0,0 +1,116 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Frameworks;
+
+namespace SplitPackages
+{
+    public static class Frameworks
+    {
+        public static string NetStandardApp15 => FrameworkConstants.CommonFrameworks.NetStandardApp15.DotNetFrameworkName;
```
pranavkm: This you need to also add `NetCoreApp10` now. Our executables will be switching to this tfm very soon.

src/SplitPackages/Frameworks.cs
```diff
@@ -0,0 +1,116 @@
...
+        public static string Net451 => FrameworkConstants.CommonFrameworks.Net451.DotNetFrameworkName;
+        public static string DnxCore50 => FrameworkConstants.CommonFrameworks.DnxCore50.DotNetFrameworkName;
```
pranavkm: Do we have any projects targeting this as yet? We could do a quick pass to remove this entirely (or is this just an import for old code?)

src/SplitPackages/Frameworks.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                Framework = framework;
+            }
+
+            public FrameworkClasification(string framework, IEnumerable<string> imports)
+                : this(framework)
+            {
+                Imports = imports;
+            }
+            public bool IsAll => Framework == null;
```
pranavkm: New line

src/SplitPackages/ProjectJsonFileBuilder.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private void LogClassification(PackageItem dependency, Frameworks.FrameworkClasification classification)
+        {
+            var imports = !classification.Imports.Any() ?
+                "" :
+                $"with imports {string.Join(", ", classification.Imports.Select(i => Frameworks.GetMoniker(i)))}";
+
+            _logger.LogInformation($@"'{dependency.Identity}' with frameworks
+{string.Join(Environment.NewLine, dependency.SupportedFrameworks)} is classified as
+{classification.ToString()}
```
pranavkm: Shouldn't need the `ToString` here. It should be implicitly called.

src/SplitPackages/ProjectJsonFileBuilder.cs
```diff
@@ -0,0 +1,189 @@
...
+                if (!(classification.IsAll || classification.IsNet451 || classification.IsNetStandardApp15))
+                {
+                    throw new InvalidOperationException($"Cannot classify package '{dependency.Identity}'.");
+                }
+            }
+        }
+
+        private void LogClassification(PackageItem dependency, Frameworks.FrameworkClasification classification)
+        {
+            var imports = !classification.Imports.Any() ?
```
pranavkm: Could you use an if here? Kinda hard to read the ternary operator.

pranavkm:
:shipit:

src/SplitPackages/Frameworks.cs
javiercn: It's just an import for old code

PR #52, opened 2016-05-02
Add transitive reference to PlatformAbstractions

ryanbrandenburg:


javiercn:
What is this required for?

ryanbrandenburg:
This project references Microsoft.DotNet.ProjectModel, which recently lost PlatformAbstractions as a transitive dependency, but still needs it at build time.

@NTaylorMullen is that explanation any good?

ryanbrandenburg:
Not needed.

PR #55, opened 2016-05-12
Use case-insensitive comparison when checking the assembly public key token

joelverhagen:
This is a little usability improvement, especially considering `sn.exe -T` outputs a lowercase hex string.

@ajaybhargavb


Eilon:
@ajaybhargavb can you review? Once signed off, @joelverhagen can merge.

ajaybhargavb:
:shipit:

PR #56, opened 2016-05-12
Drop a version file in the package cache

victorhurdugaci:
Fixes https://github.com/aspnet/Coherence-Signed/issues/305

- Drop a marker file to keep track of the package cache version (needed by Antares)
- Changed the file extension from `zip` to `packagecache.zip` so that Antares can find it based on that

Please review @javiercn


javiercn:
:shipit:

PR #58, opened 2016-05-19
Run DefaultCompositeRule as part of SigningVerification

pranavkm:


src/NuGetPackageVerifier/Rules/AuthenticodeSigningRule.cs
```diff
@@ -17,14 +17,10 @@ public class AuthenticodeSigningRule : IPackageVerifierRule
             IPackageMetadata package,
             IPackageVerifierLogger logger)
         {
-            var packagePath = nupkgFile.FullName;
-            var nupkgWithoutExt = Path.Combine(
-                Path.GetDirectoryName(packagePath),
-                Path.GetFileNameWithoutExtension(packagePath));
-
+            var extractPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
pranavkm: I'll move this into a separate commit. This lets us run the verification on directories without write permissions (such as the CI drop share)

src/NuGetPackageVerifier/CompositeRules/SigningVerificationCompositeRule.cs
```diff
@@ -15,6 +15,7 @@ public class SigningVerificationCompositeRule : IPackageVerifierRule
             new AuthenticodeSigningRule(),
             new PowerShellScriptIsSignedRule(),
             new RequiredNuSpecInfoRule(),
+            new DefaultCompositeRule(),
```
pranavkm: There's a bunch of assemblies without `AssemblyDescription`. For some reason UniverseCoherence hasn't failed for this. Is this a newly introduced attribute?

pranavkm:
cc @ajaybhargavb \ @Eilon

src/NuGetPackageVerifier/CompositeRules/SigningVerificationCompositeRule.cs
Eilon: Can you remind me why we're adding this here? I vaguely recall discussing it but don't remember the reason.

As far as missing descriptions - which assemblies are missing it? That seems surprising...

pranavkm: We missed verification in BrowserLink because the file was missing. I thought I'd be nice to run strong name verify rule and serviceable rule on these binaries again just as a secondary check ?

Eilon: Ah sure, that's fine. It should be quick.

Which assemblies were missing metadata/descriptions?

pranavkm: Scaffolding for instance: https://github.com/aspnet/Scaffolding/blob/dev/src/Microsoft.VisualStudio.Web.CodeGeneration/Properties/AssemblyInfo.cs#L7.

Eilon: Ah you presumably mean the `project.json` file (which generates the various other attributes). I added it to my list of metadata to fix/add.

Eilon:
I guess :shipit: , assuming everything in all the repos is fixed, such that this will pass.

src/NuGetPackageVerifier/CompositeRules/SigningVerificationCompositeRule.cs
ajaybhargavb: We don't have `RequiredPackageMetadataRule` enabled yet. https://github.com/aspnet/DnxTools/issues/11

ajaybhargavb:
:shipit:

pranavkm:
Waiting on Coherence-Signed to pass first before I check this in.

src/NuGetPackageVerifier/Rules/AuthenticodeSigningRule.cs
```diff
@@ -62,21 +58,16 @@ public class AuthenticodeSigningRule : IPackageVerifierRule
             }
             finally
             {
-                CleanUpFolder(nupkgWithoutExt, logger);
+                CleanUpFolder(extractPath, logger);
```
javiercn: You might want to put this in it's own try catch and retry a couple of times. I've had to do this in other tools where I was removing temp directories

javiercn:
:shipit:

PR #65, opened 2016-06-09
Add a lock on process runner to prevent concurrent writes when writing to a string builder

javiercn:


javiercn:
@pranavkm

pranavkm:
:shipit:

PR #71, opened 2016-06-18
Generate extra project.json file with final versions

javiercn:


javiercn:
@pranavkm @victorhurdugaci

src/SplitPackages/Program.cs
```diff
@@ -159,10 +160,49 @@ private int Execute()
...
+
+            var builder = CreateBaseJsonFileBuilderForCache(destinationPath, "final.cache.project.json");
+            builder.AddDependencies(correctedPackages);
+            builder.Execute();
+        }
+
+        private string GetFinalVersion(string version)
+        {
+            var frameworkVersion = new NuGetVersion(version);
+            if (frameworkVersion.Release.Contains("rtm"))
```
pranavkm: Could we do `StartsWith("rtm-")`? Just so that it's consistent with Coherence-Signed? Ok to do this in 1.0.1 if it works fine now.

pranavkm:
:shipit:

javiercn:
We should have this logic in one place and flow the correct version across to the tools instead

PR #78, opened 2016-07-05
Add support for TeamCity logging

pranavkm:
cc @ajaybhargavb \ @Eilon

### DIFF
--- a/src/NuGetPackageVerifier/Program.cs
+++ b/src/NuGetPackageVerifier/Program.cs
@@ -21,14 +21,10 @@ namespace NuGetPackageVerifier
 
         public static int Main(string[] args)
         {
-            // TODO: Take a switch saying whether to use TeamCity logger
-
             // TODO: Show extraneous packages, exclusions, etc.
 
             // TODO: Get this from the command line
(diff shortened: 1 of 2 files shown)


src/NuGetPackageVerifier/Program.cs
```diff
@@ -213,8 +217,6 @@ public static int Main(string[] args)
 
                 foreach (var unprocessedPackage in unprocessedPackages)
                 {
-                    logger.LogWarning("\tUnprocessed package: {0} ({1})", unprocessedPackage.Id, unprocessedPackage.Version);
```
pranavkm: Coherence-Signed is full of these messages since we don't list any of the packages there.:

`EXEC Unprocessed package: Microsoft.AspNetCore.Authentication.Facebook (1.1.0-alpha1-21491)`

Doesn't seem super helpful having this around anymore. Let me know if you think otherwise.

Eilon: Yeah, agreed.

Eilon:
:shipit: 

And lesson learned for me: if I sprinkle enough "TODO" comments in my code, someone will eventually do it! 😄

PR #81, opened 2016-07-25
Enable upload package cache to Azure Storage

troydai:
@pranavkm @javiercn

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -17,7 +17,7 @@ k-standard-goals
     var dependenciesPackagerPath = Path.Combine(BASE_DIR, "src", "DependenciesPackager");
     var configuration = Environment.GetEnvironmentVariable("Configuration") ?? "Debug";
     Exec("dotnet", "publish " + dependenciesPackagerPath + " --no-build " + "--configuration " + configuration);
-    foreach(var project in new[]{ "NuGetPackageVerifier", "SplitPackages", "DependenciesPackager" })
+    foreach(var project in new[]{ "NuGetPackageVerifier", "SplitPackages", "DependenciesPackager", "PackageCacheUploader" })
     {
       var outputDir = Path.Combine(TARGET_DIR, "build");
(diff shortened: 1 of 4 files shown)


pranavkm:
cc @javiercn

src/DependenciesPackager/Program.cs
```diff
@@ -174,6 +178,11 @@ static int Main(string[] args)
                 "The prefix to use for the zip file name",
                 CommandOptionType.SingleValue);
 
+            var uploadToAzure = app.Option(
+                "--upload-to-azure",
```
pranavkm: Seems kinda out of place with the rest of the stuff this thing does. Would it make sense to have a different task that does the upload?

src/DependenciesPackager/Program.cs
```diff
@@ -702,6 +719,31 @@ private int RunDotnetRestore(string projectJson, string packagesFolderName)
...
         }
 
+        private async Task UploadPackagesCache(string zipFilePath)
+        {
+            if (string.IsNullOrEmpty(zipFilePath) || !File.Exists(zipFilePath))
+            {
+                throw new ArgumentException(nameof(zipFilePath));
+            }
+
+            var connectionString = Environment.GetEnvironmentVariable("AZURE_STORAGE_ACCOUNT");
```
pranavkm: Have these be passed in (instead of env variables).

pranavkm:
:watch: for the question about splitting the tool up.

src/DependenciesPackager/Program.cs
troydai: Yes we can do that. The con is 
1) Bigger makefile.proj  to accommodate more tasks and conditions;
2) Need to ensure zip file name is in sync between tasks.

troydai:
@pranavkm move everything to its own project.

pranavkm:
:shipit:

PR #82, opened 2016-07-26
Further updates to support build package cache on Linux CI

troydai:


pranavkm:
+ @javiercn

src/DependenciesPackager/Program.cs
```diff
@@ -164,10 +169,10 @@ static int Main(string[] args)
                 "Avoids deleting the package folders and the publish folder used for crossgen",
                 CommandOptionType.NoValue);
 
-            var runtimes = app.Option(
+            var runtime = app.Option(
```
pranavkm: Why do we need to do this?

troydai: The JIT is various from platform to platform (year the actual term probably should be platform to be more accurate). We need to specify the value to extract correct JIT from package. In addition to use this string in file name to distinguish package caches for different platforms.

src/DependenciesPackager/Program.cs
```diff
@@ -184,7 +189,7 @@ static int Main(string[] args)
                 packagesVersion,
                 useCli,
                 quiet,
-                runtimes,
+                runtime,
```
javiercn: Why this change? We want to support generating packages for both bitnesses x86 and x64 in zip package. If there isn't anything compelling you to change this, please leave as it is.

troydai: Then this should be architecture instead of runtime. You can't crossgen linux bits on windows. I doubt we can crossgen x64 bits on x86.

troydai: This is NGen's behavior. And I'm confirming this with @shhsu

javiercn: I'm aware of that, but it's easier to have a single parameter with the rids and just pass that, than break it into platform and architectures.

troydai: I haven't tried. But does it produce multiple zip file when multiple runtimes are passed in?

troydai: Confirmed with coreclr team. Crossgen has to be done on the same target platform, which means send in multiple runtimes is essentially useless. The only exception is that for Windows, it is supported to crossgen 32bits assembly on 64bits assembly.

My motivation of changing this parameter from multiple value to single value is to 1) eliminate incorrect assumption and 2) simplify work flow.

troydai: Settled with @glennc, we will break the windows package cache into two zip packages x64 vs x86. Therefore I'll keep my change.

troydai:
@javiercn sign off?

javiercn:
:shipit: and fix coherence-final and coherence

troydai:
Please take a look this as well: https://github.com/aspnet/Coherence-Final/pull/5

troydai:
Shut this down for now. The change requires too many moving pieces come together.

PR #86, opened 2016-07-28
Update DependenciesPackager output file name schema

troydai:
/cc @javiercn @glennc


troydai:
This is a minor change. I may not way for a sign off, since I want to test it out on CI tonight to avoid rush hour.

PR #88, opened 2016-07-29
Read connection string from environment var

troydai:
To prevent the connection strong from being printed in log.

/cc @pranavkm


Eilon:
@javiercn can you review?

troydai:
I'll close this actually. The change is not needed since Teamcity will automatic obscure secret in log. Thanks for the reminding, @Eilon .

PR #97, opened 2016-09-09
Don't consider net451 only packages to create optimized cache project…

pranavkm:
….json


pranavkm:
cc @javiercn \ @pakrym

PR #98, opened 2016-09-19
Rename repo to "BuildTools"

natemcmaster:
Remove additional DNX references.

cc @Eilon @anurse


natemcmaster:
Why is AppVeyor building this repo? Can we disable it?

Issue #94, opened 2016-09-06
NuGet packages folder now has lowercase paths, which affects DependenciesPackager


joelverhagen:
Right now, the dependencies packager tool is constructing paths using `OriginalCase` package ID and version, not `lowercase` in two places: [hash path](https://github.com/aspnet/DnxTools/blob/0da7ded2c9a5368db95e078d4326f4426a473766/src/DependenciesPackager/Program.cs#L437-L438), [directory name](
https://github.com/aspnet/DnxTools/blob/0da7ded2c9a5368db95e078d4326f4426a473766/src/DependenciesPackager/Program.cs#L487-L488).

This change in NuGet, along with it's downstream affected parties is outlined here:
https://github.com/NuGet/Home/issues/2522

I think the right fix here it to consume NuGet APIs to build these paths, much like what .NET CLI is doing ([example](https://github.com/dotnet/cli/blob/c5b2da0db5ae73da6fbbdbcd279607149f7426f5/src/Microsoft.DotNet.ProjectModel/Resolution/PackageDependencyProvider.cs#L58)). The NuGet type to use is probably [`VersionFolderPathResolver`](https://github.com/NuGet/NuGet.Client/blob/7aa6498e6adabc00753181d295ff2b4f97bb9e44/src/NuGet.Core/NuGet.Packaging/VersionFolderPathResolver.cs).

If this work is not done, the `DOTNET_HOSTING_OPTIMIZATION_CACHE` on Linux will no longer work, since the `.deps.json` file of .NET Core 1.1.0 apps will have lowercase paths. Windows will keep working since it's a case-insensitive file system, but it should be fixed as well since this is basically just a convenient coincidence. From what I understand, this will not break the Linux app, it will just mean startup time is much slower.

/cc @troydai @javiercn

muratg:
@troydai could you sync up with @victorhurdugaci and get this one? We may need to consume the new CLI to be able to do this, and not sure when that can happen.

cc @pranavkm

troydai:
There are two solutions, we discussed.

[...]

PR #100, opened 2016-09-20
Generate package cache in lower case layout.

troydai:
Issue #94 

@joelverhagen 

Adopt the cheaper solution.

### DIFF
--- a/src/DependenciesPackager/CrossgenContext.cs
+++ b/src/DependenciesPackager/CrossgenContext.cs
@@ -125,6 +125,7 @@ namespace DependenciesPackager
                 {
                     var subDir = CreateSubDirectory(outputPath, package, asset);
                     var targetPath = Path.Combine(subDir.FullName, asset.FileName);
+
                     if (!CrossgenAssembly(_stagingFolder, package, asset, targetPath))
                     {
                         _logger.LogWarning($"Failed to crossgen asset {asset.FileName}. Copy the original asset instead.");
@@ -140,11 +141,12 @@ namespace DependenciesPackager
             {
                 var desp = (PackageDescription)entry.Library;
                 var hash = desp.PackageLibrary.Files.Single(file => file.EndsWith(".sha512"));
+                var desination = Path.Combine(outputPath,
+                    entry.Library.Identity.Name.ToLowerInvariant(),
+                    entry.Library.Identity.Version.ToNormalizedString().ToLowerInvariant(),
+                    hash.ToLowerInvariant());
 
-                File.Copy(
-                    Path.Combine(entry.Library.Path, hash),
-                    Path.Combine(outputPath, entry.Library.Identity.Name, entry.Library.Identity.Version.ToNormalizedString(), hash),
-                    overwrite: true);
+                File.Copy(Path.Combine(entry.Library.Path, hash), desination, overwrite: true);
             }
         }
 
@@ -181,10 +183,13 @@ namespace DependenciesPackager
 
         private DirectoryInfo CreateSubDirectory(string cacheBasePath, PackageEntry package, LibraryAsset asset)
         {
+            // special treatment of the cases in the path to accommodating NuGet package search
+            // https://github.com/aspnet/BuildTools/issues/94
+            // https://github.com/NuGet/Home/issues/2522
             var subDirectoryPath = Path.Combine(
                 cacheBasePath,
-                package.Library.Identity.Name,
-                package.Library.Identity.Version.ToNormalizedString(),
+                package.Library.Identity.Name.ToLowerInvariant(),
+                package.Library.Identity.Version.ToNormalizedString().ToLowerInvariant(),
                 Path.GetDirectoryName(asset.RelativePath));
 
             _logger.LogInformation($"Creating sub directory on {subDirectoryPath}.");
--- a/src/DependenciesPackager/OutputFilesContext.cs
+++ b/src/DependenciesPackager/OutputFilesContext.cs
@@ -34,7 +34,7 @@ namespace DependenciesPackager
             _project = project;
             _logger = logger;
 
-            _dependencies = new HashSet<string>(_project.Dependencies.Select(d => d.Name));
+            _dependencies = new HashSet<string>(_project.Dependencies.Select(d => d.Name), StringComparer.OrdinalIgnoreCase);
         }
 
         public string OutputPath { get; }


src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -123,8 +123,8 @@ public void Crossgen(string outputPath)
             {
                 foreach (var asset in package.Assets)
                 {
-                    var subDir = CreateSubDirectory(outputPath, package, asset);
-                    var targetPath = Path.Combine(subDir.FullName, asset.FileName);
```
joelverhagen: `package.Assets` -- are these the DLLs and such in the package? If so, this is incorrect. Assets have original case. The things have have lowercase are things that have the package version and ID baked into the file name.

joelverhagen: Suppose there is a package "Newtonsoft.Json" with version "9.0.1-BETA" with a DLL "lib/net45/Newtonsoft.Json.dll".

Lowercased:

- <code>~/.nuget/packages/<b>newtonsoft.json</b></code>
- <code>~/.nuget/packages/newtonsoft.json/<b>9.0.1-beta</b></code>
- <code>~/.nuget/packages/newtonsoft.json/9.0.1-beta/<b>newtonsoft.json.9.0.1-beta.nupkg</b></code>
- <code>~/.nuget/packages/newtonsoft.json/9.0.1-beta/<b>newtonsoft.json.9.0.1-beta.nupkg.sha512</b></code>
- <code>~/.nuget/packages/newtonsoft.json/9.0.1-beta/<b>newtonsoft.json.nuspec</b></code>

Original case:

- <code>~/.nuget/packages/newtonsoft.json/9.0.1-beta/<b>lib/net45/Newtonsoft.Json.dll</b></code>

src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -179,13 +179,14 @@ private bool CrossgenAssembly(string publishFolder, PackageEntry package, Librar
...
         }
 
-        private DirectoryInfo CreateSubDirectory(string cacheBasePath, PackageEntry package, LibraryAsset asset)
+        private DirectoryInfo CreateSubDirectoryInLowerCase(string cacheBasePath, PackageEntry package, LibraryAsset asset)
         {
             var subDirectoryPath = Path.Combine(
                 cacheBasePath,
                 package.Library.Identity.Name,
                 package.Library.Identity.Version.ToNormalizedString(),
-                Path.GetDirectoryName(asset.RelativePath));
```
joelverhagen: You should be using NuGet APIs to build these paths:
https://github.com/NuGet/NuGet.Client/blob/4cccb13833ad29d6a0bcff055460d964f1b49cfe/src/NuGet.Core/NuGet.Packaging/VersionFolderPathResolver.cs

Available in the NuGet.Packaging package:
https://dotnet.myget.org/gallery/nuget-build

src/DependenciesPackager/CrossgenContext.cs
troydai: Ok, we leave only `sha512` and `dll` file in the packages cache. So in this case, can I simply the rules to be: lower case for all file path except the file name of the assemblies (dll), which are kept as they were.

src/DependenciesPackager/CrossgenContext.cs
troydai: Is this resolve available in the 3.5.0-beta2? Because it is used by project model: https://github.com/dotnet/cli/blob/rel/1.0.0-preview2/src/Microsoft.DotNet.ProjectModel/project.json#L12

I try to avoid upgrading the dependencies on CLI sdk.

joelverhagen: No, but CLI `rel/1.0.0` branch is depending on `3.6.0-beta.1.msbuild.9`, which has the proper functionality:
https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/project.json

src/DependenciesPackager/CrossgenContext.cs
joelverhagen: Pretty close. Any folder names in the .nupkg (e.g. `lib`) are left as-is as well. For example, `LIB/Net45/Newtonsoft.Json.dll` remains as-is:

`~/.nuget/packages/newtonsoft.json/9.0.1-beta/LIB/Net45/Newtonsoft.Json.dll`.

But yes, the `sha512` path has lowercase everything (after the `outputPath` that is).

troydai:
This is an example. Please note:
1. hash is lower case.
2. Assembly name is as-is.
3. No other files.
```
✔ ~/code/playground/PackageCache/out/1.0.1/x64/microsoft.aspnetcore.antiforgery 
21:45 $ ls -R
1.0.1

./1.0.1:
lib							microsoft.aspnetcore.antiforgery.1.0.1.nupkg.sha512

./1.0.1/lib:
netstandard1.3

./1.0.1/lib/netstandard1.3:
Microsoft.AspNetCore.Antiforgery.dll
```

src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -143,7 +144,7 @@ public void CopyPackageSignatures(string outputPath)
 
                 File.Copy(
                     Path.Combine(entry.Library.Path, hash),
-                    Path.Combine(outputPath, entry.Library.Identity.Name, entry.Library.Identity.Version.ToNormalizedString(), hash),
+                    Path.Combine(outputPath, entry.Library.Identity.Name, entry.Library.Identity.Version.ToNormalizedString(), hash).ToLower(),
```
joelverhagen: It looks like `outputPath` part will also be lowercased, which is not intended, right?

src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -181,10 +182,13 @@ private bool CrossgenAssembly(string publishFolder, PackageEntry package, Librar
...
         private DirectoryInfo CreateSubDirectory(string cacheBasePath, PackageEntry package, LibraryAsset asset)
         {
+            // special treatment of the cases in the path to accomodating NuGet package search
+            // https://github.com/aspnet/BuildTools/issues/94
+            // https://github.com/NuGet/Home/issues/2522
             var subDirectoryPath = Path.Combine(
                 cacheBasePath,
-                package.Library.Identity.Name,
-                package.Library.Identity.Version.ToNormalizedString(),
+                package.Library.Identity.Name.ToLower(),
```
joelverhagen: These should be `ToLowerInvariant`, since that's the method NuGet uses to lowercase these thingies.

troydai: ok.

src/DependenciesPackager/CrossgenContext.cs
troydai: good catch, it shouldn't be.

joelverhagen:
LGTM :shipit:

troydai:
@joelverhagen Thanks. I'll squash the commit and hold on the merge until we make sure the timeline is coordinated so that package cache is not broken (reversely) for 1.1.0

PR #101, opened 2016-09-21
Add package type verifier rule

natemcmaster:
Adds rule to verify package type. http://docs.nuget.org/create/package-types

Breaks the the format of NuGetPackageVerifier.json schema. But this only affect one usage which is going to be updated to this anyways. (Updated here: https://github.com/aspnet/EntityFramework.Tools/pull/1)

Old:
```
"Microsoft.EntityFrameworkCore.Tools.DotNet": {
    "DOC_MISSING": {
        "lib/netcoreapp1.0/dotnet-ef.dll": "Not a class library. Docs not required for CLI tools"
    }
},
```

New:
```
"Microsoft.EntityFrameworkCore.Tools.DotNet": {
	"nowarn": {
	    "DOC_MISSING": {
	        "lib/netcoreapp1.0/dotnet-ef.dll": "Not a class library. Docs not required for CLI tools"
	    }
	},
    "packageTypes": [ "DotnetCliTool" ]
},
```

### DIFF
--- a/src/NuGetPackageVerifier/PackageVerifierOptions.cs
+++ b/src/NuGetPackageVerifier/PackageVerifierOptions.cs
@@ -1,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace NuGetPackageVerifier
+{
+    public class PackageVerifierOptions
+    {
+        // ignored warnings (with a justification)
+        // key = issueid
+        // values = <filename, justification>
+        public IDictionary<string, IDictionary<string, string>> NoWarn { get; set; }
+            = new Dictionary<string, IDictionary<string, string>>(StringComparer.OrdinalIgnoreCase);
+
+        public IList<string> PackageTypes { get; set; }
+    }
+}
--- a/src/NuGetPackageVerifier/README.md
+++ b/src/NuGetPackageVerifier/README.md
@@ -1,0 +1,21 @@
+NuGetPackageVerifier
+--------------------
+
+Internal tool for verifying a nupkg meets certain requirements.
+
+NuGetPackageVerifier.json schema
+```js
+{
+    "$ruleSetName$": { /* if $ruleSetName$ == 'Default', the ruleset is run for packages not listed in any other ruleset */
+        "rules": [ "$ruleNameToRun$" ],
+        "packages": {
+            "$packageId$": {
+                "nowarn": {
+                    "$warningId$": "$justification$"
+                },
+                "packageTypes": [ "$packageType$" ] /* Optional. For validating http://docs.nuget.org/create/package-types */
+            }
+        }
+    }
+}
+```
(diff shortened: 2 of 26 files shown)


natemcmaster:
@ajaybhargavb could you review this? I think you most wrote the NuGetPackage verifier, right?

src/NuGetPackageVerifier/Rules/PackageTypesRule.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace NuGetPackageVerifier.Rules
+{
+    public class PackageTypesRule : IPackageVerifierRule
+    {
+        public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
+        {
+            var discoveredTypes = new HashSet<string>(context.Metadata.PackageTypes.Select(t=>t.Name));
+            var expectedTypes = new HashSet<string>(context.Options.PackageTypes ?? Enumerable.Empty<string>());
+
+            var intersection = new HashSet<string>(expectedTypes.Intersect(discoveredTypes));
```
ajaybhargavb: You don't have to do it but I think this code can be simplified by doing something like,
```c#
var missingTypes = expectedTypes.Except(discoveredTypes);
var unexpectedTypes = discoveredTypes.Except(expectedTypes);
```

src/NuGetPackageVerifier/Rules/PackageTypesRule.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class PackageTypesRule : IPackageVerifierRule
+    {
+        public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
+        {
+            var discoveredTypes = new HashSet<string>(context.Metadata.PackageTypes.Select(t=>t.Name));
```
ajaybhargavb: nit: `t=>t.Name` --> `t => t.Name`

src/NuGetPackageVerifier/README.md
```diff
@@ -0,0 +1,21 @@
...
+Internal tool for verifying a nupkg meets certain requirements.
+
+NuGetPackageVerifier.json schema
+```js
+{
+    "$ruleSetName$": { /* if $ruleSetName$ == 'Default', the ruleset is run for packages not listed in any other ruleset */
+        "rules": [ "$ruleNameToRun$" ],
+        "packages": {
+            "$packageId$": {
+                "nowarn": {
```
natemcmaster: Will do.

src/NuGetPackageVerifier/Rules/PackageTypesRule.cs
natemcmaster: 👍 yes, much better implementation. Will update.

src/NuGetPackageVerifier/PackageVerifierOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+
+namespace NuGetPackageVerifier
+{
+    public class PackageVerifierOptions
+    {
+        // ignored warnings (with a justification)
+        // key = issueid
+        // values = <filename, justification>
+        public IDictionary<string, IDictionary<string, string>> NoWarn { get; set; }
```
Eilon: These aren't "NoWarn", they are "Exclusions". Can we rename?

PR #102, opened 2016-09-28
Organize CompositeRules

ryanbrandenburg:
Makes the composite rules duplicate less as per @pranavkm's request.

### DIFF
--- a/src/NuGetPackageVerifier/CompositeRules/DefaultCompositeRule.cs
+++ b/src/NuGetPackageVerifier/CompositeRules/DefaultCompositeRule.cs
@@ -9,10 +9,8 @@ namespace NuGetPackageVerifier.Rules
     {
         IPackageVerifierRule[] _rules = new IPackageVerifierRule[]
         {
-            new AssemblyHasCommitHashAttributeRule(),
             new AssemblyHasCompanyAttributeRule(),
             new AssemblyHasCopyrightAttributeRule(),
-            new AssemblyHasCorrectBuildConfigurationRule(),
             new AssemblyHasCorrectJsonNetVersionRule(),
             new AssemblyHasDocumentFileRule(),
             new AssemblyHasDescriptionAttributeRule(),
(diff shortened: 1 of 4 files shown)


src/NuGetPackageVerifier/Rules/AssemblyHasReleaseConfigurationRule.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace NuGetPackageVerifier.Rules
 {
-    public class AssemblyHasCorrectBuildConfigurationRule : AssemblyHasAttributeRuleBase
+    public class AssemblyHasReleaseConfigurationRule : AssemblyHasAttributeRuleBase
```
pranavkm: `AssemblyIsBuiltInReleaseConfiguraitonRule`?

src/NuGetPackageVerifier/CompositeRules/DefaultCompositeRule.cs
```diff
@@ -9,10 +9,8 @@ public class DefaultCompositeRule : IPackageVerifierRule
     {
         IPackageVerifierRule[] _rules = new IPackageVerifierRule[]
```
pranavkm: @Eilon we'll keep the two rules around for the time being. This avoids the need to update the verifier json in all the repos at this point.

src/NuGetPackageVerifier/Rules/AssemblyHasReleaseConfigurationRule.cs
Eilon: Meh. Why change this?

pranavkm: We had a hard time figuring out what the name of Release configuration rule was. Having `Release` in the name would have helped

Eilon: O I C. Ok fine rename it. I guess it doesn't show up anywhere outside of here.

PR #110, opened 2016-10-10
Introducing tool to check for api breaking changes.

javiercn:
This tool generates a baseline for the APIs of an assembly and allows comparing the baselines of different versions of an assembly.

The baseline is a json document listing all the types on an assembly and their members (filtering can be set to exported types (public, protected) and to skip types on the .Internal namespace).

All elements of the baseline contain an Id which is equivalent to their C# signature.

The comparison works by comparing signatures and ensuring that all the elements in the baseline on the left, are present on the baseline on the right. The comparison happens in two steps, first we try to find the existing type on the new baseline by comparing Ids. If we find it, we move on to compare the members within the type for any missing member. If no type with the same Id is found on the new baseline, we try to find a type with the same name, we record the _potential_ breaking change (due to a signature change) and we move on to compare the methods withing the types. For example:
Given the following class (V1)

``` C#
namespace Namespace {
    public class SampleChange : IEnumerable<System.Int32>
    {
        public System.Void MethodA()
    }
}
```

And the following class (V2)

``` C#
namespace Namespace {
    public class SampleChange : IEnumerable<System.String>
    {
       public System.Void MethodA()
    }
}
```

Their Ids are respectively:
`public class SampleChange : IEnumerable<System.Int32>`
`public class SampleChange : IEnumerable<System.String>`

When we can't find any item with an Id equal to `public class SampleChange : IEnumerable<System.Int32>` we fallback to find a type with a `Namespace.SampleChange` name and compare their members.

Foreach change that we find, we print a line on the console with the old member (and the new member if we find it) or an indication that it wasn't found.


src/ApiCheck/AssemblyLoader.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public static Assembly LoadAssembly(
+                string assemblyPath,
+                string lockFile,
+                NuGetFramework framework,
+                string packagesFolder)
+        {
+            var ctx = CreateProjectContext(framework, lockFile, packagesFolder);
+            var exporter = ctx.CreateExporter("Pranav");
```
javiercn: Missing cleanup, sorry about that :p

javiercn:
The tool still needs a dispatcher and to print the output to the command line

src/ApiCheck/BaselineGenerator.cs
```diff
@@ -0,0 +1,350 @@
...
+                    fieldBaseline.ReturnType = TypeBaseline.GetTypeNameFor(field.FieldType.GetTypeInfo());
+
+                    return fieldBaseline;
+                case MemberTypes.Event:
+                case MemberTypes.Property:
+                case MemberTypes.NestedType:
+                    // All these cases are covered by the methods they implicitly define on the class
+                    // (Properties and Events) and when we enumerate all the types in an assembly (Nested types).
+                    return null;
+                case MemberTypes.TypeInfo:
```
javiercn: NestedTypes are listed in Assembly.DefinedTypes which flattens everything and makes it easier to deal with generating the baseline without resorting to recursion.

src/ApiCheck/Baseline/MemberBaseline.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            if (New)
+            {
+                yield return "new";
+            }
+
+            if (ReturnType != null)
+            {
+                yield return ReturnType;
+            }
```
javiercn: The return type is System.Void for void returning methods. ReturnType == null for constructors only

src/ApiCheck/Baseline/BaselineVisibility.cs
```diff
@@ -0,0 +1,15 @@
...
+using Newtonsoft.Json.Converters;
+
+namespace ApiCheck.Baseline
+{
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum BaselineVisibility
+    {
+        Public,
+        Protected,
+        Internal,
```
javiercn: Its interesting to have the full classification for testing purposes. It's equally important to make sure that the tool classifies public, protected, etc types and members as public, protected, etc. As it is to make sure that something internal or private does not get classified with a different visibility by mistake.

The tool has other mechanisms in place to filter the set of types included in the generated API listing that filter any unwanted type (like anything non public/protected or anything on the .Internal namespace).

src/ApiCheck/BaselineComparer.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using ApiCheck.Baseline;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace ApiCheck
+{
+    public class BaselineComparer
+    {
+        private readonly BaselineDocument _newBaseline;
```
javiercn: Is ApiDescription good enough or your prefer ApiListing (or something else) I don't have a strong opinion about any of them.

src/ApiCheck/BaselineGenerator.cs~RF4c64d1e.TMP
```diff
@@ -0,0 +1,754 @@
+﻿using System;
```
javiercn: VS did its magic here, I've removed it.

src/ApiCheck/BreakingChangeHandlers.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace ApiCheck
+{
+    public class BreakingChangeHandlers
+    {
+        public static bool FindTypeUsingFullName(BreakingChangeContext ctx)
```
javiercn: Fair enough, I will take it out and when I don't find a type I will just spill the guts for it and all of its members, if we have a lot of noise we can bring it back

src/ApiCheck/project.json
```diff
@@ -0,0 +1,26 @@
...
+    "Microsoft.DotNet.ProjectModel": "1.0.0-*",
+    "Microsoft.Extensions.CommandLineUtils": "1.0.0",
+    "Microsoft.Extensions.Logging": "1.0.0",
+    "Microsoft.Extensions.Logging.Console": "1.0.0",
+    "Newtonsoft.Json": "9.0.1"
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
```
javiercn: This can't be top level as its not compatible with net452

test/ApiCheck.Test/BaselineComparerTests.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            // Act
+            var changes = comparer.GetDifferences();
+
+            // Assert
+            var change = Assert.Single(changes, bc => bc.OldItem.Id == "public struct ComparisonScenarios.StructToMakeGeneric");
+        }
+
+        [Fact]
+        public void Compare_Detects_NamespaceChanges()
```
javiercn: While I do agree that a fundamental level that is what is happening, I want to make sure that the tests are worded in a higher level language that reflects the logical action that happened on the code base. (IE, I moved a type from namespace A into namespace B or nested it into another class and so on)

test/ApiCheck.Test/ReportGenerationTests.cs
```diff
@@ -0,0 +1,974 @@
...
+            var report = generator.GenerateBaseline();
+
+            // Assert
+            Assert.NotNull(report);
+            Assert.NotNull(report.Types);
+            Assert.Single(report.Types, t => t.Id == "public interface Scenarios.IComplexInterface : Scenarios.IBasicInterface");
+        }
+
+        [Fact]
+        public void DetectsMultipleLevelInterface()
```
javiercn: It means an interface that is part of a hierarchy of more than one level of interaces, for example:
````C#
public interface A {}
public interface B : A {}
public interface C : B {}
```

javiercn:
Updated

src/ApiCheck/Baseline/BaselineVisibility.cs
Eilon: I think we can exclude `private`, and then you can add a test that makes sure that something that's actually private just doesn't show up. I could see this generating a lot of noise in various reports.

But it might actually make sense to keep `internal` because of InternalsVisibleTo and cases where we require 100% binary compatibility...

src/ApiCheck/ApiListingComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class ApiListingComparer
+    {
+        private readonly ApiListing _newBaseline;
+        private readonly ApiListing _oldBaseline;
+
+        public ApiListingComparer(
+            ApiListing oldBaseline,
+            ApiListing newBaseline)
+        {
+            _oldBaseline = oldBaseline;
```
Eilon: Change parameter/field names here and perhaps elsewhere.

src/ApiCheck/BaselineGenerator.cs
Eilon: That's good, but then wouldn't this code be a problem? This would presumably explode on nested types, no? I see there are some tests around this but I don't see how this code can work...

src/ApiCheck/Program.cs
```diff
@@ -0,0 +1,254 @@
...
+                generationFilters.Add(t => t.IsPublic || t.IsNestedPublic || t.IsNestedFamily);
+                reloadingFilters.Add(t =>
+                    t.Visibility == BaselineVisibility.Public ||
+                    t.Visibility == BaselineVisibility.Protected ||
+                    t.Visibility == BaselineVisibility.ProtectedInternal);
+            }
+
+            if (excludeInternalNamespace.HasValue())
+            {
+                generationFilters.Add(t => !t.Namespace.EndsWith("Internal"));
```
Eilon: Actually even that isn't exactly right: If there is an "Internal" segment anywhere in the namespace it's "internal", so the check isn't EndsWith, rather, it's `.Split(".").Any(ns => ns == "Internal")` or something like that.

test/ApiCheck.Test/ReportGenerationTests.cs
```diff
@@ -0,0 +1,1080 @@
...
+            var generator = CreateGenerator(V1Assembly);
+
+            // Act
+            var report = generator.GenerateBaseline();
+
+            // Assert
+            Assert.NotNull(report);
+            Assert.NotNull(report.Types);
+            var type = Assert.Single(report.Types, t => t.Id == "public class Scenarios.MethodTypesClass");
+            var method = Assert.Single(type.Members, m => m.Id == "private System.Boolean PrivateBoolReturningMethodWithOptionalCharParameter(System.Char charParameter = 'c')");
```
javiercn: Classification is separated from filtering on the tool. We want to make sure all elements are categorized correctly independently of whether we filter them or not.

javiercn:
Addressed remaining feedback

src/ApiCheck/ApiListing/ApiListingFilters.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+using ApiCheck.Description;
+
+namespace ApiCheck
+{
+    public static class ApiListingFilters
+    {
+        public static bool NonExportedMembers(MemberInfo e)
```
Eilon: Predicates usually have some interrogative question in them, e.g. `HasNonExportedMember` - is that what is meant here? (And the other filters?)

Eilon: (Otherwise I don't quite understand what this predicate is doing.)

src/ApiCheck/ApiListing/ApiListingFilters.cs
```diff
@@ -0,0 +1,79 @@
...
+        public static bool InternalNamespaceTypes(ApiElement e)
+        {
+            var type = e as TypeDescriptor;
+            if (type == null)
+            {
+                return false;
+            }
+
+            var segments = type.Name.Split('.');
+            return segments.Length > 1 && segments[segments.Length - 2].Equals("Internal");
```
Eilon: What is this check checking?

I think it can be simpler (and correct) this way: `type.NAMESPACE.Split('.').Any(ns => string.Equals(ns, "Internal", StringComparison.Ordinal);`

Note: You really want to check just the namespace, not the type name (though that's of course super unlikely to ever matter), and then check if ANY of those segments is exactly the word `Internal`.

src/ApiCheck/ApiListing/MemberKind.cs
```diff
@@ -0,0 +1,16 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace ApiCheck.Description
+{
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum MemberKind
+    {
+        Constructor,
+        Method,
```
Eilon: Might be good to add a doc comment to this enum value specifically to indicate that this represents the accessor methods of properties and events, and thus there are no specific member kinds for properties and events.

src/ApiCheck/ApiListing/TypeKind.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace ApiCheck.Description
+{
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum TypeKind
```
Eilon: I guess we never have `delegate` types? Or are those bundled into one of these existing types?

src/ApiCheck/ApiListingGenerator.cs
```diff
@@ -0,0 +1,409 @@
...
+            foreach (var type in _assembly.DefinedTypes.Where(type => _filters.All(filter => !filter(type))))
+            {
+                var ApiListingType = GeneratetypeDescriptor(type);
+                document.Types.Add(ApiListingType);
+            }
+
+            return document;
+        }
+
+        private TypeDescriptor GeneratetypeDescriptor(TypeInfo type)
```
Eilon: Missing a capital `T` in the method name 😄

src/ApiCheck/ApiListingGenerator.cs
```diff
@@ -0,0 +1,409 @@
...
+        {
+            var typeDescriptor = new TypeDescriptor();
+
+            typeDescriptor.Name = TypeDescriptor.GetTypeNameFor(type);
+
+            typeDescriptor.Kind = type.IsInterface ? TypeKind.Interface :
+                !type.IsValueType ? TypeKind.Class :
+                type.IsEnum ? TypeKind.Enumeration :
+                TypeKind.Struct;
```
Eilon: See earlier comment: this seems like a slightly dangerous expression to me because it assumes "all unknown things are structs" and I'm not certain that's true. Might want to explicitly check for structs (which I realize might not be trivial), and then have an "unknown" case that can perhaps be treated as a bug in the logic here (i.e. don't ignore unknown types - but rather fail on them).

src/ApiCheck/ApiListingGenerator.cs
```diff
@@ -0,0 +1,409 @@
...
+                    }
+
+                    methodDescriptor.ReturnType = TypeDescriptor.GetTypeNameFor(method.ReturnType.GetTypeInfo());
+
+                    return methodDescriptor;
+                case MemberTypes.Field:
+                    var field = (FieldInfo)member;
+                    if (type.IsEnum && !field.IsLiteral)
+                    {
+                        // Skip storage for enumerations.
```
Eilon: Hmm does this mean it's not recording the values of enums? It probably should be, because that's also a really bad breaking change, no? Or is this something else?

src/ApiCheck/ApiListingGenerator.cs
```diff
@@ -0,0 +1,409 @@
...
+                        fieldDescriptor.ReadOnly = field.IsInitOnly;
+                        fieldDescriptor.ReturnType = TypeDescriptor.GetTypeNameFor(field.FieldType.GetTypeInfo());
+                    }
+
+                    return fieldDescriptor;
+                case MemberTypes.Event:
+                case MemberTypes.Property:
+                case MemberTypes.NestedType:
+                    // All these cases are covered by the methods they implicitly define on the class
+                    // (Properties and Events) and when we enumerate all the types in an assembly (Nested types).
```
Eilon: I like the comments here, thanks!

src/ApiCheck/ApiListingGenerator.cs
```diff
@@ -0,0 +1,409 @@
...
+                    candidateParameter.IsOptional != methodParameter.IsOptional)
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private ParameterDescriptor GenerateparameterDescriptor(ParameterInfo parameter)
```
Eilon: `P` in method name.

src/ApiCheck/Program.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            var generator = new ApiListingGenerator(assembly, newApiListingFilters);
+            var newApiListing = generator.GenerateApiListing();
+
+            var comparer = new ApiListingComparer(oldApiListing, newApiListing);
+
+            var differences = comparer.GetDifferences();
+            foreach (var difference in differences)
+            {
+                logger.LogInformation($@"Missing class or member on {difference}");
```
Eilon: `Missing *type* or member on ...`

Eilon:
Re: private types (GitHub won't let me reply to the discussion thread ☹️) --

I don't see the point of classifying private types. They have no meaning whatsoever when it comes to any kind of API reviews, listings, or checks. There will never be a need for them to show up in this system. As such, there is no value to classifying them, because at most they are noise. I'm fine having tests for them, but only negative tests, to ensure that private types don't show up in the listing. (We're not trying to build some general purpose API system - we're building an API compatibility check.)

javiercn:
Updated

src/ApiCheck/ApiListing/ApiListingFilters.cs
javiercn: This checks that the last segment of the namespace is .Internal, for example
Scenario.Internal.Class or
Scenario.Internal.Class+NestedClass

src/ApiCheck/ApiListing/TypeKind.cs
javiercn: Delegates are tracked as classes given that they are just classes that extend MulticastDelegate

src/ApiCheck/ApiListingGenerator.cs
javiercn: I think the right answer for this is to have tests (which we have), not an extra value. That said, it's cheap so I've added an extra Undefined value and throw immediately after.

src/ApiCheck/ApiListing/ApiListingFilters.cs
Eilon: This still isn't the right check. The namespace "Foo.Internal.Bar" is still internal and that's what we need to check here.

src/ApiCheck/ApiListing/ApiListingFilters.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public static bool IsInInternalNamespace(MemberInfo e)
+        {
+            var type = e as TypeInfo;
+            if (type == null)
+            {
+                return false;
+            }
+
+            return type.Namespace.EndsWith(".Internal");
```
Eilon: See comment below as to why this is incorrect.

src/ApiCheck/ApiListing/TypeKind.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        Undefined,
```
Eilon: BTW this is more like `Unknown` as opposed to undefined.

src/ApiCheck/Loader/CoreClrAssemblyLoader.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        static CoreClrAssemblyLoader()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                NativeLibraryExtensions = new[] { ".dll" };
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                NativeLibraryExtensions = new[] { ".dylib" };
```
Eilon: Why is there any logic anywhere for native dependencies?

src/ApiCheck/project.json
```diff
@@ -0,0 +1,28 @@
...
+          "type": "platform",
+          "version": "1.0.0"
+        },
+        "System.Runtime.Loader": "4.0.0"
+      }
+    },
+    "net452": {
+      "dependencies": {
+        "System.Runtime.InteropServices.RuntimeInformation": "4.3.0-preview1-*"
```
Eilon: Do we need a pre-rel package for this?

src/ApiCheck/ApiListing/ApiListingFilters.cs
javiercn: @Eilon ah, ok. I always thought that our rule was BaseNamespace+.Internal only. No wonder why I didn't get what you were saying.

src/ApiCheck/Loader/CoreClrAssemblyLoader.cs
javiercn: It's copy/pasta from the razor tooling version. We'll have to change this in a month or so, so I'm not really worried about it right now.

javiercn:
Updated

src/ApiCheck/ApiListingComparer.cs
```diff
@@ -42,12 +58,42 @@ public IList<BreakingChange> GetDifferences()
...
+                        }
                     }
                 }
             }
 
-            return breakingChanges;
+            return new ApiComparisonResult(breakingChanges, exclusions);
+        }
+
+        private BreakingChange GetInvalidChange(TypeDescriptor type, MemberDescriptor member, List<ApiChangeExclusion> exclusions)
```
Eilon: Need to come up with a better name for this... it's not "invalid".

src/ApiCheck/Program.cs
```diff
@@ -195,9 +211,20 @@ public static int Main(string[] args)
                 Console.WriteLine();
             }
 
-            if (differences.Count > 0)
+            var remainingExclusions = result.RemainingExclusions;
+            foreach (var exclusion in remainingExclusions)
+            {
+                Console.WriteLine($@"Exclusion not applied:
```
Eilon: `The following exclusion is in the exclusion file, but is no longer necessary:`

Eilon: BTW it would also be nice to either colorize or at least prefix each "error" case with something like "ERROR: " so that it's super clear it's a bad thing.

javiercn:
updated

src/ApiCheck/ApiListingComparer.cs
```diff
@@ -70,6 +80,130 @@ public ApiComparisonResult GetDifferences()
...
+        }
+
+        private bool HasCompatibleVisibility(TypeDescriptor oldType, TypeDescriptor newType)
+        {
+            switch (oldType.Visibility)
+            {
+                case ApiElementVisibility.Public:
+                    return newType.Visibility == ApiElementVisibility.Public;
+                case ApiElementVisibility.Protected:
+                    // Is going from protected to public a breaking change ?
```
Eilon: Good question. Assume for the moment it is a breaking change because we really shouldn't be doing that 😄

src/ApiCheck/ApiListingComparer.cs
```diff
@@ -70,6 +80,130 @@ public ApiComparisonResult GetDifferences()
...
+                    return true;
+                default:
+                    throw new InvalidOperationException("Unrecognized visibility");
+            }
+        }
+
         private BreakingChange GetInvalidChange(TypeDescriptor type, MemberDescriptor member, List<ApiChangeExclusion> exclusions)
```
Eilon: Mentioned in a previous iteration but not yet addressed: `Need to come up with a better name for this... it's not "invalid".`

src/ApiCheck/ApiListingGenerator.cs
javiercn: It's the internal storage for enums (they are in the end structs with a generated field, this just skips that field).

PR #111, opened 2016-10-14
Microsoft.NETCore.App 1.0.0 => 1.1.0

troydai:
@pranavkm @javiercn

### DIFF
--- a/src/SplitPackages/Program.cs
+++ b/src/SplitPackages/Program.cs
@@ -227,7 +227,7 @@ namespace SplitPackages
 
             var hardcodedDependencies = new[]
             {
-                new PackageInformation(null, "Microsoft.NetCore.App", "1.0.0", new[] { Frameworks.NetCoreApp10 })
+                new PackageInformation(null, "Microsoft.NetCore.App", "1.1.0", new[] { Frameworks.NetCoreApp10 })
             };
 
             builder.AddFramework(Frameworks.NetCoreApp10);


src/SplitPackages/Program.cs
```diff
@@ -227,7 +227,7 @@ private ProjectJsonFileBuilder CreateBaseJsonFileBuilderForCache(string destinat
 
             var hardcodedDependencies = new[]
             {
-                new PackageInformation(null, "Microsoft.NetCore.App", "1.0.0", new[] { Frameworks.NetCoreApp10 })
+                new PackageInformation(null, "Microsoft.NetCore.App", "1.1.0", new[] { Frameworks.NetCoreApp10 })
```
pranavkm: Would this affect tools? We're still trying to figure out if tools should continue targeting `Microsoft.NETCore.App 1.0.0`.

troydai: This is used to generate the cache.project.json and final.cache.project.json file. These two files are used to pick correct packages for package cache.

Are these two files used for other purpose? Or is there other unexpected outcome of this tool?

/cc @javiercn

pranavkm: I wasn't sure if having a `netcoreapp1.0` targeting tool would affect this (those are the only things we ship that target netcoreapp). If this is specifically used to produce a project.json, this should be fine

troydai: from my reading on the source code, this is the case.

pranavkm:
:shipit:

PR #114, opened 2016-11-04
Updated ApiCheck to include Microsoft.AspNetCore.BuildTools due to conflicts with another package

javiercn:


javiercn:
@pranavkm @Eilon FYI, I'm merging to unblock integration

javiercn:
There was a package with the same name already

Eilon:
@javiercn where was the actual conflict? Something on nuget.org? Or somewhere else?

PR #117, opened 2016-11-06
Correctly parse dependencies under each TFM

JunTaoLuo:
@pranavkm While updating packages for the 1.0.3 patch I found that dependencies under specific TFMs were not being updated by the version tool. I think the tool is looking for the dependencies in the wrong place?

Also, not sure what runtime dependencies look like so I'm not sure if this applies.


Eilon:
@pranavkm please take a look when you return back to civilization.

PR #118, opened 2016-11-08
Some bug fixes to the baseline generation and breaking change detection logic

javiercn:
Fixed bug with nested generics.
Fixed bug about abstract classes changing to non abstract being detected as a breaking change.
Fixed bug about members moving up in a class hierarchy where the base class is on a different assembly.

### DIFF
--- a/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListing/TypeDescriptor.cs
+++ b/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListing/TypeDescriptor.cs
@@ -157,9 +157,37 @@ namespace ApiCheck.Description
 
             if (type.IsGenericType)
             {
-                var name = type.GetGenericTypeDefinition().FullName;
-                typeName = name.Substring(0, name.IndexOf('`'));
-                typeName = $"{typeName}<{string.Join(", ", type.GetGenericArguments().Select(ga => GetTypeNameFor(ga.GetTypeInfo())))}>";
+                if (type.DeclaringType == null || type.DeclaringType.GetTypeInfo() == type)
+                {
+                    var name = type.GetGenericTypeDefinition().FullName;
+                    typeName = name.Substring(0, name.IndexOf('`'));
+                    typeName = $"{typeName}<{string.Join(", ", type.GetGenericArguments().Select(ga => GetTypeNameFor(ga.GetTypeInfo())))}>";
+                }
+                else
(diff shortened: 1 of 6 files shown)


javiercn:
@Eilon @pranavkm FYI, I'm checking it in to unblock the remaining baselines that are not check-in yesterday

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListing/TypeDescriptor.cs
```diff
@@ -157,9 +157,37 @@ public static string GetTypeNameFor(TypeInfo type)
...
+                    typeName = name.Substring(0, name.IndexOf('`'));
+                    typeName = $"{typeName}<{string.Join(", ", type.GetGenericArguments().Select(ga => GetTypeNameFor(ga.GetTypeInfo())))}>";
+                }
+                else
+                {
+                    var container = type.DeclaringType.GetTypeInfo();
+                    var prefix = GetTypeNameFor(container);
+                    var name = type.GetGenericTypeDefinition().FullName;
+                    var currentTypeGenericArguments = type.GetGenericTypeDefinition().GetGenericArguments()
+                    .Where(p => !container.GetGenericArguments().Any(cp => cp.Name == p.Name))
```
pranavkm: Can you skip these instead. `.Skip(container.GetGenericArguments().Length)`? I think the arguments from the containing type always appear before the nested types.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingGenerator.cs
```diff
@@ -151,31 +159,39 @@ private static TypeKind GetTypeKind(TypeInfo type)
...
 
-                if (typeArgument.BaseType != null &&
-                    typeArgument.BaseType != typeof(object)
-                    && typeArgument.BaseType != typeof(ValueType))
+                if (typeArgument.IsGenericParameter)
                 {
-                    constraintDescriptor.BaseTypeOrInterfaces.Add(TypeDescriptor.GetTypeNameFor(typeArgument.BaseType.GetTypeInfo()));
-                }
+                    if (typeArgument.BaseType != null &&
+                        typeArgument.BaseType != typeof(object)
```
pranavkm: `typeArgument.BaseType != typeof(object) &&` (to be consistent with the previous line)

PR #119, opened 2016-11-09
Scan the package for a compatible NetStandard dll in case one is not available on the library export

javiercn:


javiercn:
@pranavkm @Eilon FYI, this is a fix for System.Data.SqlClient.dll failing to load, I diagnosed it with @pakrym and came up with this solution, it's not general purpose, but works for our needs.

PR #120, opened 2016-11-09
Detect added members to interfaces

javiercn:
@Eilon @pranavkm

### DIFF
--- a/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
+++ b/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
@@ -58,22 +58,47 @@ namespace ApiCheck
 
         private void CompareMembers(TypeDescriptor type, TypeDescriptor newType, List<ApiChangeExclusion> exclusions, List<BreakingChange> breakingChanges)
         {
+            var removedOrChanged = 0;
             foreach (var member in type.Members)
             {
                 var newMember = newType.FindMember(member.Id);
                 var isAcceptable = IsAcceptableMemberChange(_newApiListing, newType, member);
                 if (isAcceptable)
                 {
+                    if (newMember == null)
+                    {
+                        removedOrChanged++;
+                    }
+
                     continue;
                 }
 
                 if (newMember == null)
                 {
+                    removedOrChanged++;
                     var memberChange = FilterExclusions(type, member, exclusions);
                     if (memberChange != null)
                     {
                         breakingChanges.Add(memberChange);
                     }
+                }
+            }
+
+            if (type.Kind == TypeKind.Interface && type.Members.Count - removedOrChanged < newType.Members.Count)
+            {
+                var members = newType.Members.ToList();
+                foreach (var member in newType.Members)
+                {
+                    var change = FilterExclusions(type, null, exclusions);
+                    if (change == null)
+                    {
+                        members.Remove(member);
+                    }
+                }
+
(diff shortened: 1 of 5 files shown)


javiercn:
This check was missing

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ChangeKind.cs
```diff
@@ -14,6 +14,7 @@ namespace ApiCheck
     public enum ChangeKind
     {
         Removal,
-        Modification
+        Modification,
+        Addition
```
Eilon: Yo, add a trailing comma!,

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
```diff
@@ -76,6 +76,24 @@ private void CompareMembers(TypeDescriptor type, TypeDescriptor newType, List<Ap
                     }
                 }
             }
+
+            if (type.Kind == TypeKind.Interface && type.Members.Count < newType.Members.Count)
```
pranavkm: Is it sufficient to verify counts? For instance if the member was renamed, the count is the same.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
```diff
@@ -76,6 +76,24 @@ private void CompareMembers(TypeDescriptor type, TypeDescriptor newType, List<Ap
...
+            {
+                var members = newType.Members.ToList();
```
pranavkm: `var nonFilteredMembers = newType.Members.Where(type => FilterExclusion(type, null, exclusions) != null).ToList();`?

test/ApiCheckBaseline.V1/ComparisonScenarios.cs
```diff
@@ -52,4 +52,9 @@ public class ClassWithMethods
     public class TypeWithExtraInterface
     {
     }
+
+    public interface IInterfaceToAddMembersTo
```
pranavkm: Can we add another scenario for `IInterfaceMemberRenamed`? and another one where the new one has members removed (so old one has more?)

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
Eilon: I believe that would have been covered by the various other checks already.

This check needs to exist because I think at this point we know that everything that was in InterfaceV1 is in InterfaceV2, but we don't know about new stuff in InterfaceV2.

Eilon: (Sorry - I mean it needs to exist *and is sufficient*)

pranavkm: Gotcha. Okies. I'm not super familiar with how the code here is structured so if it looks good to you :shipit:

src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
javiercn: FilterExclusions also modifies the exclusions list, so I don't want to do it in a query style

javiercn:
@pranavkm Addressed feedback

PR #124, opened 2016-11-28
NuGetPackageVerifier as an MSBuild target

natemcmaster:
Wrap NuGetPackageVerifier  in an MSBuild target so it can be included in a build project as a post-pack step.

~~Also upgrades to NuGet 4.0 to support new metadata elements such as `<repository>`~~

cc @pranavkm @ajaybhargavb

### DIFF
--- a/src/NuGetPackageVerifier.Task/VerifyPackages.cs
+++ b/src/NuGetPackageVerifier.Task/VerifyPackages.cs
@@ -1,0 +1,71 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using Microsoft.Build.Framework;
+using MSBuildTask = Microsoft.Build.Utilities.Task;
+
+namespace NuGetPackagerVerifier
+{
+    /// <summary>
+    /// An MSBuild task that acts as a shim into NuGetPackageVerifier.dll
+    /// </summary>
+    public class VerifyPackages : MSBuildTask
+    {
+        [Required]
+        public string RuleFile { get; set; }
+
+        [Required]
+        public string ArtifactDirectory { get; set; }
+
+        public override bool Execute()
+        {
+            if (string.IsNullOrEmpty(RuleFile) || !File.Exists(RuleFile))
+            {
+                Log.LogError("RuleFile '{RuleFile}' does not exist");
+                return false;
+            }
+
+            // prevent trailing slash from breaking the surrounding quotes
+            ArtifactDirectory = ArtifactDirectory?.TrimEnd(new [] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });
+
+            if (string.IsNullOrEmpty(ArtifactDirectory) || !Directory.Exists(ArtifactDirectory))
+            {
+                Log.LogError("ArtifactDirectory '{ArtifactDirectory}' does not exist");
+                return false;
+            }
+
+            var exeExtension = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                ? ".exe"
+                : string.Empty;
+
+            var taskAssemblyFolder = Path.GetDirectoryName(GetType().GetTypeInfo().Assembly.Location);
+            var toolPath = Path.Combine(taskAssemblyFolder, "../../NuGetPackageVerifier.dll");
+            var dotnetPath = new FileInfo(AppContext.GetData("FX_DEPS_FILE") as string)
+                .Directory? // 1.0.1
+                .Parent? // Microsoft.NETCore.App
+                .Parent? // shared
+                .Parent? // DOTNET_HOME
+                .GetFiles("dotnet" + exeExtension)
+                .SingleOrDefault();
+
+            var psi = new ProcessStartInfo
+            {
+                FileName = dotnetPath?.Exists == true
+                    ? dotnetPath.FullName
+                    : "dotnet",  // Fallback to system PATH and hope for the best
+                Arguments = $"\"{toolPath}\" \"{ArtifactDirectory}\" \"{RuleFile}\""
+            };
+
+            Log.LogMessage($"Executing '{psi.FileName} {psi.Arguments}'");
+            var process = Process.Start(psi);
+            process.WaitForExit();
+            return process.ExitCode == 0;
+        }
+    }
+}
--- a/src/NuGetPackageVerifier/NuGetPackageVerifier.nuspec
+++ b/src/NuGetPackageVerifier/NuGetPackageVerifier.nuspec
@@ -8,5 +8,6 @@
   </metadata>
   <files>
     <file src="netcoreapp1.0\" target="/" />
+    <file src="..\..\build\NuGetPackageVerifier.Task\$configuration$\netstandard1.6\NuGetPackageVerifier.Task.dll" target="tools/netstandard1.6/" />
   </files>
 </package>
--- a/src/NuGetPackageVerifier/build/NuGetPackageVerifier.targets
+++ b/src/NuGetPackageVerifier/build/NuGetPackageVerifier.targets
@@ -1,0 +1,4 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <UsingTask AssemblyFile="$(MSBuildThisFileDirectory)..\tools\netstandard1.6\NuGetPackageVerifier.Task.dll"
+               TaskName="NuGetPackagerVerifier.VerifyPackages" />
+</Project>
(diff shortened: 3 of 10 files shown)


src/NuGetPackageVerifier/build/NuGetPackageVerifier.targets
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="VerifyPackages"
+          AfterTargets="Pack"
```
pranavkm: How does this work? Is that a target defined on a per-project basis? If so are we going to install this package in every project?

natemcmaster: Sidebar: project's a bit of an overloaded term in MSBuild world. So, to clarify, translate these terms accordingly:

makefile.shade => 'build project'
project.json => 'code project'


The intent of this target is to be agnostic about whether it runs in a build or code project. That said, the vision for now is to only use this in a build project.

A WIP version of that vision: https://github.com/aspnet/DotNetTools/blob/58d429609db3f4f594f03e0023dce36b99182fb8/build.xml. The goal I'm working towards an MSBuild version of KoreBuild that has many of the same features of KoreBuild. i.e. each repo will only need to define a minimal makefile that extends a default lifecycle.

pranavkm: > The goal I'm working towards an MSBuild version of KoreBuild that has many of the same features of KoreBuild

Maybe we should run this by @anurse. @anurse started work on an alternate build system https://github.com/aspnet/KoreBuild/tree/dev/src/Microsoft.AspNetCore.Build / https://github.com/aspnet/BuildTrain/tree/train/1.0.0 a while ago but we never got around to switching to it. I'd rather we not invest in any work until we have some consensus from the engineering team on what we want here.

natemcmaster:
Bummer. Consensus is a long way out. I was hoping to at least prototype this approach. I figured this is minimal risk as it's not a breaking change and is merely a shim for MSbuild into calling NuGetPackageVerifier.exe.

I'm going to split out the upgrade to NuGet 4.0.0 into a separate PR as, regardless of KoreBuild's future, NuGetPackageVerifier.exe dies when trying to verify packages created via NuGet's new pack targets for MSBuild.

pranavkm:
:( If this unblocks some prototyping, consider using a local build of this package. The package upgrade seems fine.

src/NuGetPackageVerifier/build/NuGetPackageVerifier.targets
pranavkm: Going by @Eilon \ @davidfowl's comments in https://github.com/aspnet/specs/pull/49, I think it might be fine to unblock this. 

That said, instead of having executables that get run, would it make sense to just these be MSBuild tasks? It would make things like logging and reporting errors much nicer (TeamCity has first class support for desktop msbuild loggers, we might need to port these for msbuild core). Additionally, it would also decouple tools from the life cycle. We could have tool target files export msbuild tasks and have the importing project compose these. Having the lifecycle spread out between the build project + individual tools might be harder to manage and tedious to debug if they never get invoked.

pranavkm: For starters though, we could just have these export targets (that launch the exe) and make the switch to tasks later on. The composition model is the one I'm concerned about.

natemcmaster: > instead of having executables that get run, would it make sense to just these be MSBuild tasks?

Yes,I was thinking the same thing too. It will take more effort to port this to a task, and one we'll have to do carefully. I expect we will run into assembly loading issues with MSBuild for .NET Core. It's default load context may already have NuGet assemblies.

> Having the lifecycle spread out between the build project + individual tools might be harder to manage and tedious to debug if they never get invoked.

A reasonable concern, but one I'm sure we have the brainpower to tackle. There are already some good tools for debugging MSBuild. https://github.com/KirillOsenkov/MSBuildStructuredLog is a good example.

pranavkm: But isn't it easier to know something is wrong when `CallTarget` or `DependsOnTargets` fails because the target you asked for does not exist versus discovering we weren't running a tool because we renamed a target? Debugging is helpful when it fails with a noise, but target chaining fails pretty silently.

natemcmaster: If your main concern is that verifier won't run and bad packages will get out, how about we do something simpler:
 - Pack creates *.nupkg files in an intermediate location, such as `/obj/packages/`
 - VerifyPackages moves *.nupkg files to `/artifacts/build/` if and only if they pass verification steps.

But regarding target chaining, we could use this pattern instead:

```xml
<!-- in Lifecycle.targets -->
<Target Name="Pack" DependsOnTargets="$(PackDependsOn)" />

<!-- in NugetPackageVerifier.props -->
<PropertyGroup>
   <PackDependsOn>$(PackDependsOn);VerifyPackages</PackDependsOn>
</PropertyGroup>

<!-- in NugetPackageVerifier.targets -->
<Target Name="VerifyPackages" />
```

pranavkm: This really doesn't lend to using the tool in uncommon scenarios. For instance, we run the tool as part of Coherence-Signed today (and we used to run it as part of the the partner package mirroring). Relying on tool name as a target seems less magical than relying on the tool putting together the lifecycle.

natemcmaster: Hmm, that's a usage I had not considered. I think maybe this PR is the wrong approach.  Instead of adding NuGetPackageVerifier as a target, it should be task. This task becomes available by taking a PackageDependency on NuGetPackageVerifier. But integration into the build is up to the consumer.

E.g. consumer usage
```xml
<!-- in Coherence-Signed.targets or KoreBuild.targets -->
<Project>
   <ItemGroup>
      <PackageDependency Include="NuGetPackageVerifier" Version="1.0.0" />
   </ItemGroup>

   <Target Name="VerifyPackages" DependsOnTargets="Pack">
      <NuGetPackageVerifier RuleFile="NPV.json" Directory="artifacts/build" />
   </Target>
</Project>
```

The problem, of course, is that you can't define a task without writing some C#. Last I heard, MSBuild for .NET Core doesn't support code tasks yet, so we'd have to compile/package a task. To keep it simple for now, though, that task can be super minimal and just shell execute NugetPackageVerifier.exe.

Thoughts?

pranavkm: >  To keep it simple for now, though, that task can be super minimal and just shell execute NugetPackageVerifier.exe.

👍 

The other alternative is to rely on well known property names until we switch to a task:

```xml
<!-- NuGetPackageVerifier.targets -->
<Target Name="NuGetPackageVerifier">
   <Error Message="_NPVRuleFile not set" Condition="'$(_NPVRuleFile)' = ''" />
   <Error Message="_NPVDirectory not set" Condition="'$(_NPVDirectory)' = ''" />
  <Exec Command="$(MsBuildThisFileDirectory)/NPE.exe --directory $(_NPVDirectory) --file $(_NPVRuleFile)" />
</Target>

<!-- in Coherence-Signed.targets or KoreBuild.targets -->
<PackageDependsOn>$(PackageDependsOn);VerifyPackages
<VerifyPackages>
   <MsBuild Projects="$(MsBuildThisProject)" Properties="$(_NPVRuleFile)=..;$(_NPVDirectory)=..." Targets="NuGetPackageVerifier" />
 </VerifyPackages>
```

src/NuGetPackageVerifier/NuGetPackageVerifier.nuspec
```diff
@@ -2,13 +2,15 @@
...
     <description>Verifies Asp.Net Core packages.</description>
   </metadata>
   <files>
-    <file src="build\*.targets" target="build\" />
-    <file src="*.dll" target="\" />
-    <file src="*.exe" target="\" />
+    <file src="net451\build\NuGetPackageVerifier.targets" target="build/" />
+    <file src="net451\*.dll" target="/" />
+    <file src="net451\*.exe" target="/" />
+    <file src="..\..\build\NuGetPackageVerifier.Task\$configuration$\net46\NuGetPackageVerifier.Task.dll" target="tools/net46/" />
```
pranavkm: Think this is gone now? @ajaybhargavb ?

natemcmaster:
I'm trying to validate this works. Does NGPV.exe invoke itself? This is the output I get when I invoke the task.

```
NORMAL: Read 1 package set(s) from rule.json
NORMAL: Found 8 packages in ..\..\artifacts\build

Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at NuGetPackageVerifier.Program.Main(String[] args)
```

natemcmaster:
🆙 :date: Fixed the NRE (was unrelated to the changes I was making).

natemcmaster:
Rebased on Ajay's latest work.

src/NuGetPackageVerifier.Task/VerifyPackages.cs
```diff
@@ -0,0 +1,71 @@
...
+                .Parent? // DOTNET_HOME
+                .GetFiles("dotnet" + exeExtension)
+                .SingleOrDefault();
+
+            var psi = new ProcessStartInfo
+            {
+                FileName = dotnetPath?.Exists == true
+                    ? dotnetPath.FullName
+                    : "dotnet",  // Fallback to system PATH and hope for the best
+                Arguments = $"\"{toolPath}\" --verbose \"{ArtifactDirectory}\" \"{RuleFile}\""
```
natemcmaster: Good point. I had it enabled for debugging, but it's probably not necessary.

PR #125, opened 2016-11-28
Upgrade NuGetPackageVerifier to NuGet 4.0.0-rc-2048

natemcmaster:
NuGet 3.5.0-beta2 fails if the metadata includes the new `<repository>` element that will be added to the nuspec when using NuGet 4 pack targets (aka dotnet-CLI for MSBuild).

cc @prafullbhosale @pranavkm


pranavkm:
cc @ajaybhargavb 

:shipit:

PR #126, opened 2016-11-30
Generate package cache with the transitive closure of our dependencies except the shared runtime

javiercn:
* Include all the dependencies in the package cache with the exception of the shared framework.
* Use Microsoft.NetCore.App instead of NETStandard to exclude depencencies.
* Avoid removing files from the temporary restore folder, copy required files to the
  destination instead.
* Do not include an intermediate publish step for running crossgen on the assemblies,
  use a response file with the semi-colon separated list of paths to the application
  and platform assemblies.
* Write crossgen output to a file instead of the console.
* Simplify and revamp diagnostics:
    * Print the list of files on the package cache.
    * Print the list of assemblies for which crossgen failed.
* Include an exclusion list with the tool to avoid failing when we do not crossgen an
  assembly successfully.


javiercn:
@natemcmaster

src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -192,22 +365,25 @@ private DirectoryInfo CreateSubDirectory(string cacheBasePath, PackageEntry pack
...
-            var publishFolder = _context.GetStagingFolderPath(parent);
-
-            if (!Directory.Exists(publishFolder))
-            {
-                _logger.LogInformation($"Creating directory {publishFolder}.");
-                Directory.CreateDirectory(publishFolder);
-            }
+            _responseFilePath = Path.Combine(_destinationFolder, "aspnet-crossgen.rsp");
+            var builder = new StringBuilder();
+            builder.AppendLine($"/Platform_Assemblies_Paths \"{string.Join(";", _referenceAssemblyPaths)}\"");
```
natemcmaster: IIRC a response file usually requires one line per argument (at least according to System.CommandLine), but each tool handles rsp slightly different

src/DependenciesPackager/PackageRestoreContext.cs
```diff
@@ -105,11 +64,26 @@ private int RunDotnetRestore()
...
             }
 
-            return new ProcessRunner(_dotnetSdkPath, arguments)
+            var exitCode = new ProcessRunner(_dotnetSdkPath, arguments)
                 .AddEnvironmentVariable("NUGET_XMLDOC_MODE", "Skip")
                 .WriteOutputToConsole()
                 .WriteOutputToConsole()
                 .Run();
+
+            if (exitCode != 0)
```
natemcmaster: reduce to `return exitCode == 0`

src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -3,72 +3,130 @@
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
 using Microsoft.DotNet.ProjectModel;
```
natemcmaster: Out of scope for this PR, but have you looked into doing the NuGet math without `Microsoft.DotNet.ProjectModel` APIs? They were deprecated along with the rest of project.json

src/DependenciesPackager/CrossgenContext.cs
```diff
@@ -110,78 +182,179 @@ public void FetchJitTools(string runtime, string restoreFolder)
...
+
                 foreach (var asset in package.Assets)
                 {
-                    var subDir = CreateSubDirectory(outputPath, package, asset);
+                    var subDir = CreateAssetSubDirectory(outputPath, package, asset);
                     var targetPath = Path.Combine(subDir.FullName, asset.FileName);
 
-                    if (!CrossgenAssembly(_stagingFolder, package, asset, targetPath))
+                    var crossGen = CrossgenAssembly(package, asset, targetPath);
+                    lock (lck)
```
javiercn: Just to print them in order, if not they can get intermingled and it makes the output hard to read (was happening on the console when I made it parallel))

src/DependenciesPackager/CrossgenContext.cs
javiercn: You mean because of the string.Join()?  It's the first time i'm writing a response file, It probably works with all in new lines although I see `/flag parameters` as the argument.  I say we keep it this way, it doesn't make a big difference to have 3 lines vs 6 lines on the file

natemcmaster: Actually, my concern was about argument escaping with spaces and backslashes. It is not just enough to surround arguments containing spaces with quotes. You also have to examine if an argument contains spaces and backlashes. See https://github.com/aspnet/DotNetTools/blob/e0f85971a32c21f4246b41ffabfca60e039b13f4/src/Shared/ArgumentEscaper.cs for example.

src/DependenciesPackager/CrossgenContext.cs
javiercn: Not yet, definitely something to look into. https://github.com/aspnet/BuildTools/issues/127

javiercn:
@natemcmaster Updated

src/DependenciesPackager/exclusions.txt
```diff
@@ -0,0 +1,2 @@
+﻿System.Net.Http.Formatting.dll
```
natemcmaster: Let's put this in Coherence-final instead

Issue #103, opened 2016-09-28
Make NuGetPackageVerifier target netcoreapp1.0


pranavkm:
The verifier has dependencies which make it desktop only. We should figure out if there are alternatives we could use to make it xplat compatible.

PR #128, opened 2016-12-05
Make NuGetPackageVerifier target netcoreapp

ajaybhargavb:
Issue - #103 

@pranavkm @natemcmaster @Eilon 

Korebuild needs to be updated for this change. I'll update here with the link once that's out.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -34,7 +34,7 @@ k-standard-goals
 
       var version = Path.GetFileNameWithoutExtension(existingNupkg).Substring(project.Length + 1);
       var nuspecPath = Path.Combine("src", project, project + ".nuspec");
-      var basePath = Path.Combine(outputDir, project, configuration, "net451", "win7-x64");
+      var basePath = Path.Combine(outputDir, project, configuration);
 
       if (project == "DependenciesPackager")
       {
--- a/src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
+++ b/src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
@@ -4,9 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Reflection;
-using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
+using System.Linq;
+using System.Runtime.Loader;
+using Mono.Cecil;
 using NuGet.Packaging;
 using NuGetPackageVerifier.Logging;
 
@@ -45,22 +45,23 @@ namespace NuGetPackageVerifier.Rules
                             if (AssemblyHelpers.IsAssemblyManaged(assemblyPath))
                             {
                                 isManagedCode = true;
-                                var clrStrongName = (IClrStrongName)RuntimeEnvironment.GetRuntimeInterfaceAsObject(
-                                    new Guid("B79B0ACD-F5CD-409b-B5A5-A16244610B92"),
-                                    new Guid("9FD93CCF-3280-4391-B3A9-96E1CDE77C8D"));
-
-                                bool verificationForced;
-                                hresult = clrStrongName.StrongNameSignatureVerificationEx(assemblyPath, true, out verificationForced);
-                                if (hresult == 0)
+                                using (var assembly = AssemblyDefinition.ReadAssembly(assemblyPath))
                                 {
-                                    isStrongNameSigned = true;
-                                }
+                                    if (assembly.Modules.Any())
+                                    {
+                                        isStrongNameSigned = assembly.Modules.All(
+                                            module => module.Attributes.HasFlag(ModuleAttributes.StrongNameSigned));
+                                    }
+                                    else
+                                    {
+                                        throw new InvalidOperationException("The managed assembly does not contain any modules.");
+                                    }
 
-                                var testAssembly = AssemblyName.GetAssemblyName(assemblyPath);
-                                var tokenHexString = BitConverter.ToString(testAssembly.GetPublicKeyToken()).Replace("-", "");
-                                if (_publicKeyToken.Equals(tokenHexString, StringComparison.OrdinalIgnoreCase))
-                                {
-                                    hasCorrectPublicKeyToken = true;
+                                    var tokenHexString = BitConverter.ToString(assembly.Name.PublicKeyToken).Replace("-", "");
+                                    if (_publicKeyToken.Equals(tokenHexString, StringComparison.OrdinalIgnoreCase))
+                                    {
+                                        hasCorrectPublicKeyToken = true;
+                                    }
                                 }
                             }
                         }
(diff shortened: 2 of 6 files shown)


src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
```diff
@@ -10,6 +10,7 @@
 using NuGet.Packaging;
 using NuGetPackageVerifier.Logging;
 
+#if NET451
```
pranavkm: Apparently we might be able to do this using Mono.Cecil. According to http://thread.gmane.org/gmane.comp.gnome.mono.cecil/3315, we could use their introspection APIs to determine if an assembly is strong named (including determining if it's delay signed). 

If we can

a. empirically prove that this works (try this with non-strong named, delay signed and strong name signed binaries on binaries produced on Windows + xplat)
b. @Eilon is happy with us not using a blessed API for strong name verification

We should do it. It would also mean we wouldn't have to cross compile this binary.

natemcmaster: In theory also possible with System.Reflection.Metadata

src/NuGetPackageVerifier/project.json
```diff
@@ -9,7 +9,7 @@
     "HtmlAgilityPack": "1.4.9",
     "Newtonsoft.Json": "9.0.1",
     "NuGet.Packaging": "4.0.0-rc-2048",
-    "Mono.Cecil": "0.9.5.4"
+    "Mono.Cecil": "0.10.0-beta1-v2"
```
ajaybhargavb: The beta version supports netcoreapp

natemcmaster: nvm. looks like this update is required for netstandard support.

src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
pranavkm: We're already using Cecil to do everything else here so I'd rather stick to it.

ajaybhargavb:
🆙 📅. I am currently verifying if the rule works for all cases that @pranavkm suggested. Will also wait for @Eilon to look at this before merging.

src/NuGetPackageVerifier/project.json
```diff
@@ -9,18 +9,18 @@
...
-      }
+    "netcoreapp1.1": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "version": "1.1.0-*",
+          "type": "platform"
+        },
+        "System.Runtime.Loader": "4.3.0"
+      },
+      "imports": [ "portable-net45+wp8" ]
```
natemcmaster: Just curious: what is this required for?

natemcmaster:
Changes LGTM.

src/NuGetPackageVerifier/project.json
ajaybhargavb: This is needed to make `HtmlAgilityPack` compatible

src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
```diff
@@ -45,18 +44,18 @@ public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context
...
                                 isManagedCode = true;
-                                var clrStrongName = (IClrStrongName)RuntimeEnvironment.GetRuntimeInterfaceAsObject(
-                                    new Guid("B79B0ACD-F5CD-409b-B5A5-A16244610B92"),
-                                    new Guid("9FD93CCF-3280-4391-B3A9-96E1CDE77C8D"));
-
-                                bool verificationForced;
-                                hresult = clrStrongName.StrongNameSignatureVerificationEx(assemblyPath, true, out verificationForced);
-                                if (hresult == 0)
+                                isStrongNameSigned = true;
+                                var assembly = AssemblyDefinition.ReadAssembly(assemblyPath);
```
ajaybhargavb: We need to write the file to disk for `AssemblyLoadContext.GetAssemblyName()` to work. So I'd say it doesn't matter here.

ajaybhargavb:
🆙 📅

src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
```diff
@@ -45,22 +45,23 @@ public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context
                             if (AssemblyHelpers.IsAssemblyManaged(assemblyPath))
```
pranavkm: What happens when you call `ReadAssembly` on a non-managed assembly? I'd think it throws something useful.

src/NuGetPackageVerifier/WinTrust.cs
```diff
@@ -93,23 +93,24 @@ public static bool IsAuthenticodeSigned(string path)
         {
             var fileInfo = new WinTrustFileInfo()
```
pranavkm: Does this work in xplat? The DllImport seems suspect

src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
pranavkm: Could we use that to infer things? That way we wouldn't need to extract files to disk.

ajaybhargavb: 👍 will do

src/NuGetPackageVerifier/WinTrust.cs
pranavkm: We're generally not going to be Authenticode signing in dev workflows. So it might be fine to explicitly fail here if we're not on Windows.

src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
ajaybhargavb: I tried doing this but `AssemblyDefinition.ReadAssembly(Stream)` does not seem to work. It throws a `System.ArgumentException: Value does not fall within the expected range.` But works when an actual file path is specified. I googled it and it seems like a bug in cecil but not very sure.

pranavkm: Oh well. Was worth a try.

ajaybhargavb:
Ping @Eilon

makefile.shade
```diff
@@ -34,7 +34,7 @@ k-standard-goals
 
       var version = Path.GetFileNameWithoutExtension(existingNupkg).Substring(project.Length + 1);
       var nuspecPath = Path.Combine("src", project, project + ".nuspec");
-      var basePath = Path.Combine(outputDir, project, configuration, "net451", "win7-x64");
+      var basePath = Path.Combine(outputDir, project, configuration);
```
natemcmaster: Doesn't this mess up the packaging for some of the other tools? 

FWIW I refactored this part of the build script in #124

ajaybhargavb: This doesn't affect any other tool as of this PR. But you'll need to update #124 to not target net451 once this goes in.

ajaybhargavb: ^ if @Eilon approves it that is.

Eilon: Sorry, don't know what this is...

ajaybhargavb: Also, this change here doesn't matter because as @natemcmaster said, this has been refactored in #124.

src/NuGetPackageVerifier/Rules/AssemblyStrongNameRule.cs
```diff
@@ -45,22 +45,23 @@ public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context
...
                                 {
-                                    hasCorrectPublicKeyToken = true;
+                                    if (assembly.Modules.Any())
+                                    {
+                                        isStrongNameSigned = assembly.Modules.All(
+                                            module => module.Attributes.HasFlag(ModuleAttributes.StrongNameSigned));
+                                    }
+                                    else
+                                    {
+                                        throw new InvalidOperationException("The managed assembly does not contain any modules.");
```
Eilon: I was worried here about the case of satellite resource DLLs and whether or not they have at least one module, but it seems like they always do have one, so no worries here. 🙈

Issue #116, opened 2016-11-04
Add a --quiet option to disable logging info messages in NuGetPackageVerifier


ajaybhargavb:
Currently there is no support to hide Info log messages. It would be better to have the verifier just log messages for errors and warnings as part of the build.

Eilon:
Which tool is this about?

Eilon:
Assuming it's NuGetPackageVerifier, please go ahead and add it. And update the bug title 😄

PR #131, opened 2016-12-08
Added a --quiet option to NuGetPackageVerifier to hide Info log messages

ajaybhargavb:
Issue - #116

### DIFF
--- a/src/NuGetPackageVerifier/Program.cs
+++ b/src/NuGetPackageVerifier/Program.cs
@@ -22,9 +22,17 @@ namespace NuGetPackageVerifier
         public static int Main(string[] args)
         {
             // TODO: Show extraneous packages, exclusions, etc.
+            var ignoreAssistanceMode = IgnoreAssistanceMode.None;
 
             // TODO: Get this from the command line
-            var ignoreAssistanceMode = IgnoreAssistanceMode.ShowAll;
+            var hideInfoLogs = true;
+
+            if (args.Length > 0 && string.Equals("--verbose", args[0], StringComparison.OrdinalIgnoreCase))
+            {
+                hideInfoLogs = false;
+                ignoreAssistanceMode = IgnoreAssistanceMode.ShowAll;
+                args = args.Skip(1).ToArray();
+            }
 
             if (args.Length < 1 || args.Length > 2)
             {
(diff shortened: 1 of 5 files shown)


src/NuGetPackageVerifier/Program.cs
```diff
@@ -25,6 +25,13 @@ public static int Main(string[] args)
 
             // TODO: Get this from the command line
             var ignoreAssistanceMode = IgnoreAssistanceMode.ShowAll;
+            var hideInfoLogs = false;
+
+            if (args.Length > 0 && string.Equals("--quiet", args[0], StringComparison.OrdinalIgnoreCase))
```
pranavkm: Not super vested in this, but do we want to make the default `Normal` and hide the random spew we get today `Info`?

```
N: Read 2 package set(s) from C:\BuildAgent\work\172afee273357bca\NuGetPackageVerifier.json 
N: Found 1 packages in C:\BuildAgent\work\172afee273357bca\artifacts\build 
I: Processing package set 'adx' with 1 package(s)
I: Analyzing Microsoft.AspNetCore.Antiforgery (1.2.0-t003a5150a)
I: No issues found with package Microsoft.AspNetCore.Antiforgery (1.2.0-t003a5150a)

I: Processing package set 'Default' with 0 package(s)
I: Showing JSON for ignore content: < -- Info when no error, else, normal
I: {}

N: SUMMARY: 0 error(s) and 0 warning(s) found
N: Total took 568ms
```

pranavkm: Also, can we just make this `--log-level`?

ajaybhargavb: > do we want to make the default Normal and hide the random spew we get today Info?

Does this mean we'll have four log levels instead of three?

> can we just make this --log-level?

I personally feel having a min log level switch is unnecessary at this point and we can add it later if we need it. But I don't feel strongly about this either way. Let me know if you think it is necessary.

pranavkm: Ideally all of this goes away if \ when we make this an MSBuild task (which is why I said I wasn't super vested in this). If you think this works, we'll live with it 👍

ajaybhargavb: Ah I see. This works for now. But I think it would be nice to have the SUMMARY and Total time taken not hidden. How about we have them as info and make the rest as Trace or Verbose? And --quiet will hide the Trace/Verbose messages

pranavkm: Sure. The `N` at the start of the line was for normal, the `I` was for Info. So we would have a couple of visible text.

ajaybhargavb:
🆙 📅

ajaybhargavb:
🆙 📅 with another small change discussed offline

PR #132, opened 2016-12-08
DependenciesPackager: add preserve case flag 

natemcmaster:
For creating the Docker legacy package cache.


natemcmaster:
We need it for the aspnet core Docker images. This can go away once we completely drop support for apps built with project.json...but that probably won't happen soon.

PR #139, opened 2016-12-21
Removed unnecessary Console.WriteLine()

ajaybhargavb:
@BrennanConroy


BrennanConroy:
What was it for?

BrennanConroy:
🐑 :it:

ajaybhargavb:
It was there to just make things pretty. We don't need that anymore because we only print things when `--verbose` is passed in.

PR #138, opened 2016-12-13
Upgrade to VS 2017

natemcmaster:
Port this project to VS 2017.

cc @ajaybhargavb @javiercn

PR #143, opened 2017-01-06
Unify tools version to 1.0.2

natemcmaster:
Stop cross-compiling SplitPackages for net451.

Will require some changes to KoreBuild to consume newer versions and change the invocation for netcoreapp1.0 tools.

Follow up to #138 

cc @jbagga


natemcmaster:
Ping @pranavkm

natemcmaster:
Has diverged from dev. Will revisit later.

PR #144, opened 2017-01-07
Add rule for pre-release dependencies for RTM packages

jbagga:
Addresses https://github.com/aspnet/Coherence-Signed/issues/396

Work in progress for adding a `NuGetPackageVerifier` execution in [Coherence-Signed](https://github.com/aspnet/Coherence-Signed/blob/dev/dnx.msbuild).

cc @pranavkm @Eilon


src/NuGetPackageVerifier/Rules/DependenciesVersionRule.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Collections.Generic;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class DependenciesVersionRule : IPackageVerifierRule
+    {
+        public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
+        {
+            if (!context.Metadata.Version.IsPrerelease)
```
pranavkm:
```
if (context.Metadata.Version.IsPrerelease)
{
   yield break;
}
```

src/NuGetPackageVerifier/CompositeRules/PackageDependenciesCompositeRule.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class PackageDependenciesCompositeRule : IPackageVerifierRule
```
pranavkm: Remove this type and add the new rule to `DefaultCompositeRule`.

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -259,5 +259,20 @@ public static PackageVerifierIssue IdIsNotOwned(string id, string[] allowedOwner
         {
             return new PackageVerifierIssue("PACKAGE_OWNERSHIP", id, $"The id '{id}' is not owned by one of {string.Join(", ", allowedOwners)}.", PackageIssueLevel.Error);
         }
+
+        public static PackageVerifierIssue DependencyVersionHasUpperBound(string id, string dependencyId)
+        {
+            return new PackageVerifierIssue("PACKAGE_DEPENDENCY", id, $"The version range of package dependency '{dependencyId}' for package '{id}' has an upper bound.", PackageIssueLevel.Warning);
```
pranavkm: Think the id should be `PACKAGE_DEPENDENCY_VERSION_UPPER_BOUND`. We need to be able to exclude it from the verifier and consequently uniquely identify it

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -259,5 +259,20 @@ public static PackageVerifierIssue IdIsNotOwned(string id, string[] allowedOwner
...
+        }
+
+        public static PackageVerifierIssue DependencyVersionDoesNotHaveLowerBound(string id, string dependencyId)
+        {
+            return new PackageVerifierIssue("PACKAGE_DEPENDENCY", id, $"The version range of package dependency '{dependencyId}' for package '{id}' does not have a lower bound.", PackageIssueLevel.Warning);
```
pranavkm: Same comment as above.

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -259,5 +259,20 @@ public static PackageVerifierIssue IdIsNotOwned(string id, string[] allowedOwner
...
+        }
+
+        public static PackageVerifierIssue DependencyVersionIsPrereleaseForRTMPackage(string id, string dependencyId)
+        {
+            return new PackageVerifierIssue("PACKAGE_DEPENDENCY", id, $"The package dependency'{dependencyId}' for package '{id}' is a prerelease package version.", PackageIssueLevel.Error);
```
pranavkm: Can you add more details to error? Like stating that a non-prerelease package is depending on a prerelease package

src/NuGetPackageVerifier/CompositeRules/PackageDependenciesCompositeRule.cs
Eilon: This rule needs to specifically run *after* we do the timestamp/notimestamp split, so I'm pretty sure it needs to go into a new composite rule. However, I think the new composite rule should still have a more generic name, such as `NoTimestampPackagesCompositeRule` or something.

src/NuGetPackageVerifier/CompositeRules/PackageDependenciesCompositeRule.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        IPackageVerifierRule[] _rules = new IPackageVerifierRule[]
```
Eilon: `private`.

src/NuGetPackageVerifier/PackageIssueFactory.cs
Eilon: Or specifically state the actual problem: `The RTM package '{fooName}' ${fooVersion} cannot depend on a pre-release package {barName}' ${barVersion}.`

src/NuGetPackageVerifier/Rules/DependenciesVersionRule.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                foreach (var dependencyGroup in context.Metadata.DependencyGroups)
+                {                    
+                    foreach (var packageDependency in dependencyGroup.Packages)
+                    {
+                        if (packageDependency.VersionRange.HasUpperBound)
```
Eilon: Are these checks required prior to checking `MinVersion.IsPrerelease`? Or what happens if we omit this check?

jbagga: From my discussion with @pranavkm if the package does not have a specified minimum version or has a maximum version, it has not been declared according to the convention we follow (>= packageVersion)

Eilon: Ah, ok, that's fine then.

src/NuGetPackageVerifier/CompositeRules/PackageDependenciesCompositeRule.cs
pranavkm: Isn't it generally problematic if we have RTM packages relying on prerelease packages?

Eilon: Yes... of course. But if we run it on "timestamp" packages then our packages are always pre-release because they are `x.y.z-rtm-12345` so this rule will never check anything.

Eilon: Actually, I wonder if the upper-bound/lower-bound thing should be a separate rule that runs per-repo. And then the most important rule about rel/pre-rel deps would be the only one in this new composite rule.

jbagga:
@Eilon @pranavkm Updated

src/NuGetPackageVerifier/CompositeRules/PackageDependenciesCompositeRule.cs
jbagga: @Eilon @pranavkm Does `Target Name="NuGetVerify"` in the [dnx.msbuild file](https://github.com/aspnet/Coherence-Signed/blob/dev/dnx.msbuild#L172) run all the rules in `NuGetPackageVerifier`? Since we only call this once after rebuilding packages and creating timestamp free packages, why not have this new pre-rel package rule in the `DefaultCompositeRule`? If this is not what's happening, how/where in the msbuild file can one specify which composite rule to run? And in that case I think we will need more than one execution of `NuGetVerify`.

jbagga: So I found [this](https://github.com/aspnet/Coherence-Signed/blob/dev/tools/NugetPackageVerifier.json). Since it isn't running the `DefaultCompositeRule`, for now I am adding the new composite rule with just the pre-rel package rule to the list at the end of this file.

src/NuGetPackageVerifier/CompositeRules/NoTimestampPackagesCompositeRule.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System.Collections.Generic;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class NoTimestampPackagesCompositeRule : IPackageVerifierRule
+    {
+        private IPackageVerifierRule[] _rules = new IPackageVerifierRule[]
+        {
+            new PrereleaseDependenciesVersionRule()
```
pranavkm: Could we add it to https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/CompositeRules/SigningVerificationCompositeRule.cs if we really want to avoid making it part of the default ruleset?

jbagga: @Eilon See [this](https://github.com/aspnet/BuildTools/pull/144#discussion_r95694636). Currently `SigningVerificationCompositeRule` is the only one in the `NugetPackageVerifier.json` file.

src/NuGetPackageVerifier/CompositeRules/PackageDependenciesCompositeRule.cs
Eilon: @jbagga let's chat in person.

jbagga:
@pranavkm @Eilon Updated.

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -259,5 +260,20 @@ public static PackageVerifierIssue IdIsNotOwned(string id, string[] allowedOwner
...
         }
+
+        public static PackageVerifierIssue DependencyVersionHasUpperBound(string id, string dependencyId)
+        {
+            return new PackageVerifierIssue("PACKAGE_DEPENDENCY_VERSION_UPPER_BOUND", id, $"The version range of package dependency '{dependencyId}' for package '{id}' has an upper bound.", PackageIssueLevel.Warning);
+        }
+
+        public static PackageVerifierIssue DependencyVersionDoesNotHaveLowerBound(string id, string dependencyId)
+        {
+            return new PackageVerifierIssue("PACKAGE_DEPENDENCY_VERSION_LOWER_BOUND", id, $"The version range of package dependency '{dependencyId}' for package '{id}' does not have a lower bound.", PackageIssueLevel.Warning);
```
pranavkm: Add TFMS here?

src/NuGetPackageVerifier/CompositeRules/SigningVerificationCompositeRule.cs
```diff
@@ -16,6 +16,7 @@ public class SigningVerificationCompositeRule : IPackageVerifierRule
             new RequiredNuSpecInfoRule(),
             new PackageOwnershipRule(),
             new DefaultCompositeRule(),
+            new PrereleaseDependenciesVersionRule()
```
Eilon: Trailing comma please,

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Reflection;
 using NuGet.Versioning;
+using NuGet.Frameworks;
```
Eilon: Sort

PR #147, opened 2017-01-17
Add common external package version variables

BrennanConroy:
@mikaelm12 @pranavkm @natemcmaster


natemcmaster:
What will be the corresponding change for something that consumes the Internal.AspNetCore.Sdk project, like aspnet/Common?

pranavkm:
@natemcmaster - we're going to update individual repos to include this value. I guess we can do that once this package builds.

mikaelm12:
😄

natemcmaster:
I don't think that'll work. The package isn't installed when you run restore for the first time.

natemcmaster:
Recommend either creating Internal.AspNetCore.Test.Sdk which depends on the xunit version we need, or waiting until https://github.com/Microsoft/msbuild/issues/1493 is ready

PR #150, opened 2017-01-19
Resolve RuntimeFrameworkVersion from version of Microsoft.NETCore.App

natemcmaster:
Resolve RuntimeFrameworkVersion from version of Microsoft.NETCore.App, not TFM

Workaround dotnet/sdk#686


src/Internal.AspNetCore.Sdk/build/Workarounds.targets
```diff
@@ -0,0 +1,35 @@
+<!--
+WARNING: These targets are intended for building Microsoft's ASP.NET Core repos, and is not intended
+for use outside of Microsoft.
+
+This file is dedicated to working around issues in Microsoft.NET.Sdk
+-->
+<Project ToolsVersion="15.0">
+  <!-- workaround https://github.com/dotnet/sdk/issues/686 -->
+  <Target Name="SetRuntimeFrameworkVersionToPackageVersion"
+          Condition="'$(OutputType)'=='Exe' AND '$(TargetFrameworkIdentifier)'=='.NETCoreApp'"
```
BrennanConroy: Just `.NETCoreApp` is fine?

natemcmaster: Just being safe. I don't believe we have .NETCoreApp libraries.

BrennanConroy:
:shipit:

PR #151, opened 2017-01-23
Add packages to already-owned-packages.txt

jbagga:
@pranavkm


pranavkm:
Could you add `Microsoft.AspNetCore.Mvc.Razor.ViewCompilation` to the list and check this in?

PR #156, opened 2017-02-02
Patch version update

JunTaoLuo:
https://github.com/aspnet/Coherence/issues/147

TODO:
- [x] Make updates command line args that can specify repo or package
- [x] Update packages baseline

### DIFF
--- a/src/VersionTool/Program.cs
+++ b/src/VersionTool/Program.cs
@@ -1,12 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Text.RegularExpressions;
-using Microsoft.DotNet.ProjectModel;
-using Microsoft.Extensions.CommandLineUtils;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
+﻿using Microsoft.Extensions.CommandLineUtils;
 
 namespace VersionTool
 {
@@ -22,7 +14,7 @@ namespace VersionTool
 
                 c.HelpOption("-h|--help");
 
-                c.OnExecute(() => OnList(pathOption));
+                c.OnExecute(() => ListCommand.Execute(pathOption));
             });
 
             var updateVersionCommand = app.Command("update-version", (c) =>
@@ -34,7 +26,7 @@ namespace VersionTool
 
                 c.HelpOption("-h|--help");
 
-                c.OnExecute(() => OnUpdateVersion(pathOption, matchingOption, versionArgument));
+                c.OnExecute(() => UpdateVersionCommand.Execute(pathOption, matchingOption, versionArgument));
             });
 
             var listDepdendency = app.Command("list-dependency", (c) =>
@@ -46,7 +38,7 @@ namespace VersionTool
 
                 c.HelpOption("-h|--help");
 
-                c.OnExecute(() => OnListDependency(pathOption, matchingOption, dependencyArgument));
+                c.OnExecute(() => ListDependencyCommand.Execute(pathOption, matchingOption, dependencyArgument));
             });
 
             var updateDependency = app.Command("update-dependency", (c) =>
@@ -59,9 +51,22 @@ namespace VersionTool
 
                 c.HelpOption("-h|--help");
 
-                c.OnExecute(() => OnUpdateDependency(pathOption, matchingOption, dependencyArgument, versionArgument));
+                c.OnExecute(() => UpdateDependencyCommand.Execute(pathOption, matchingOption, dependencyArgument, versionArgument));
             });
 
+            var updatePatch = app.Command("update-patch", (c) =>
+            {
+                var pathOption = c.Option("-d|--directory", "Directory containing all repos", CommandOptionType.SingleValue);
+                var updatePatchConfigOption = c.Option("-u|--update-patchconfig", "Update patchconfig with all packages found and their current versions", CommandOptionType.NoValue);
+                var updatePackageOption = c.Option("-p|--update-package", "Update the given package before patch following the format <package>:<version>", CommandOptionType.MultipleValue);
+                var updateRepoOption = c.Option("-r|--update-repo", "Update the packages in the given repo before patch ", CommandOptionType.SingleValue);
+
+                var patchConfig = c.Argument("patchConfig", "Path to the configuration file for patch update");
+
+                c.HelpOption("-h|--help");
+
+                c.OnExecute(() => UpdatePatchCommand.Execute(pathOption, updatePatchConfigOption, updatePackageOption, updateRepoOption, patchConfig));
+            });
 
             app.HelpOption("-h|--help");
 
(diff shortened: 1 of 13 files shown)


src/VersionTool/ProcessRunner.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace VersionTool
+{
+    internal class ProcessRunner
```
pranavkm: If it's not super painful -  could you use the `Command` type from `https://www.nuget.org/packages/Microsoft.DotNet.Cli.Utils/1.0.0-preview2-003121`? It's pretty functional and avoids having boiler plate wrapper code for `Process`.

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
                 {
                     SaveJObject(project.ProjectFilePath, root);
 
-                    Console.WriteLine($"Updated: {project.ProjectFilePath}");
+                    //Console.WriteLine($"Updated: {project.ProjectFilePath}");
```
pranavkm: Remove

src/VersionTool/Program.cs
```diff
@@ -59,9 +60,22 @@ public static void Main(string[] args)
...
+                var pathOption = c.Option("-d|--directory", "Directory containing all repos", CommandOptionType.SingleValue);
+                var updatePatchConfigOption = c.Option("-u|--update-patchconfig", "Update patchconfig with all packages found and their current versions", CommandOptionType.NoValue);
+                var updatePackageOption = c.Option("-p|--update-package", "Update the given package before patch following the format <package>;<version>", CommandOptionType.MultipleValue);
+                var updateRepoOption = c.Option("-r|--update-repo", "Update the packages in the given repo before patch ", CommandOptionType.MultipleValue);
+
+                var patchConfig = c.Argument("patchConfig", "Path to the configuration file for patch update");
+
+                c.HelpOption("-h|--help");
+
+                c.OnExecute(() => OnUpdatePatch(pathOption, updatePatchConfigOption, updatePackageOption, updateRepoOption, patchConfig));
```
pranavkm: Code organization: would be nicer to move the executed methods in to separate types and simply pass in the values to the method:

```
public static class UpdatePatch
{
   public void Execute(UpdatePatchOptions options)
   {
     ...
   }
}
```

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+        private static int OnUpdatePatch(
+            CommandOption directoryOption,
+            CommandOption udpatePatchConfigOption,
+            CommandOption updatePackageOption,
+            CommandOption updateRepoOption,
+            CommandArgument patchConfigArgument)
+        {
+            var directory = directoryOption.Value();
+            var patchConfigPath = patchConfigArgument.Value;
+            var patchConfig = LoadJObject(patchConfigPath);
```
pranavkm: Let's deserialize this to a `PatchConfig` type. We know the structure, so no reason for it to loosely typed.

src/VersionTool/Program.cs
```diff
@@ -59,9 +60,22 @@ public static void Main(string[] args)
...
 
-                c.OnExecute(() => OnUpdateDependency(pathOption, matchingOption, dependencyArgument, versionArgument));
+                c.OnExecute(() => OnExecuteUpdateDependency(pathOption, matchingOption, dependencyArgument, versionArgument));
             });
 
+            var updatePatch = app.Command("update-patch", (c) =>
+            {
+                var pathOption = c.Option("-d|--directory", "Directory containing all repos", CommandOptionType.SingleValue);
+                var updatePatchConfigOption = c.Option("-u|--update-patchconfig", "Update patchconfig with all packages found and their current versions", CommandOptionType.NoValue);
+                var updatePackageOption = c.Option("-p|--update-package", "Update the given package before patch following the format <package>;<version>", CommandOptionType.MultipleValue);
```
pranavkm: Can we use `:`s instead? Semi-colons and Powershell are kinda a pain

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+
+                    if (parsedPackageUpdate.Length != 2)
+                    {
+                        throw new Exception($"Invalid package update option {packageUpdate}.");
+                    }
+
+                    var packageName = parsedPackageUpdate[0];
+                    var newPackageVersion = parsedPackageUpdate[1];
+
+                    var repo = repos.Where(r => r.Packages.Contains(packageName)).SingleOrDefault();
```
pranavkm: `FirstOrDefault`

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                    if (repo != null)
+                    {
+                        // update version
+                        OnUpdateVersion(
+                            new List<string>(new[] { Path.Combine(repo.Path, "src", packageName) }),
+                            new List<string>(),
+                            parsedPackageUpdate[1]);
+
+                        // update entry in packages
+                        packages.Where(p => string.Equals(p.Name, packageName, StringComparison.OrdinalIgnoreCase)).Single().NewVersion = newPackageVersion;
```
pranavkm:
```
var package = packages.First(p => string.Equals(p.Name, packageName, StringComparison.OrdinalIgnoreCase);
package.NewVersion = version;
```

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                            new List<string>(new[] { directory }),
+                            new List<string>(),
+                            parsedPackageUpdate[0],
+                            parsedPackageUpdate[1]);
+                    }
+                }
+            }
+
+            // Handle repo updates
+            Console.WriteLine("Update repo");
```
pranavkm: This should be inside the `if`

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+            if (updateRepoOption.HasValue())
+            {
+                foreach (var repoUpdate in updateRepoOption.Values)
```
pranavkm: We're nearly never going to be updating multiple repos for a bug. Let's this be single itemed value.

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                {
+                    var repo = repos.Where(r => string.Equals(r.Name, repoUpdate, StringComparison.OrdinalIgnoreCase)).Single();
```
pranavkm: `repos.FirstOrDefault`. Throw if the repo isn't found (mistyped)

src/VersionTool/ProcessRunner.cs
```diff
@@ -0,0 +1,136 @@
...
+            return this;
+        }
+
+        public ProcessRunner WithWorkingDirectory(string workingDirectory)
+        {
+            _workingDirectory = workingDirectory;
+            return this;
+        }
+
+        public ProcessRunner AddEnvironmentVariable(string name, string value)
```
pranavkm: Doesn't look like a bunch of these methods are used. Can we clean this type up?

pranavkm: Alternatively, consider using `Command` from `Microsoft.Dotnet.Cli.Utils`.

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+        private static bool ContainsModifiedSrcProjectFile(string path)
+        {
+            var srcPath = Path.Combine(path, "src");
+            if (!Directory.Exists(srcPath))
+            {
+                return false;
+            }
+
+            var stdout = new StringBuilder();
+            var exitCode = new ProcessRunner("git", "diff -- src/**/project.json")
```
pranavkm: What's this doing?

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                var unUpdatePackage = packages.Where(p => p.Name == package.Name && p.CurrentVersion == p.NewVersion).SingleOrDefault();
+                if (unUpdatePackage != null)
+                {
+                    Console.WriteLine($"{package.Name} is a src project and has not been updated in this patch, updating version.");
+
+                    var rule = GetRuleForPackage(package, rules);
+
+                    if (rule == null)
+                    {
+                        throw new Exception($"No update rule found for package {package.Name} {package.Version}.");
```
pranavkm: Would this run a default rule that says update transitives by .1 if it hasn't been updated in this patch?

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+            SaveJObject(patchConfigPath, patchConfig);
+
+            return 0;
+        }
+
+        private static void UpdatePackagesInRepo(Repository repo, string directory, IEnumerable<Package> packages, IEnumerable<Rule> rules)
+        {
+            foreach (var package in EnumerateProjects(new[] { Path.Combine(repo.Path, "src") }))
+            {
+                var unUpdatePackage = packages.Where(p => p.Name == package.Name && p.CurrentVersion == p.NewVersion).SingleOrDefault();
```
pranavkm: `Where.SingleOrDefault` -> `FirstOrDefault`

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                    unUpdatePackage.NewVersion = rule.NewVersion;
+                }
+                else
+                {
+                    Console.WriteLine($"{package.Name} is a src project but has been updated, skipping.");
+                }
+            }
+        }
+
+        private static void BuildRepoGraph(List<Repository> repos)
```
pranavkm: Code organization - this can be in a separate type.

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+        {
+            var includedSubDirs = new[] { "src", "test", "samples", "tools" };
+
+            foreach (var repo in repos)
+            {
+                foreach (var subdir in includedSubDirs)
```
pranavkm: `subDir`

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                {
+                    var repoSubDir = Path.Combine(repo.Path, subdir);
+                    if (Directory.Exists(repoSubDir))
+                    {
+                        if (subdir == "src")
+                        {
+                            repo.Packages.UnionWith(Directory.EnumerateDirectories(repoSubDir).Select(path => new DirectoryInfo(path).Name));
```
pranavkm: `Directory.EnumerateDirectories(repoSubDir).Select(Path.GetFileName)`

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+                        }
+
+                        foreach (var projectJsonPath in Directory.GetFiles(repoSubDir, "project.json", SearchOption.AllDirectories))
+                        {
+                            AddDependencies(projectJsonPath, repo.PackageDependencies);
+                        }
+                    }
+                }
+
+                repo.PackageDependencies.ExceptWith(repo.Packages);
```
pranavkm: Could we not add to `PackageDependencies` if it's the `src` dir to avoid this?

src/VersionTool/Program.cs
```diff
@@ -225,13 +253,324 @@ private static int OnList(CommandOption pathOption)
...
+        }
+
+        private class Package
+        {
+            public string Name;
+            public string CurrentVersion;
+            public string NewVersion;
+        }
+
+        private class Repository
```
pranavkm: Separate file.

src/VersionTool/Program.cs
JunTaoLuo: Yea this tool could use some clean up. Everything was put into the same class previously so I didn't bother organizing the new code but it's probably better that way.

src/VersionTool/Program.cs
JunTaoLuo: Oops, I was debugging something and add/removed some of the logging. Should have checked after I pushed my last commit.

src/VersionTool/Program.cs
JunTaoLuo: This is debug code, I need to remove it

src/VersionTool/Program.cs
JunTaoLuo: For multiple repo updates, just run the tool again with the new repo as an argument.

src/VersionTool/Program.cs
JunTaoLuo: I would say let's keep rules explicit. I don't want to accidentally update some versions to something unexpected. For example we had a patch where tools were versioned 1.0.0-preview2-final to 1.0.0-preview4-final. In that case, if you didn't find a specific rule it would have been updated to 1.0.1-preview2-final.

src/VersionTool/Program.cs
JunTaoLuo: This is the hacky way of checking whether there's any version updates to `src` packages. It is probably possible to remove this mechanism of detection actually and keep track of the changes made during patch update.

Originally I envision the tool to be run after you update a version number in a `project.json` of the package you are working on. Before you commit the change, you would use this tool to update all the dependent packages. Then you would commit all the changes to each repo. Now that we are passing the package/repo to update by the command line, I think this might not be needed.

src/VersionTool/ProcessRunner.cs
JunTaoLuo: In fact, this may not be needed anymore. I added this to be able to run `git diff` but as per the comments below. That mechanism of detecting package version changes may not be needed anymore since we pass them in via command line args.

src/VersionTool/Program.cs
JunTaoLuo: Actually, this will probably still be needed. For example, if you update a System package, you would probably want the tool to go through every repo to figure out whether a src package needs an update instead of manually checking and then calling the tool on every repo.

src/VersionTool/Program.cs
JunTaoLuo: What do you mean? you need to add the dependencies that are contained in all the `src` dependencies but you don't want project dependencies in the same repo. Those are not useful.

src/VersionTool/Program.cs
JunTaoLuo: Actually why `FirstOrDefault`? There shouldn't be duplicates and if there is, something has gone wrong. In that case I would prefer if the tool just barfs. On that note, I should do a bit more exception handling and argument validation.

JunTaoLuo:
:up::date:

src/VersionTool/Program.cs
```diff
@@ -73,252 +78,5 @@ public static void Main(string[] args)
 
             app.Execute(args);
         }
-
```
pranavkm: 👍 pretty!

src/VersionTool/UpdatePatchCommand.cs
```diff
@@ -0,0 +1,273 @@
...
+                var unUpdatedPackage = patchConfig.Packages.Where(p => p.Name == package.Name && p.CurrentVersion == p.NewVersion).FirstOrDefault();
+                if (unUpdatedPackage != null)
+                {
+                    Console.WriteLine($"{package.Name} is a src project and has not been updated in this patch, updating version.");
+
+                    var rule = GetRuleForPackage(package, patchConfig.Rules);
+
+                    if (rule == null)
+                    {
+                        throw new Exception($"No update rule found for package {package.Name} {package.Version}.");
```
pranavkm: Could you include an example in the error so somebody that runs it has an idea on what they are expected to do?

src/VersionTool/VersionTool.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.0.0" />
+    <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="1.0.0-*" />
```
pranavkm: Let's pin this to the RTM version (https://www.nuget.org/packages/Microsoft.DotNet.Cli.Utils/1.0.0-preview2-003121). We know it works with 1.0.0 NETCore.App

src/VersionTool/Utilities.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    public static class Utilities
+    {
+        public static IEnumerable<Project> EnumerateProjects(IEnumerable<string> paths)
+        {
+            return paths.SelectMany(p =>
+            {
+                return
+                    Directory.EnumerateFiles(p, "project.json", SearchOption.AllDirectories)
+                    .Select(f => ProjectReader.GetProject(f));
```
pranavkm: Might have to skip the `.build` dir. It might have the tools project.json.

src/VersionTool/UpdatePatchCommand.cs
```diff
@@ -0,0 +1,273 @@
...
+            foreach (var repo in patchConfig.Repos.OrderBy(r => r.Order))
+            {
+                if (ContainsModifiedSrcProjectFile(repo.Path))
+                {
+                    UpdatePackagesInRepo(repo, directory, patchConfig);
+                }
+            }
+
+            // Update patch config with updated packages
+            UpdatePatchConfig(patchConfig, patchConfigPath, repoConverter);
```
pranavkm: Can you include a help message that says this needs to be checked in?

src/VersionTool/Repository.cs
```diff
@@ -0,0 +1,28 @@
...
+    public class Repository
+    {
+        public Repository(string name, string path)
+        {
+            Name = name;
+            Path = path;
+        }
+
+        public string Name { get; }
+        public string Path { get; }
```
pranavkm: You might be able to stick `[JsonIgnore]` on these properties and avoid a custom converter.

src/VersionTool/Repository.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public HashSet<string> Packages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
pranavkm: Properties `public  HashSet<string> Packages { get; } = new HashSet<string>`

src/VersionTool/UpdatePatchCommand.cs
pranavkm: Like instructions at the on completion of the command or something.

Issue #159, opened 2017-02-03
Match assembly and package version


jbagga:
Add rule to `NuGetPackageVerifier` to match assembly version to package version

PR #160, opened 2017-02-06
Add PackageVersionMatchesAssemblyVersionRule

jbagga:
Add `PackageVersionMatchesAssemblyVersionRule` to `NuGetPackageVerifier`
Addresses #159 

cc @ajaybhargavb

### DIFF
--- a/src/NuGetPackageVerifier/PackageIssueFactory.cs
+++ b/src/NuGetPackageVerifier/PackageIssueFactory.cs
@@ -39,12 +39,12 @@ namespace NuGetPackageVerifier
                 PackageIssueLevel.Error);
         }
 
-        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(Version assemblyVersion, Version packageVersion, string packageId)
+        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(string currentFilePath, Version assemblyVersion, Version packageVersion, string packageId)
         {
             return new PackageVerifierIssue(
-                "ASSEMBLY_VERSION",
-                packageId,
-                $"The assembly version '{assemblyVersion}' does not match '{packageVersion}' for package {packageId}.",
+                "ASSEMBLY_VERSION_MISMATCH",
+                currentFilePath,
+                $"The assembly version '{assemblyVersion}' does not match the package version '{packageVersion}' for package {packageId}.",
                 PackageIssueLevel.Warning);
         }
 
--- a/src/NuGetPackageVerifier/PackageIssueFactory.cs
+++ b/src/NuGetPackageVerifier/PackageIssueFactory.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Reflection;
 using NuGet.Frameworks;
 using NuGet.Versioning;
@@ -36,6 +37,15 @@ namespace NuGetPackageVerifier
                     @"The managed assembly '{0}' in this package is missing the '[assembly: AssemblyMetadata(""Serviceable"", ""True"")]' attribute.",
                     assemblyPath),
                 PackageIssueLevel.Error);
+        }
+
+        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(Version assemblyVersion, Version packageVersion, string packageId)
+        {
+            return new PackageVerifierIssue(
+                "ASSEMBLY_VERSION",
+                packageId,
+                $"The assembly version '{assemblyVersion}' does not match '{packageVersion}' for package {packageId}.",
+                PackageIssueLevel.Warning);
         }
 
         public static PackageVerifierIssue AssemblyMissingHashAttribute(string assemblyPath)
--- a/src/NuGetPackageVerifier/PackageIssueFactory.cs
+++ b/src/NuGetPackageVerifier/PackageIssueFactory.cs
@@ -39,12 +39,12 @@ namespace NuGetPackageVerifier
                 PackageIssueLevel.Error);
         }
 
-        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(string currentFilePath, Version assemblyVersion, Version packageVersion, string packageId)
+        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(string assemblyPath, Version assemblyVersion, Version packageVersion, string packageId)
         {
             return new PackageVerifierIssue(
                 "ASSEMBLY_VERSION_MISMATCH",
-                currentFilePath,
-                $"The assembly version '{assemblyVersion}' does not match the package version '{packageVersion}' for package {packageId}.",
+                assemblyPath,
+                $"The managed assembly version '{assemblyVersion}' does not match the package version '{packageVersion}' for package {packageId}.",
                 PackageIssueLevel.Warning);
         }
 
--- a/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
+++ b/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Mono.Cecil;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class PackageVersionMatchesAssemblyVersionRule : AssemblyHasAttributeRuleBase, IPackageVerifierRule
+    {
+        private static Version _packageVersion;
+        private static string _packageId;
+
+        public override IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
+        {
+            _packageVersion = context.Metadata.Version.Version;
+            _packageId = context.Metadata.Id;
+            return base.Validate(context);
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var assemblyVersion = assemblyAttributes.SingleOrDefault(a => a.AttributeType.FullName ==
+            (typeof(AssemblyInformationalVersionAttribute).FullName ?? typeof(AssemblyVersionAttribute).FullName))
+            .AttributeType.Module.Assembly.Name.Version;
+
+            if (_packageVersion != assemblyVersion)
+            {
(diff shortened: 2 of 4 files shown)


src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -38,6 +39,15 @@ public static PackageVerifierIssue AssemblyMissingServicingAttribute(string asse
                 PackageIssueLevel.Error);
         }
 
+        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(Version assemblyVersion, Version packageVersion, string packageId)
+        {
+            return new PackageVerifierIssue(
+                "ASSEMBLY_VERSION",
+                packageId,
```
ajaybhargavb: Need to use assembly file path here as that includes the TFM of the assembly.

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -38,6 +39,15 @@ public static PackageVerifierIssue AssemblyMissingServicingAttribute(string asse
...
+                $"The assembly version '{assemblyVersion}' does not match '{packageVersion}' for package {packageId}.",
```
ajaybhargavb: Suggest including the actual assembly file path along with this message.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,37 @@
...
+            _packageId = context.Metadata.Id;
+            return base.Validate(context);
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var assemblyVersion = assemblyAttributes.SingleOrDefault(a => a.AttributeType.FullName ==
+            (typeof(AssemblyInformationalVersionAttribute).FullName ?? typeof(AssemblyVersionAttribute).FullName))
+            .AttributeType.Module.Assembly.Name.Version;
```
ajaybhargavb: This is going to null ref here if none of those attributes are found. We probably should let `AssemblyHasVersionAttributesRule` take care of that case and just return here if the attribute is not found.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,37 @@
...
+            _packageVersion = context.Metadata.Version.Version;
+            _packageId = context.Metadata.Id;
+            return base.Validate(context);
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var assemblyVersion = assemblyAttributes.SingleOrDefault(a => a.AttributeType.FullName ==
+            (typeof(AssemblyInformationalVersionAttribute).FullName ?? typeof(AssemblyVersionAttribute).FullName))
```
ajaybhargavb: Better to use `.Equals()` and specify `StringComparison.Ordinal`

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Linq;
+using System.Reflection;
+using Mono.Cecil;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class PackageVersionMatchesAssemblyVersionRule : AssemblyHasAttributeRuleBase, IPackageVerifierRule
+    {
+        private static Version _packageVersion;
+        private static string _packageId;
```
ajaybhargavb: Can drop the static here. We create only one instance of the rule anyway.

jbagga:
@ajaybhargavb Updated

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,44 @@
...
+            _packageVersion = context.Metadata.Version.Version;
+            _packageId = context.Metadata.Id;
+            return base.Validate(context);
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var assemblyVersion = assemblyAttributes.SingleOrDefault(a =>
+            a.AttributeType.FullName.Equals(
```
ajaybhargavb: Indent

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,44 @@
...
+                (typeof(AssemblyInformationalVersionAttribute).FullName
+                ?? typeof(AssemblyVersionAttribute).FullName), StringComparison.Ordinal))
```
ajaybhargavb: For PranavPoints: Put the `??` on the previous line and bring `StringComparison.Ordinal` to the next line. Also don't need the extra parentheses ().

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,44 @@
...
+                .AttributeType.Module.Assembly.Name.Version;
+
+            if (assemblyVersion == null)
+            {
+                yield return null;
```
ajaybhargavb: yield break

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var assemblyVersion = assemblyAttributes.SingleOrDefault(a =>
+            a.AttributeType.FullName.Equals(
+                (typeof(AssemblyInformationalVersionAttribute).FullName
+                ?? typeof(AssemblyVersionAttribute).FullName), StringComparison.Ordinal))
+                .AttributeType.Module.Assembly.Name.Version;
```
ajaybhargavb: Did you try running this code for the case when both attributes are missing? I still think this is going to throw a null ref exception. Could probably split this,
```C#
var versionAttribute = assemblyAttributes.SingleOrDefault(...);

if (versionAttribute == null)
{
    yield break;
}
            
var assemblyVersion = versionAttribute.AttributeType...
```

jbagga:
@ajaybhargavb Updated

src/NuGetPackageVerifier/PackageIssueFactory.cs
```diff
@@ -38,6 +39,15 @@ public static PackageVerifierIssue AssemblyMissingServicingAttribute(string asse
                 PackageIssueLevel.Error);
         }
 
+        public static PackageVerifierIssue AssemblyVersionDoesNotMatchPackageVersion(string currentFilePath, Version assemblyVersion, Version packageVersion, string packageId)
+        {
+            return new PackageVerifierIssue(
+                "ASSEMBLY_VERSION_MISMATCH",
+                currentFilePath,
```
ajaybhargavb: Could just name this `assemblyPath` to maintain consistency with other methods in this class.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,45 @@
...
+            _packageId = context.Metadata.Id;
+            return base.Validate(context);
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var versionAttribute = assemblyAttributes.SingleOrDefault(a =>
+                a.AttributeType.FullName.Equals(
+                typeof(AssemblyInformationalVersionAttribute).FullName ??
```
ajaybhargavb: Indent

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -0,0 +1,45 @@
...
+            _packageId = context.Metadata.Id;
+            return base.Validate(context);
+        }
+
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+            Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
+        {
+            var versionAttribute = assemblyAttributes.SingleOrDefault(a =>
+                a.AttributeType.FullName.Equals(
+                    typeof(AssemblyInformationalVersionAttribute).FullName ??
```
natemcmaster: AssemblyInformationalVersionAttribute and AssemblyVersionAttribute are two different things. We shouldn't swap between them arbitrarily.

natemcmaster:
This rule will break when we build RTM packages. And RTM package might look like this

Item | Value
-----|----
AssemblyInformationalVersionAttribute | 1.0.0-rtm-build###
AssemblyVersion | 1.0.0.0
 package version | 1.0.0

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
jbagga: Should it instead only look at `AssemblyInformationalVersionAttribute`? The reason I did this was because according to NuGet's [docs](https://docs.microsoft.com/en-us/nuget/schema/nuspec#Replacement_Tokens) if `AssemblyInformationalVersionAttribute` is not specified, the `$version$` value defaults to `AssemblyVersionAttribute`.

natemcmaster: Those docs refer to "old csproj" as in VS 2013. This behavior has changed slightly.

jbagga:
@natemcmaster Since only the `Version` is compared, your example will compare 1.0.0 to 1.0.0 and pass

natemcmaster:
Good point. I tested on RTM package and it passes since it appears to ignore the pre-release label. We talked offline and we can probably add additional smarts here to ensure RTM vs prerelease packages have the right assembly attribute checked against package version.

ajaybhargavb:
> We talked offline and we can probably add additional smarts here to ensure RTM vs prerelease packages have the right assembly attribute checked against package version.

While we are doing that, we should also make sure https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/Rules/AssemblyHasVersionAttributesRule.cs#L22 remains consistent with how we want to check for versions.

PR #161, opened 2017-02-07
Generated assembly attribute containing commit hash

natemcmaster:
Add a project named Microsoft.AspNetCore.BuildTools.Tasks that contains MSBuild tasks for common build step.s

Update Internal.AspNetCore.Sdk to include this tasks package, and add targets to automatically generate `System.Reflection.AssemblyMetadata("CommitHash", "$git hash$")` on build.

GetGitCommitInfo based on work @anurse did for KoreBuild 2.0.

Required adding nuspec files to workaround https://github.com/NuGet/Home/issues/4521


build/common.props
```diff
@@ -16,4 +16,12 @@
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
   </PropertyGroup>
 
-</Project>
\ No newline at end of file
+  <!-- resolve using git.exe because the GetGitCommitInfoTask can't be loaded before it is compiled -->
+  <!-- TODO set this in repo.targets so it applies to the entire build -->
+  <Target Name="ResolveCommitHashWithExe" Condition="'$(CommitHash)'==''" BeforeTargets="PrepareForBuild">
+    <Exec Command="git rev-parse HEAD" ConsoleToMSBuild="true" EchoOff="true" Condition="'$(CommitHash)'==''">
```
anurse: If we need to do this in places where we can't use a custom task, it might just be worth using git.exe everywhere. The main reason I didn't do that was because I didn't want our build to be dependent upon git being in the path.

src/Internal.AspNetCore.Sdk/build/Git.targets
```diff
@@ -0,0 +1,63 @@
...
+          DependsOnTargets="ResolveCommitHash;PrepareForBuild;GenerateIntermediateCommitHash;CoreGenerateCommitHashAttribute"
+          Condition="'$(GenerateCommitHashAttribute)'=='true' and '$(DesignTimeBuild)'!='true'" >
+    <Warning Text="Property 'CommitHash' was not set"
+             Condition="'$(CommitHash)'==''" />
+  </Target>
+
+  <Target Name="GenerateIntermediateCommitHash" Condition="'$(CommitHash)' != ''">
+    <PropertyGroup>
+      <!-- shorten to help avoid max path length issues -->
+      <IntermediateCommitHash>$(IntermediateOutputPath)$(CommitHash.Substring(0, 10)).commit</IntermediateCommitHash>
```
pranavkm: Couldn't we write a .commit file (conditionally) with the content rather than make one for each new commit?

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// The full commit SHA of the current commit referenced by HEAD.
+        /// </summary>
+        [Output]
+        public string CommitHash { get; set; }
+
+        /// <summary>
+        /// The folder containing the '.git', not the .git folder itself.
+        /// </summary>
+        [Output]
+        public string GitProjectRootPath { get; set; }
```
pranavkm: `RepositoryRootPath` \ `SolutionRoot`?

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public override bool Execute()
+        {
+            Branch = null;
+            CommitHash = null;
+
+            var gitRoot = new DirectoryInfo(WorkingDirectory);
+            while (gitRoot != null)
+            {
+                if (gitRoot.GetDirectories(".git").Any())
```
pranavkm:
```C#
var gitDirectory = Path.Combine(gitRoot, ".git");
if (Directory.Exists(".git"))
{
  break;
}
```

avoids the unnecessary enumeration.

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        [Output]
+        public string GitProjectRootPath { get; set; }
+
+        public override bool Execute()
+        {
+            Branch = null;
+            CommitHash = null;
+
+            var gitRoot = new DirectoryInfo(WorkingDirectory);
```
pranavkm: Code organization: Split the locating the git root in to a separate method.

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// The folder containing the '.git', not the .git folder itself.
+        /// </summary>
+        [Output]
+        public string GitProjectRootPath { get; set; }
+
+        public override bool Execute()
+        {
+            Branch = null;
```
pranavkm: Is resetting this necessary?

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+            if (!File.Exists(headFile))
+            {
+                Log.LogError("Unable to determine active git branch.");
+                return false;
+            }
+
+            var content = File.ReadAllText(headFile).Trim();
+            if (!content.StartsWith("ref: refs/heads/"))
+            {
+                Log.LogError("'.git/HEAD' file in unexpected format, unable to determine active git branch");
```
pranavkm: Let's include the `content` as part of the error. Would be easy to diagnose if this ever fails.

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+                return false;
+            }
+
+            Branch = content.Substring(16);
```
pranavkm:
```
const string HeadContentStart = "ref: refs/heads/";
...
Branch = content.Substring(HeadContentStart.Length);
```

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (string.IsNullOrEmpty(Branch))
+            {
+                Log.LogMessage("Current branch appears to be empty. Failed to retrieve current branch.");
+            }
+
+            var branchFile = Path.Combine(GitProjectRootPath, ".git", "refs", "heads", Branch.Replace('/', Path.DirectorySeparatorChar));
```
pranavkm: Might not need the `Replace`.

build/common.props
natemcmaster: Yeah, I considered that too. But decided to keep using the task for the reason you mentioned.
Plus, we don't want to require the source code to be in git either. This implementation will warn but not fail if git is unavailable.

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
pranavkm: `Internal.AspnetCore.BuildTool.Tasks`

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.nuspec
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
pranavkm: Can't we use the proj file to generate this?

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <Description>MSBuild tasks. This package is intended for Microsoft use only</Description>
+    <EnableDefaultItems Condition="'$(TargetFramework)' == 'netstandard1.0'">false</EnableDefaultItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NETStandard.Library" Version="1.6.0" PrivateAssets="All" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
+    <PackageReference Include="Microsoft.Build.Framework" Version="15.1.0-preview-000523-01" PrivateAssets="All" />
```
pranavkm: Do these need to be pinned?

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
anurse: Probably not. It was probably in my original code as a paranoia step.

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.csproj
anurse: `Internal.AspNetCore.BuildTools.Tasks`

src/Internal.AspNetCore.Sdk/build/Git.targets
natemcmaster: The intermediate step is required to preserve incremental build. Inputs/Outputs checks file timestamps. I tried to implement this in one hop, but all variations I tried broke incremental build.

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
natemcmaster: MSBuild shouldn't instantiate these properties since they're marked as 'output'. So yeah, can be removed.

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.csproj
natemcmaster: No, they can float. Although, it doesn't really matter. MSBuild will lift the task to whatever version it carries.

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.nuspec
natemcmaster: Yes, but it's gnarly. It's easier to just use nuspec than to tricking 'dotnet pack' into giving us the layout we want. I tried both.

src/Microsoft.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
natemcmaster: +1 RepositoryRootPath.

src/Internal.AspNetCore.Sdk/build/Git.targets
pranavkm: Well, something along the lines of 
```xml
<ReadLinesFromFile File="$(IntermediateOutputPath)\commit" Output="OldCommit" />
<WriteLinesToFile File="$(IntermediateOutputPath)\commit" Condition="$(OldCommit) != $(Commit)" />
```

avoids having more than one file on disk we have to scan

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.csproj
pranavkm: Yup. The intent being we really don't want folks to assume these are general purpose build tools that Microsoft is shipping.

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.nuspec
pranavkm: If you feel so. There's a bunch of metadata duplicated here.

src/Microsoft.AspNetCore.BuildTools.Tasks/Microsoft.AspNetCore.BuildTools.Tasks.nuspec
```diff
@@ -0,0 +1,21 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+  <metadata>
+    <id>Microsoft.AspNetCore.BuildTools.Tasks</id>
+    <version>$version$</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
```
pranavkm: Shouldn't need this

src/Internal.AspNetCore.Sdk/build/Git.targets
```diff
@@ -0,0 +1,63 @@
...
+
+    <ItemGroup>
+      <_OldCommitFiles Include="$(IntermediateOutputPath)*.commit" Exclude="$(IntermediateCommitHash)" />
+    </ItemGroup>
+
+    <Delete Files="@(_OldCommitFiles)" />
+    <Touch Files="$(IntermediateCommitHash)" AlwaysCreate="true" Condition="!Exists('$(IntermediateCommitHash)')" />
+  </Target>
+
+  <Target Name="CoreGenerateCommitHashAttribute"
```
pranavkm: Super unrelated - is this the tabbing strategy we plan on using? I'd prefer we just tab because it means you don't have to go redo these if the value on the first line changes (doesn't necessarily apply here).

src/Internal.AspNetCore.Sdk/build/Git.targets
natemcmaster: I don't see how this solves the problem. You would still need two files to make incremental build work.

src/Internal.AspNetCore.Sdk/build/Git.targets
natemcmaster: Can you clarify? Do you mean you want
`<Target ...(all attributes)>` on one line?

natemcmaster:
Ping @pranavkm are you okay with this?

natemcmaster:
Awesome, thanks for clarifying. Will update.

Btw, deleting old obj/*.commit files is required to make sure anytime you switch branches or move between commits you've built previously triggers a rebuild.

anurse:
> Btw, deleting old obj/*.commit files is required to make sure anytime you switch branches or move between commits you've built previously triggers a rebuild.

The worst case scenario here though is that if I switch branches without doing this and a rebuild isn't triggered (because all the other files are the same), the commit hash will be incorrect, right? That doesn't seem too scary to me.

natemcmaster:
Maybe harmless for local builds, but what about CI agents that are not
guaranteed to "git clean -xfd" between builds?

pranavkm:
> but what about CI agents that are not guaranteed to "git clean -xfd" between builds?

We really don't build without doing a super clean build on the CI. Never a good idea to use left over bits when doing a build

PR #164, opened 2017-02-09
Change comparison for RTM vs pre-rel packages

jbagga:
Addresses #159 
In response to PR feedback on #160 

cc @natemcmaster


natemcmaster:
Can you rebase on dev? It's hard to tell has changed.

PR #165, opened 2017-02-09
Compare all version attributes to package version

jbagga:
Addresses #159 

In response to feedback on #160

cc @natemcmaster @Eilon

### DIFF
--- a/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
+++ b/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
@@ -22,7 +22,9 @@ namespace NuGetPackageVerifier.Rules
             return base.Validate(context);
         }
 
-        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
+        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(
+            string currentFilePath,
+            AssemblyDefinition assembly,
             Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
         {
             var assemblyInformationalVersionAttribute = assemblyAttributes.SingleOrDefault(a =>
@@ -55,8 +57,8 @@ namespace NuGetPackageVerifier.Rules
                     _packageId);
             }
 
-            var assemblyVersion = assemblyInformationalVersionAttribute.AttributeType.Module.Assembly.Name.Version.ToString();
-            if (!_packageVersion.Version.ToString().Equals(assemblyVersion))
+            var assemblyVersion = assembly.Name.Version;
+            if (!_packageVersion.Version.Equals(assemblyVersion))
             {
                 yield return PackageIssueFactory.AssemblyVersionDoesNotMatchPackageVersion(
                     currentFilePath,
@@ -70,17 +72,18 @@ namespace NuGetPackageVerifier.Rules
         {
             if (packageVersion == null)
             {
-                throw new ArgumentNullException("packageVersion");
+                throw new ArgumentNullException(nameof(packageVersion));
             }
 
             if (assemblyNuGetVersion == null)
             {
-                throw new ArgumentNullException("assemblyNuGetVersion");
+                throw new ArgumentNullException(nameof(assemblyNuGetVersion));
             }
 
-            if (packageVersion.Major.Equals(assemblyNuGetVersion.Major) &&
-                packageVersion.Minor.Equals(assemblyNuGetVersion.Minor) &&
-                packageVersion.Patch.Equals(assemblyNuGetVersion.Patch))
+            // Pre-release and build metadata does not need to match
+            if (packageVersion.Major == assemblyNuGetVersion.Major &&
+                packageVersion.Minor == assemblyNuGetVersion.Minor &&
+                packageVersion.Patch == assemblyNuGetVersion.Patch)
             {
                 return true;
             }
--- a/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
+++ b/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
@@ -6,17 +6,20 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using Mono.Cecil;
+using NuGet.Versioning;
 
 namespace NuGetPackageVerifier.Rules
 {
     public class PackageVersionMatchesAssemblyVersionRule : AssemblyHasAttributeRuleBase
     {
-        private Version _packageVersion;
+        private bool _isPrerelease;
+        private NuGetVersion _packageVersion;
         private string _packageId;
 
         public override IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
         {
-            _packageVersion = context.Metadata.Version.Version;
+            _isPrerelease = context.Metadata.Version.IsPrerelease;
+            _packageVersion = context.Metadata.Version;
             _packageId = context.Metadata.Id;
             return base.Validate(context);
         }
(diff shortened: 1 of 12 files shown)


src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,39 +6,54 @@
...
         {
+
             var versionAttribute = assemblyAttributes.SingleOrDefault(a =>
                 a.AttributeType.FullName.Equals(
-                    typeof(AssemblyInformationalVersionAttribute).FullName ??
-                    typeof(AssemblyVersionAttribute).FullName,
+                    typeof(AssemblyInformationalVersionAttribute).FullName,
                     StringComparison.Ordinal));
 
             if (versionAttribute == null)
```
natemcmaster: Issue an error instead? We should always have AssemblyInformationalVersionAttribute

jbagga: There is already a [rule](https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/Rules/AssemblyHasVersionAttributesRule.cs#L22) for that

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,39 +6,54 @@
...
     {
-        private Version _packageVersion;
+        private bool _isPrerelease;
+        private NuGetVersion _packageVersion;
         private string _packageId;
 
         public override IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
         {
-            _packageVersion = context.Metadata.Version.Version;
+            _isPrerelease = context.Metadata.Version.IsPrerelease;
```
Eilon: Hmm first time I'm noticing this, but I don't think this code is technically safe (though in practice it will happen to work). The problem is that the same package rule instances are used for all packages, so package rules need to avoid having per-package state. We should fix this in a separate PR (after this PR). I filed https://github.com/aspnet/BuildTools/issues/166 to track fixing this.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,39 +6,54 @@
...
+            _packageVersion = context.Metadata.Version;
             _packageId = context.Metadata.Id;
             return base.Validate(context);
         }
 
         public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath,
             Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes)
         {
+
             var versionAttribute = assemblyAttributes.SingleOrDefault(a =>
```
Eilon: Rename this variable to `assemblyInformationalVersionAttribute`.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,39 +6,54 @@
...
                 a.AttributeType.FullName.Equals(
-                    typeof(AssemblyInformationalVersionAttribute).FullName ??
-                    typeof(AssemblyVersionAttribute).FullName,
+                    typeof(AssemblyInformationalVersionAttribute).FullName,
                     StringComparison.Ordinal));
 
             if (versionAttribute == null)
             {
                 yield break;
```
Eilon: This shouldn't exit early. It should always try to check both informational version and real version (whenever each is available).

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,39 +6,54 @@
...
             }
 
-            var assemblyVersion = versionAttribute.AttributeType.Module.Assembly.Name.Version;
-            if (!_packageVersion.Equals(assemblyVersion))
+            if (_isPrerelease)
```
Eilon: See comment above, we always need to check both info version and real version. The check does need to vary based on IsPreRelease (because real version is never pre-release), but both attributes always need to be checked.

jbagga:
@Eilon Updated

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,40 +6,86 @@
...
+                    _packageVersion.Version,
+                    _packageId);
+            }
+        }
+
+        private bool VersionEquals(NuGetVersion packageVersion, NuGetVersion assemblyNuGetVersion)
+        {
+            if (packageVersion == null)
+            {
+                throw new ArgumentNullException("packageVersion");
```
Eilon: `nameof(x)`?

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,40 +6,86 @@
...
+            }
+
+            if (assemblyNuGetVersion == null)
             {
-                yield return PackageIssueFactory.AssemblyVersionDoesNotMatchPackageVersion(currentFilePath, assemblyVersion, _packageVersion, _packageId);
+                throw new ArgumentNullException("assemblyNuGetVersion");
             }
+
+            if (packageVersion.Major.Equals(assemblyNuGetVersion.Major) &&
```
Eilon: Can these be plain old `==` checks? Or is there some special type for these version segments?

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,40 +6,86 @@
...
+                packageVersion.Minor.Equals(assemblyNuGetVersion.Minor) &&
+                packageVersion.Patch.Equals(assemblyNuGetVersion.Patch))
+            {
```
Eilon: Might also want to add a comment somewhere here that we specifically ignore the "prerelease" part of the version because we're ok with that not matching.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
Eilon: (and other places in this file too)

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,40 +6,86 @@
...
+            if (!VersionEquals(_packageVersion, assemblyFileNuGetVersion))
+            {
+                yield return PackageIssueFactory.AssemblyFileVersionDoesNotMatchPackageVersion(
+                    currentFilePath,
+                    assemblyFileNuGetVersion,
+                    _packageVersion,
+                    _packageId);
+            }
+
+            var assemblyVersion = assemblyInformationalVersionAttribute.AttributeType.Module.Assembly.Name.Version.ToString();
```
Eilon: Hmm this seems like an odd way to get the assembly version. That should presumably be directly available from the assembly itself, no? Though, this code still looks a bit off to me -- is this getting the assembly version of the assembly that defines AssemblyInformationalVersionAttribute, which is mscorlib? Or is it getting the assembly version of the assembly we're testing?

Either way, I'm suspecting that the rule base class that this rule derives from is just lacking a few things that would make writing this code easier...

jbagga: `Mono.Cecil` does not expose `AssemblyVersion` attribute explicitly. It is embedded in the assembly information of each attribute (it is the assembly we are testing).

jbagga: It will clean up a lot if the rule base class made `AssemblyDefinition` accessible though

jbagga:
@Eilon Updated

PR #167, opened 2017-02-10
Replace makefile.shade

natemcmaster:
Port sake targets to MSBuild.

Also, updated the README.md and added an issue template to make it clear that these tools are meant for our use, not general consumption.


natemcmaster:
cc @jbagga - BuildTools doesn't currently build with the latest version of korebuild. Can you take a look at these changes so we can get unblocked on building this repo?

natemcmaster:
> The publish directory is missing from artifacts. Is it not required anymore?

That was an implementation detail of how we built these packages in Sake. We don't need it.

src/DependenciesPackager/DependenciesPackager.csproj
```diff
@@ -18,4 +18,17 @@
...
   </ItemGroup>
 
+  <!-- packaging settings -->
+  <PropertyGroup>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <NuspecFile>$(MSBuildThisFileDirectory)$(MSBuildProjectName).nuspec</NuspecFile>
+  </PropertyGroup>
+
+  <Target Name="SetPackageDependencies" BeforeTargets="GenerateNuspec" DependsOnTargets="Publish">
+    <PropertyGroup>
```
javiercn: You should have a property named SetPackageDependenciesDependsOn and do this through <Target Name="SetPackageDependencies" DependsOnTargets="$(SetPackageDependenciesDependsOn)">

Mixing BeforeTargets and DependsOnTargets is kind of ugly and using properties with DependsOnTargets allows you to:
1) Allow third party scripts to plug in transparently into your build script by just defining <Property>$(Property);newvalue</Property>
2) Allows you to see all the dependency graph of tasks in one place. (the property group where all these things are defined.
3) You don't have to be hunting around to modify the dependency graph.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Microsoft.AspNetCore.BuildTools.ApiCheck.csproj
```diff
@@ -21,4 +21,29 @@
...
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <NuspecFile>$(MSBuildThisFileDirectory)$(MSBuildProjectName).nuspec</NuspecFile>
+    <IntermediatePackDir>$(MSBuildThisFileDirectory)bin\$(Configuration)\publish\</IntermediatePackDir>
+    <PublishDir>$(IntermediatePackDir)$(TargetFramework)\</PublishDir>
+    <NuspecBasePath>$(IntermediatePackDir)</NuspecBasePath>
+  </PropertyGroup>
+
+  <Target Name="PublishAll">
+    <ItemGroup>
+      <_TargetFrameworks Remove="@(_TargetFrameworks)"/>
```
javiercn: What is this exactly doing?

samples/NuGetPackageVerifier.Sample/NuGetPackageVerifier.Sample.proj
```diff
@@ -1,7 +1,8 @@
-﻿<Project ToolsVersion="15.0">
+﻿<Project>
```
natemcmaster: Not required for this sample.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Microsoft.AspNetCore.BuildTools.ApiCheck.csproj
natemcmaster: Clear the itemgroup and does a cross-targeting build. The "Publish" task requires you set a TargetFramework.

src/DependenciesPackager/DependenciesPackager.csproj
natemcmaster: That may be true when developing MSBuild targets in an SDK, but seems like overkill for this one target in this one project.

Also, FWIW the SDK team is slowly trying to move off *DependsOn properties. These are the legacy of the first MSBuild which didn't have AfterTargets/BeforeTargets.

PR #169, opened 2017-02-13
Introduce library to support ApiCheck and Optimization Cache on MSBuild

javiercn:


BuildTools.sln
```diff
@@ -37,64 +37,186 @@ Project("{13B669BE-BB05-4DDF-9536-439F39A36129}") = "Internal.AspNetCore.Sdk", "
 EndProject
 Project("{13B669BE-BB05-4DDF-9536-439F39A36129}") = "Internal.AspNetCore.BuildTools.Tasks", "src\Internal.AspNetCore.BuildTools.Tasks\Internal.AspNetCore.BuildTools.Tasks.csproj", "{6A631446-BBDD-4743-B576-7F9793B6BE45}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NugetReferenceResolver", "src\NugetReferenceResolver\NugetReferenceResolver.csproj", "{DBFD3DC9-53CC-4DD5-9763-2F6D0EB708B5}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
natemcmaster: Don't add all these x86/x64 configurations. We only need AnyCPU.

src/NugetReferenceResolver/RuntimeGraph.cs
```diff
@@ -0,0 +1,926 @@
...
+        public static string GetCurrentRuntimeId()
+        {
+            if (RuntimeEnvironment.OperatingSystemPlatform != Platform.Windows)
+            {
+                return RuntimeEnvironment.GetRuntimeIdentifier();
+            }
+            else
+            {
+                var arch = RuntimeEnvironment.RuntimeArchitecture.ToLowerInvariant();
+                if (RuntimeEnvironment.OperatingSystemVersion.StartsWith("6.1", StringComparison.Ordinal))
```
natemcmaster: Why have you re-implemented what is already returned by `RuntimeEnvironment.GetRuntimeIdentifier()`?

src/NugetReferenceResolver/RuntimeGraph.cs
```diff
@@ -0,0 +1,926 @@
...
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NugetReferenceResolver
+{
+    public class RuntimeGraph
+    {
+        private const string RuntimeCompatibility = @"
```
natemcmaster: This whole graph is already available on `Microsoft.Extensions.DependencyModel.DependencyContext.RuntimeGraph.`

natemcmaster:
Maybe I don't fully understand what you are trying to do here, but it seems like there is already API in the SDK that we can re-use to accomplish this purpose. Have you looked into using the `ResolvePackageDependencies` task?

BuildTools.sln
javiercn: It's VS doing it, not me :(

src/NugetReferenceResolver/RuntimeGraph.cs
javiercn: Yep, but that is only available on .net core and this thing is cross-compiled, so just put it in here and skipped the dependency. Its basically reference data.

src/NugetReferenceResolver/RuntimeGraph.cs
javiercn: This is the same logic that the old cli had for this.

javiercn:
Not quite the same logic. ResolvePackageDependencies will give you the list of dependencies but we actually need to be able to manipulate that graph to:
1) Compute a subset of the dependencies and their transitive closure.
2) Compute a subset of the dependencies without a given package and their dependencies.
3) Be able to include this on any .exe to have process level isolation for things like API-Check that load dlls

BuildTools.sln
natemcmaster: Please revert anyways.

src/NugetReferenceResolver/Package.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet.LibraryModel;
+using NuGet.Versioning;
```
natemcmaster: Nit: sort usings. This applies to several other files, too.

src/NugetReferenceResolver/PackageAssembly.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
natemcmaster: Nit: copyright header. This applies to several other files, too.

src/NugetReferenceResolver/PackageGraph.cs
```diff
@@ -0,0 +1,302 @@
...
+            IEnumerable<Package> packages,
+            string targetFrameworkName)
+        {
+            PackageSources = packageFolders;
+            AllPackages = allPackages;
+            Dependencies = packages;
+            Framework = targetFrameworkName;
+        }
+
+        public PackageGraph Closure(string packageId)
```
natemcmaster: Naming `ComputeClosure` or `GetClosure`

src/NugetReferenceResolver/PackageGraph.cs
```diff
@@ -0,0 +1,302 @@
...
+                        if (!allPackages.ContainsKey(dependency.Name))
+                        {
+                            remainingPackages.Push(dependency);
+                        }
+                    }
+                }
+            }
+        }
+
+        public PackageGraph Without(string packageId)
```
natemcmaster: Naming: `WithoutPackage` or `RemovePackage`

src/NugetReferenceResolver/PackageGraph.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        public IEnumerable<Package> Dependencies { get; set; }
+
+        public IDictionary<string, Package> AllPackages { get; set; }
+
+        public IEnumerable<string> PackageSources { get; set; }
+
+        public string Framework { get; set; }
+
+        public static PackageGraph Create(LockFile lck, string targetFrameworkName)
```
natemcmaster: Naming `lck` => `lockFile`. `lck` looks like lick, lock, luck, and lack

src/NugetReferenceResolver/RuntimeGraph.cs
natemcmaster: Ok. You might consider this an embedded resource so it will be easier to update in the future. At the very least, move this into a separate file `partial class` so it's easier to see that there is also code in this file, not just data.

javiercn:
Updated @natemcmaster

PR #170, opened 2017-02-13
Integrate Optimization Cache with MSBuild

javiercn:


src/DependenciesPackager/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+﻿{
+  "profiles": {
+    "DependenciesPackager": {
+      "commandName": "Project",
+      "commandLineArgs":
+      "--project C:\\work\\Tests\\SplitPackages\\cache.csproj --sources C:\\work\\Tests\\SplitPackages\\Packages --sources C:\\work\\Tests\\SplitPackages\\ext --sources https://api.nuget.org/v3/index.json --destination C:\\work\\Tests\\SplitPackages\\Output --version 1.1.0 --runtime win7-x64 --prefix Antares --preserve-casing"
```
natemcmaster: Don't commit this file. It has your specific file paths in it.

src/DependenciesPackager/PackageRestoreContext.cs
```diff
@@ -44,24 +41,21 @@ public bool Restore()
...
                 _manifest,
                 packages,
-                sources,
-                fallbackFeeds);
+                sources);
 
             if (!Quiet)
             {
-                arguments += " --verbosity Verbose";
+                arguments += " --verbosity Diag";
```
natemcmaster: Suggest 'Normal' instead. Diagnostics is more than we need.

natemcmaster:
It seems like this code repeats what has already been implemented by the SDK. Have you looked into using  `ResolvePackageDependencies` and `ResolvePublishAssemblies` tasks to do the TFM/RID math and dependency trimming? cref https://github.com/dotnet/sdk/tree/master/src/Tasks/Microsoft.NET.Build.Tasks

PR #171, opened 2017-02-13
Integrate API Check with MSBuild

javiercn:


src/Microsoft.AspNetCore.BuildTools.ApiCheck/Loader/AssemblyLoader.cs
```diff
@@ -1,66 +1,33 @@
...
         {
-            var ctx = CreateProjectContext(projectJson, packagesFolder);
-            var exporter = ctx.CreateExporter(configuration);
-            var dependencies = exporter.GetDependencies().ToArray();
             
-#if NETCOREAPP1_0
-            var loader = new CoreClrAssemblyLoader(dependencies, ctx.RuntimeIdentifier, ctx.GetOutputPaths(configuration)?.CompilationOutputPath);
+#if NETCOREAPP1_1
+            var lockFile = new LockFileFormat().Read(assetsJson);
+            var graph = PackageGraph.Create(lockFile, "netcoreapp1.1");
```
natemcmaster: Can you make the target TFM a parameter instead of hard-coding it? This may vary from project to project.

javiercn: Sure

javiercn: Sure

PR #172, opened 2017-02-13
Add more MSBuild tasks

natemcmaster:
Add GetOSPlatform, SetEnvironmentVariable, and WaitForDebugger MSBuild tasks


src/Internal.AspNetCore.BuildTools.Tasks/SetEnvironmentVariable.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public override bool Execute()
+        {
+            if (string.IsNullOrEmpty(Variable))
+            {
+                Log.LogError($"{nameof(Variable)} cannot be null or an empty string");
+                return false;
+            }
+
+            Log.LogMessage("Setting env '{0}' = '{1}'", Variable, Value);
```
anurse: Let's expand this a little so it looks like a normal build message: "Setting Environment Variable '{0}' to '{1}'" (and while you're at it, using an interpolated string)

(And yes, I'm aware this is code you copied from me 😝)

PR #175, opened 2017-02-14
Remove workaround for the runtimeconfig.json bug

natemcmaster:
Companion PR to https://github.com/aspnet/KoreBuild/pull/169


src/Internal.AspNetCore.Sdk/build/Workarounds.targets
```diff
@@ -1,35 +0,0 @@
-<!--
```
pranavkm: 👏

PR #177, opened 2017-02-15
Support getting git SHA when not attached to a branch.

### LINKED ISSUE #176: CommitHash generation doesn't work during detached HEAD
GetGitCommitInfo task should support detached HEAD.

cc @moozzyk @anurse

natemcmaster:
Resolves #176

### DIFF
--- a/src/Internal.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
+++ b/src/Internal.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
@@ -10,6 +10,9 @@ namespace Microsoft.AspNetCore.BuildTools
 {
     public class GetGitCommitInfo : Task
     {
+        private const string _headContentStart = "ref: refs/heads/";
+        private const int _shaLength = 40;
+
         /// <summary>
         /// A folder inside the git project. Does not need to be the top folder.
         /// This task will search upwards for the .git folder.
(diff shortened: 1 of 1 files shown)


src/Internal.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -78,6 +89,13 @@ public override bool Execute()
             return true;
         }
 
+        private bool ResolveFromDetachedHead(string head)
+        {
+            CommitHash = head;
+            Log.LogWarning("The repo in '{0}' appears to be in detached HEAD mode. Unable to determine current git branch.", RepositoryRootPath);
```
moozzyk: Will the warning continue to appear?

src/Internal.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```diff
@@ -10,6 +10,9 @@ namespace Microsoft.AspNetCore.BuildTools
 {
     public class GetGitCommitInfo : Task
     {
+        private const string _headContentStart = "ref: refs/heads/";
+        private const int _shaLength = 40;
```
moozzyk: nit: should be `ShareLength` because it's a const.

src/Internal.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
natemcmaster: Yes, we still want this warning when we are unable to resolve the name of the branch being used. But there should be fewer warnings, and they won't cause build to fail. We can make improvements to KoreBuild so that the warning appears only once per solution instead of once per project.

PR #179, opened 2017-02-22
Add XmlPoke2 task as a substitute for XmlPoke on .NET Core

natemcmaster:
This task makes it easy to replace single values in an XML file using XPath.

Workaround https://github.com/Microsoft/msbuild/issues/1731

### DIFF
--- a/src/Internal.AspNetCore.BuildTools.Tasks/build/Tasks.tasks
+++ b/src/Internal.AspNetCore.BuildTools.Tasks/build/Tasks.tasks
@@ -1,5 +1,5 @@
 ﻿<Project>
-  
+
   <PropertyGroup>
     <_BuildToolsAssemblyTfm Condition="'$(MSBuildRuntimeType)' == 'Core'">netstandard1.3</_BuildToolsAssemblyTfm>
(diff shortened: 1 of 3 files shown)


src/Internal.AspNetCore.BuildTools.Tasks/build/Tasks.tasks
```diff
@@ -11,4 +11,7 @@
   <UsingTask TaskName="Microsoft.AspNetCore.BuildTools.SetEnvironmentVariable" AssemblyFile="$(_BuildToolsAssembly)" />
   <UsingTask TaskName="Microsoft.AspNetCore.BuildTools.WaitForDebugger" AssemblyFile="$(_BuildToolsAssembly)" />
   
+  <!-- TODO remove when XmlPoke becomes available in MSBuild 15.2 -->
```
pranavkm: Can we file a work item for this (if it isn't already there) and remove the TODO?

natemcmaster: Yup, here it is. https://github.com/aspnet/BuildTools/issues/180

PR #181, opened 2017-02-24
Add a GetDotNetHost task to locate the "dotnet.exe" path

natemcmaster:
This task helps locate the dotnet.exe file using MSBuild's runtime.

Required lifting the .NET core task version to netstandard1.6 so that we could use AppContext.GetData.

Compensates for https://github.com/dotnet/sdk/issues/20

### DIFF
--- a/src/Internal.AspNetCore.BuildTools.Tasks/GetDotNetHost.cs
+++ b/src/Internal.AspNetCore.BuildTools.Tasks/GetDotNetHost.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NETSTANDARD1_6
+
+using System;
+using System.Runtime.InteropServices;
+using System.IO;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.AspNetCore.BuildTools
+{
+    /// <summary>
+    /// <para>
+    /// Uses the current runtime information to find the "dotnet.exe" path.
+    /// This requires that MSBuild itself execute on top of Microsoft.NETCore.App.
+    /// </para>
+    /// <para>
+    /// This code is modeled of this API: https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/Muxer.cs.
+    /// We don't use this API directly as using Microsoft.DotNet.Cli.Utils with MSBuild causes assembly loading issues.
+    /// </para>
+    /// <para>
+    /// Also, this task may be not be necessary in future releases. At the time of writing, however, this
+    /// feature does not exist in MSBuild of the .NET Core SDK. See https://github.com/Microsoft/msbuild/issues/1669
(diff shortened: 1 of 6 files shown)


pranavkm:
Out of curiosity, what's this for?

src/Internal.AspNetCore.BuildTools.Tasks/GetDotNetHost.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.AspNetCore.BuildTools
+{
+    /// <summary>
+    /// Uses the current runtime information to find the "dotnet.exe" path.
+    /// This requires that MSBuild itself execute on top of Microsoft.NETCore.App.
+    /// </summary>
+    public class GetDotNetHost : Task
```
pranavkm: Can we use the code from https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/Muxer.cs as is (and add a back link)? We could get away with not updating the TFM. 
I'm guessing we wouldn't be able to use the package directly without some packaging shenanigans.

natemcmaster:
This compensates for https://github.com/Microsoft/msbuild/issues/1669.

I'm working on crossgen stuff.

src/Internal.AspNetCore.BuildTools.Tasks/GetDotNetHost.cs
natemcmaster: I considered that, but assembly loading in MSBuild for .NET core is a headache. In theory you should be able to use this, but because the package itself depends on MSBuild assemblies, it can cause issues.

src/Internal.AspNetCore.BuildTools.Tasks/GetDotNetHost.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private const string MuxerName = "dotnet";
+
+        [Output]
+        public string MuxerPath { get; set; }
```
javiercn: Don't know what a Muxer is. I would find a way simpler name like ExecutablePath or DotNetExecutable

natemcmaster: Will change to ExecutablePath. That makes more sense

src/Internal.AspNetCore.BuildTools.Tasks/GetDotNetHost.cs
pranavkm: I was going for the copying their code as-is in here? It sort of places the burden on them to keep it working correctly.

natemcmaster: It works. This is a copy of https://github.com/aspnet/DotNetTools/blob/dev/shared/DotNetMuxer.cs.

pranavkm: Fair enough. Can you add a comment to the file that says where it came from?

natemcmaster: Added.

PR #183, opened 2017-02-28
Add task to add or update nuget package source

natemcmaster:
It turns out XmlPoke can only update a value, not add a new XML node.

Prerequisite to https://github.com/aspnet/KoreBuild/pull/180

### DIFF
--- a/src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
+++ b/src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Microsoft.Build.Utilities;
+using Microsoft.Build.Framework;
+
+namespace Microsoft.AspNetCore.BuildTools
+{
+    /// <summary>
+    /// Update or adds a NuGet feed to a NuGet.config file. It reads <see cref="NuGetConfigPath"/>
+    /// and replaces or adds the feed named <see cref="SourceName"/> with <see cref="SourceUri'"/>.
+    /// </summary>
+    public class UpdatePackageSource : Task
+    {
+        [Required]
+        public string NuGetConfigPath { get; set; }
+
+        [Required]
+        public string SourceName { get; set; }
+
+        [Required]
+        public string SourceUri { get; set; }
+
+        public override bool Execute()
+        {
+            if (string.IsNullOrEmpty(SourceName))
+            {
+                Log.LogError("FeedName must not be empty");
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(SourceUri))
+            {
+                Log.LogError("PackageSource must not be empty");
+                return false;
+            }
+
+            var nugetConfig = XDocument.Load(NuGetConfigPath);
+            var packageSources = nugetConfig.Element("configuration")?.Element("packageSources");
+            var addElements = packageSources.Elements("add").ToList();
+
+            var valueToUpdate = addElements.FirstOrDefault(f => string.Equals(f.Attribute("key")?.Value, SourceName, StringComparison.OrdinalIgnoreCase));
+            if (valueToUpdate == null)
+            {
+                Log.LogMessage("Adding feed '{0}' to '{1}'", SourceName, SourceUri);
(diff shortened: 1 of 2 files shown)


src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+            else
+            {
+                Log.LogMessage("Updating feed '{0}' to '{1}'", SourceName, SourceUri);
+                valueToUpdate.Attribute("value").Value = SourceUri;
+            }
+
+            using (var file = new FileStream(NuGetConfigPath, FileMode.Create))
+            {
+                nugetConfig.Save(file);
```
moozzyk: Can't you use the overload that takes path?

src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                Log.LogMessage("Adding feed '{0}' to '{1}'", SourceName, SourceUri);
+                packageSources.Add(new XElement("add",
+                    new XAttribute("key", SourceName),
+                    new XAttribute("value", SourceUri)));
+            }
+            else
+            {
+                Log.LogMessage("Updating feed '{0}' to '{1}'", SourceName, SourceUri);
+                valueToUpdate.Attribute("value").Value = SourceUri;
```
moozzyk: This will NRE if the element exists and the but the attribute does not. A better way would be: `valueToUpdate.SetAttributeValue("value", SourceUri)`

src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                Log.LogError("PackageSource must not be empty");
+                return false;
+            }
+
+            var nugetConfig = XDocument.Load(NuGetConfigPath);
+            var packageSources = nugetConfig.Element("configuration").Element("packageSources");
+            var addElements = packageSources.Elements("add").ToList();
+
+            var valueToUpdate = addElements.FirstOrDefault(f => string.Equals(f.Attribute("key").Value, SourceName, StringComparison.OrdinalIgnoreCase));
```
moozzyk: Use explicit conversion operator (i.e. `(string)f.Attribute("key")`) otherwise f.Attrbute("key").Value will NRE if the "key" attribute does not exist.

moozzyk: Consider renaming to `addElementToUpdate` or similar.

src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            if (string.IsNullOrEmpty(SourceUri))
+            {
+                Log.LogError("PackageSource must not be empty");
+                return false;
+            }
+
+            var nugetConfig = XDocument.Load(NuGetConfigPath);
+            var packageSources = nugetConfig.Element("configuration").Element("packageSources");
```
moozzyk: You can get NREs here if the file is malformed. (Not sure how important it is)

src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
moozzyk: Also this can throw - do you want to catch the exception, log and return false?

natemcmaster: If it throws, it will automatically cancel the task and return false. And the overload that takes a string isn't available on .NET Standard :-/

src/Internal.AspNetCore.BuildTools.Tasks/UpdatePackageSource.cs
natemcmaster: Yeah, this was basically a copy + paste of https://github.com/aspnet/KoreBuild/blob/dev/build/shade/_use-volatile-feed.shade. I didn't bother hardening against NREs too much as this task works fine in the Sake version.

But as long as were at it, might as well improve code qualify, eh?

moozzyk: why not given that it is just a matter of `?.` and checking the value...

PR #187, opened 2017-03-02
Turn on TreatWarningsAsErrors

pranavkm:
https://github.com/dotnet/sdk/blob/67783a8d39e1cfb0cc44252a28cb10a7bae1c63b/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.props#L103 sets it to false (by default). This props file is processed after, consequently it doesn't set it to true. We verified adding `<TreatWarningsAsErrors>false</TreatWarningsAsErrors>` in the csproj allows disabling it at a project level.

### DIFF
--- a/src/Internal.AspNetCore.Sdk/targets/Common.props
+++ b/src/Internal.AspNetCore.Sdk/targets/Common.props
@@ -9,14 +9,14 @@ Usage: this should be imported once via NuGet at the top of the file.
 
   <!-- common package options -->
   <PropertyGroup>
-    <IncludeSymbols Condition="'$(IncludeSymbols)'==''">true</IncludeSymbols>
-    <NeutralLanguage Condition="'$(NeutralLanguage)'==''">en-US</NeutralLanguage>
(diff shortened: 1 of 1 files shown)


src/Internal.AspNetCore.Sdk/targets/Common.props
```diff
@@ -22,7 +22,7 @@ Usage: this should be imported once via NuGet at the top of the file.
   <!-- common build options -->
   <PropertyGroup>
     <!-- make disabling warnings opt-out -->
-    <TreatWarningsAsErrors Condition="'$(TreatWarningsAsErrors)'==''">true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
natemcmaster: Should we remove the conditions from all default props? These can still be overriden by a user csproj.

pranavkm:
🆙 📅

natemcmaster:
8af0be43f3372770411edfde77da4d82446a2b37

Issue #162, opened 2017-02-09
Port resx file generation to MSBuild task


natemcmaster:
Basically, we can lift the code directly from https://github.com/aspnet/KoreBuild/blob/dev/build/shade/_k-generate-resx.shade and put it into an MSBuild task in Internal.AspNetCore.BuildTools.Tasks.

PR #189, opened 2017-03-06
Port generate resx files to MSBuild

natemcmaster:
This adds a target named "Resx" into individual projects to generate projects.

`dotnet msbuild /t:Resx`.

Resolves https://github.com/aspnet/BuildTools/issues/162


src/Internal.AspNetCore.BuildTools.Tasks/Utilities/IndentedTextWriter.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.IO;
+using System.Text;
+
+namespace Microsoft.AspNetCore.BuildTools.Utilities
+{
+    public class IndentedTextWriter : TextWriter, IDisposable
+    {
+        private readonly TextWriter _wrapped;
+        private readonly string _spaces;
+        private static readonly string[] _knownLengths = new[]
```
pranavkm: Really not trying to write Kestrel here 😖

natemcmaster: Haha, yeah, runtime habits. `new String(' ', length)` is probably okay.

PR #197, opened 2017-03-17
Remove unused uploader tool

natemcmaster:
We don't use this tool anywhere in our build system. Anything that needs to upload files to azure can use [Azure CLI or Azure PowerShell](https://docs.microsoft.com/en-us/azure/#pivot=sdkstools&panel=sdkstools-all)


src/PackageCacheUploader/PackageCacheUploader.csproj
```diff
@@ -1,34 +0,0 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
```
pranavkm: Where was this previously used? Not super familiar with this.

natemcmaster: It was used in Coherence-Final until we realized that we didn't want to be pushing nightly builds of our package cache to dist.asp.net. We stopped used it back in November

PR #199, opened 2017-03-20
Remove XmlPoke2 and bump Tasks package version to 1.0.1

### LINKED ISSUE #180: Remove XmlPoke2 when next update to MSBuild is available
MSBuild 15.1.0 didn't include XmlPoke for .NET Core, even though it's available in .NET Framework. We added an imitation task, XmlPoke2. We can remove this when the next update of MSBuild becomes available.

cref https://github.com/Microsoft/msbuild/issues/1731

natemcmaster:
XmlPoke is available in the MSBuild used by .NET Core CLI 2.0.0

Resolves #180


src/Internal.AspNetCore.BuildTools.Tasks/Internal.AspNetCore.BuildTools.Tasks.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionPrefix>1.0.1</VersionPrefix>
```
pranavkm: `2.0.0` since it's a breaking change?

natemcmaster: Yup, good point. Will bump.

natemcmaster: Does this mean we should also bump Internal.AspNetCore.Sdk to 2.0.0?

natemcmaster:
After thinking this through a little more, this is premature. I'll do this as part of rev-ing all tools to "2.0".

Issue #201, opened 2017-03-22
Api check tool is slow and cannot handle some projects


dougbu:
Meant to file aspnet/KoreBuild#207 here:

In the old world, API checks took 40 seconds for a single repo. Haven't measured it yet, but I suspect performance is worse now. (Improvements I imagined did not materialize.)

.NET Framework checks are a bit faster than .NET Core checks because it isn't necessary to parse the .deps.json file or calculate the entire package graph. But, that is not to say these checks are _fast_. The API check still loads assemblies for all referenced types.

Should investigate ~~two~~ three alternative approaches for loading information about the exposed APIs:
- `System.Reflection.Metadata` exposes the metadata without needing information from referenced assemblies. The downside is this namespace offers a relatively low-level / hard-to-use surface. See #198 for some upsides.
- [Roslyn Analyzers](https://github.com/dotnet/roslyn/tree/master/docs/analyzers) can examine the API surface as the compiler works. The downside is this approach shifts everything about the API checks. It could still be enabled only in command-line builds and perhaps enabled only conditionally. But, none of the existing hooks would work.
- Mono Cecil

----

`dotnet Microsoft.AspNetCore.BuildTools.ApiCheck.dll` also crashes when analyzing some projects, usually with `FileNotFoundException` or `BadImageFormatException`. These problems will remain after addressing aspnet/KoreBuild#143 due to inherent limitations of using .deps.json files and `System.Reflection`. Fortunately, these issues affect just one project across our Universe repos: Microsoft.Extensions.Caching.SqlServer in the Caching repo.

The narrow set of [the front fell off](https://www.youtube.com/watch?v=3m5qxZm_JqM) problems is somewhat minor compared to the speed of the tool.

natemcmaster:
Thoughts on Roslyn. Disadvantage: requires source code. We lose ability to run api check on binaries. Advantage: VS integration (squiggles in txt editor) and CodeFixes.

Thoughts in s.r.m.  Disadvantage: API is really low level. I dont know anyone besides the clr team that knows how to use it. Advantage: its fast.

davidfowl:
We need to use the fast thing. Hard to use is really a bad excuse. We can also look at mono cecil (it's much friendlier and fast), if it has a .NET Core version (this tool should be .NET Core only since it doesn't need to load anything).

Roslyn would be epic! CODEFIXES for all the things!

natemcmaster:
[...]

Issue #198, opened 2017-03-18
ApiCheck: use System.Reflection.Metadata


natemcmaster:
We should look into using System.Reflection.Metadata for API check. It should be faster, more flexible, less fragile.

Comparison of APIs and what they requires to produce a list of type definitions:

_System.Reflection.Assembly.Load_: 

- assembly file
- project.assets.json
- plus all the packages in NuGet cache that are listed in the assets file
- framework name
- a compatible runtime. Because it loads assembly for **execution**. ApiCheck only uses the assembly for **inspection**. This means "apicheck.exe" can only inspect .NET Framework libraries, and "dotnet apicheck" can only check .NET Core and compatible .NET Standard libraries.

_System.Reflection.Metadata.MetadataReader_:
- assembly file

natemcmaster:
cc @dougbu

dougbu:
Yeah, that sounds similar to #207 but this issue is more narrow.

dougbu:
@natemcmaster I think of this as one way to address #201. Do you disagree?

natemcmaster:
Basically the same issue. In #201 you described the problems I've encountered that led me to suggest this approach.

Eilon:
Closing for the reasons described in https://github.com/aspnet/BuildTools/issues/201.

natemcmaster:
We should re-visit this. The tool's dependency on the .NET Core runtime creates a chicken-and-egg problem when the runtime breaks.

Example from today: the API check tool has to rev in lock-step with runtime.

 - 🐣  - Mirror build attempts to upgrade .NET Core to new version that has breaking change. API check fails because it can't run on the new runtime.
     ```
    System.IO.FileNotFoundException: Could not load file or assembly 'System.Diagnostics.DiagnosticSource, Version=4.0.2.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
        File name: 'System.Diagnostics.DiagnosticSource, Version=4.0.2.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' ---> System.IO.FileNotFoundException: Could not load the specified file.
        File name: 'System.Diagnostics.DiagnosticSource'
        at System.Runtime.Loader.AssemblyLoadContext.ResolveUsingEvent(AssemblyName assemblyName)
[...]

PR #203, opened 2017-03-22
ApiCheck: extract IApiListingReader and IApiListingWriter interfaces

natemcmaster:
Factors out interfaces from ApiListingGenerator.

Lays groundwork for testing alternative implementations of API check such as System.Reflection.Metadata or Roslyn. #201 or #198. 

TODO (will not be in this PR to keep changes more easily reviewable)
 - Fully abstract the API model away from System.Reflection APIs. It's almost there, but still has TypeDescriptor.Source. Currently uses `System.Reflection.TypeInfo`, which means API check is still tied to the Assembly.Load implementation
 - Separate "exclusions" away from the "reader". Readers currently take a list of predicates as filters. This can be abstracted as well to allow exclusions to be provided in multiple formats (currently stored in JSON, but could also be specified as code comments (ideal for Roslyn + codefixes) or attributes (ideal for System.Reflection.Metadata)).

### DIFF
--- a/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
+++ b/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListingComparer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using ApiCheck.Description;
+using ApiCheck.IO;
 
 namespace ApiCheck
 {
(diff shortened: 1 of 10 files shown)


dougbu:
The highest priority is to get the tool working again and then to add exclusions (and perhaps announcements) for all of the untracked breaking changes. I was going to send that PR out tomorrow morning, as well as PRs to disable checks where there are untracked changes. This breaks that work.

Further, I'm not sure any particular rewrite requires these interfaces.

natemcmaster:
Understood. Getting it working is the higher priority. Don't let this get in the way of that. This is not urgent.

Primary motivation for abstracting to interfaces is to make it easier to experiment with and compare solutions for #201.

natemcmaster:
Closing this for now as there needs to be more discussion about whether we care enough to rewrite the API check tool. If we decide to make this investment, we can re-open. I'll snapshot this code in a fork.

PR #204, opened 2017-03-22
Remove usages of System.Reflection from the ApiListing model

natemcmaster:
Follow up to https://github.com/aspnet/BuildTools/pull/203.

The 'ApiListing' classes and related types are the model of an assembly's exported types. This model currently leaks into using System.Reflection.{TypeInfo, MemberInfo, and ParameterInfo}, which should not be used as it ties the model to the reflection-based implementation of API check. System.Reflection should be treated as the source of model, but not the model itself. 

Removing these types should not affect existing baseline.json files as they were not serialized. Also, in a few cases the reflection types were set in the model, but never read by the ApiListingComparer step.

Changes:
 - Removes usages System.Reflection from the API model
 - Moves static helper methods for constructing the API model from reflection info into 'ReflectionHelper'

### DIFF
--- a/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListing/ApiListing.cs
+++ b/src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListing/ApiListing.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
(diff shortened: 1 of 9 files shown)


natemcmaster:
Closing this for now as there needs to be more discussion about whether we care enough to rewrite the API check tool. If we decide to make this investment, we can re-open. I'll snapshot this code in a fork.

Issue #200, opened 2017-03-20
Tools on .NET Core 2.0


natemcmaster:
Will be a breaking change incompatible with .NET Core 1.0 CLI. Requires bumping tools package versions to 2.0.

Alternatively, we can make tasks run in-proc with MSBuild so they are runtime-agnostic. Would still require a 2.0 version bump.

natemcmaster:
Btw, makes me happy that this is issue #200.

natemcmaster:
cc @dougbu

dougbu:
Removing myself here because this work isn't as related to getting ApiCheck.exe working again as we thought (see #205). In fact, we _cannot_ move ApiCheck.exe to netcoreapp2.0 until it is guaranteed to run using the exact target 2.0.0-* shared runtime or newer. (Probably need to build that executable against a specific shared runtime version that is guaranteed to be older than what KoreBuild targets that day.)

See branch `dougbu/netcoreapp2.0` for a commit that does part of the upgrade. Doesn't currently build because it contains changes originally done as part of #205.

This branch also does _not_ bump the package versions. That's necessary because we don't want newer tools used when building downlevel branches.

dougbu:
On a related note, the VersionTool project is unable to consume the Microsoft.Extensions.CommandLineUtils.Sources package. It exposes `public` properties of some newly-`internal` types. Should address this while dealing with other versioning issues.

natemcmaster:
@Eilon we should do this soon. The latest 2.0 builds of CLI no longer bundle .NET Core 1.0. MSBuild and dotnet CLI itself have completely upgraded to 2.0.

Eilon:
But we can't do it soon, can we? We're blocked by the NuGet bug, no?

Eilon:
Oh wait sorry this isn't about actual CLI tools, right? This is just about our own "projects" that are "command line tools"?

natemcmaster:
Right. This is about our command line apps (e.g. NuGetPackageVerifier), not our DotnetCliToolReference packages (e.g. dotnet-watch).

Eilon:
@dougbu said:
> In fact, we cannot move ApiCheck.exe to netcoreapp2.0 until it is guaranteed to run using the exact target 2.0.0-* shared runtime or newer.

What exactly does this mean? Is the concern that .NET Core 2.0 isn't stable enough yet?

dougbu:
> What exactly does this mean?

[...]

Issue #121, opened 2016-11-10
Support netcoreapp1.1 tfm


natemcmaster:
The dependency graph of apps targeting netcoreapp1.1 will be slightly different from those using netcoreapp1.0. Build tools should account for this new TFM.

I found two instances where this might apply, and we should investigate if netcoreapp1.1 makes a difference:
SplitPackages only check compatibility with netcoreapp1.0.
DependencyPackager only crossgen's for netcoreapp1.0.

natemcmaster:
+ netcoreapp1.2  cref https://github.com/NuGet/Home/issues/3576

natemcmaster:
I'm not sure how relevant "netcoreapp1.1" is anymore, but there may be related work for .NET Core 2.0. 

@Eilon do we expected ASP.NET Core 2.0 packages to be compatible with the netcoreapp1.0 TFM? i.e. we are keeping netstandard1.3 where possible. Will packages that upgrade to netstandard2.0 be required to keep a TFM that is compatible with netcoreapp1.0?

cc @DamianEdwards

Eilon:
Actually, can you remind me what this bug is about? Is it about our build tools run on, about what the tools work with and/or produce, or something else?

natemcmaster:
The bug was that package cache generation uses .NET Core 1.0, but the dependency graph for .NET Core 1.1 is different, so cross gen uses the wrong references.

Also, our coherence check only ensures that packages are compatible with .NET Framework 4.5.1 or .NET Core 1.0. I know this is changing, but not sure what the new runtime version prerequisites will be.

Eilon:
@JunTaoLuo @DamianEdwards - with the new package cache/store feature, presumably this issue doesn't apply anymore.

However, for our 2.0 wave of bits, of course the new package cache will work when running on .NET Core App 2.0, but do we expect it to work on 1.0/1.1?

JunTaoLuo:
The package cache is built for netcoreapp2.0 and it's not expected to work on 1.0/1.1. However, we can try building the package cache for netcoreapp1.1 and 1.0.

Eilon:
@DamianEdwards is that a thing? Do we need that?

DamianEdwards:
I don't really understand this issue. Can we sync in person?

JunTaoLuo:
Spoke offline, this is not a supported scenario.

PR #205, opened 2017-03-22
Add Microsoft.AspNetCore.BuildTools.ApiCheck to builds

dougbu:
- aspnet/KoreBuild#143

Add Microsoft.AspNetCore.BuildTools.ApiCheck.Task project
  - include task assembly in the Microsoft.AspNetMicrosoft.AspNetCore.BuildTools.ApiCheck package
  - add *.props and *.targets files to hook task into `Pack` target; runs just before .nupkg is created
- reference Microsoft.AspNetMicrosoft.AspNetCore.BuildTools.ApiCheck project from the Sdk project
  - requires lots of hacks; dotnet/sdk#939 workarounds fail in this scenario

Fall back to `RuntimeTargets` when `RuntimeAssemblies` yields no assemblies

Improve project-to-project reference and native assembly handling
- avoid `FileNotFoundException` in `AssemblyLoadContext.GetAssemblyName()` due to `\placeholder\` paths
- catch and ignore `BadImageFormatException` in `AssemblyLoadContext.GetAssemblyName()`
- avoid `InvalidOperationException` in `PackageGraph.RuntimeIsCompatible()` due to missing .sha512 file

Change ApiCheck.exe's options and output
- add `api-check --compact-output` option
- generally make tool's output more useful to the task
  - catch and log `FileNotFoundException`; output types and stack traces of caught `Exception`s
- remove `--framework` option; unused in .NET Framework already
- change ApiCheck.exe option from `--ApiListing` to `--api-listing` for consistency w/ other options

nits:
- remove `PackageGraph.RuntimeIsCompatible()` `candidateRuntime` parameter
  - not needed and confusingly-named; passed value included in `compatibleRuntimes`
- update solution to use current project type GUID
- ignore global.json and launchSettings.json
- make some Linq a bit more readable
- whitespace and `using` cleanup
- move all property groups above item groups in Microsoft.AspNetMicrosoft.AspNetCore.BuildTools.ApiCheck.csproj
- make a couple of constructors `private`; used only in `public static` methods
- ignore case a bit more

notes:
- don't use `UseCommandProcessor==true`; `ToolTask` writes out .cmd file with a BOM that cmd.exe hates


dougbu:
See also the following changes. They must go in first to avoid breaking every Universe build. Note these changes are temporary and should be replaced by `exceptions.net(45|core).json` fixes in short order. I'll file those issues once the PRs are in.
- aspnet/Caching#283
- aspnet/DataProtection#211
- aspnet/EntityFramework#7955
- aspnet/Hosting#980
- aspnet/HttpAbstractions#799
- aspnet/Localization#349
- aspnet/Microsoft.Data.Sqlite#336
- aspnet/Mvc#6013
- aspnet/Razor#1104
- aspnet/Security#1159

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
```diff
@@ -0,0 +1,30 @@
+﻿<!--
+  WARNING: These targets are intended for building Microsoft's ASP.NET Core repos and are not intended
+  for use outside of Microsoft.
+-->
+<Project>
+  <Target Name="ApiCheck" Condition=" '$(EnableApiCheck)' == 'true' AND $(IsPackable) == 'true' ">
```
natemcmaster: Why does IsPackable need to be enabled for ApiCheck to run?

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
```diff
@@ -0,0 +1,30 @@
+﻿<!--
```
natemcmaster: This breaks one of our design patterns for tools and targets. To keep our tools composible and to separate concerns, tools packages only deliver the tool. Integration into build cycle is specified by KoreBuild and/or Internal.AspNetCore.Sdk. 

For example, the NuGetPackageVerifier package contains the tool, but the target to light it up actually is [defined in KoreBuild](https://github.com/aspnet/KoreBuild/blob/f692ca7c8ef8a277b92aef6cfc85ffdda2beafc8/build/targets/KoreBuild.SolutionBuild.targets#L344-L346).
Likewise, the `GetGitCommitInfo` task is defined and delivered by Internal.AspNetCore.BuildTools.Tasks, but the target that ties it into MSBuild compile step is [defined in Internal.AspNetCore.Sdk](https://github.com/aspnet/BuildTools/blob/e78bac3125d8baa0791205e22b884efa9224ba2f/src/Internal.AspNetCore.Sdk/build/Git.targets).

This fix should be simple. Lift the ApiCheck target into Internal.AspNetCore.Sdk instead.

BuildTools.sln
```diff
@@ -15,34 +15,36 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{60A938B2-D95A-403C-AA7A-3683AD64DFA0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGetPackageVerifier", "src\NuGetPackageVerifier\NuGetPackageVerifier.csproj", "{657AFF5E-164E-493D-8501-8026B7C20808}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NuGetPackageVerifier", "src\NuGetPackageVerifier\NuGetPackageVerifier.csproj", "{657AFF5E-164E-493D-8501-8026B7C20808}"
```
dougbu: 🙈 Not a fan of fighting Visual Studio. We can revert this part of the change when dotnet/roslyn-project-system#1821 is fixed or wait for a future VS update to change the solution for us.

src/Microsoft.AspNetCore.BuildTools.ApiCheck.Task/Microsoft.AspNetCore.BuildTools.ApiCheck.Task.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.6</TargetFramework>
```
dougbu: 🙈 This is intentional. Our engineering team decided _not_ to run the tool within Visual Studio. We may change that after #201 is fixed.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Common.props
```diff
@@ -0,0 +1,15 @@
+﻿<!--
+  WARNING: These targets are intended for building Microsoft's ASP.NET Core repos and are not intended
+  for use outside of Microsoft.
+-->
+<Project>
+  <!-- Set nothing if handled in project file or already done in outer build. -->
+  <PropertyGroup>
+    <EnableApiCheck Condition=" '$(EnableApiCheck)' == '' AND '$(MSBuildRuntimeType)' == 'Core' ">true</EnableApiCheck>
```
dougbu: 🙈 No, we don't want this in developer's faces when they are just hacking things together.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Common.props
```diff
@@ -0,0 +1,15 @@
...
+    <EnableApiCheck Condition=" '$(EnableApiCheck)' == '' ">false</EnableApiCheck>
+    <ExcludePublicInternalTypes_InApiCheck Condition=" '$(ExcludePublicInternalTypes_InApiCheck)' == '' ">true</ExcludePublicInternalTypes_InApiCheck>
+
+    <!-- Hook API checks into Pack run, prior to creating the .nupkg but after build (if any). -->
+    <GenerateNuspecDependsOn Condition=" ! $(GenerateNuspecDependsOn.Contains('ApiCheck')) ">$(GenerateNuspecDependsOn);ApiCheck</GenerateNuspecDependsOn>
```
dougbu: Sure, I'll remove the `Condition`.

Don't understand your point about `$(BuildDependsOn)` versus `$(GenerateNuspecDependsOn)`. Failures in this target cause command-line builds to fail. Also, the intent is to check the APIs of assemblies we ship (aka pack), not everything we build.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.props
```diff
@@ -0,0 +1,15 @@
+﻿<!--
+  WARNING: These targets are intended for building Microsoft's ASP.NET Core repos and are not intended
+  for use outside of Microsoft.
+-->
+<Project>
+  <Import Project="$(MSBuildThisFileDirectory)Common.props"/>
+
+  <PropertyGroup>
+    <_ApiCheckTaskAssembly>$(MSBuildThisFileDirectory)..\tools\netstandard1.6\Microsoft.AspNetCore.BuildTools.ApiCheck.Task.dll</_ApiCheckTaskAssembly>
```
dougbu: Intentionally not part of non-Core MSBuild. But, if I left out the condition, I'll add it.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
dougbu: Why is that odd distance between tasks and where they are used a "design pattern"?

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
dougbu: Because we don't check code we don't ship.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Program.cs
```diff
@@ -59,53 +58,87 @@ public static int Main(string[] args)
...
+                    Console.WriteLine($"  {inner.GetType().FullName}: {inner.Message}");
+                    inner = inner.InnerException;
+                }
+
+                if (e.InnerException != null)
+                {
+                    Console.WriteLine("  --- End of inner exceptions ---");
+                }
 
+                Console.WriteLine(e.StackTrace);
```
dougbu: > This whole block can be simplified to `Console.WriteLine(e.ToString())`.

Unfortunately `FileNotFoundException`s that do not mention the filename sometimes bubble here. I will check if `ReflectionTypeLoadException` overrides `ToString()` to include loader exceptions in its output.

> consider using `Console.Error.WriteLine`

Why? MSBuild doesn't use `Console.Error` at all.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Common.props
natemcmaster: I disagree. We want developers aware of breaking changes. Many times their unintentionally and we don't catch them until the tool actually runs. We have developers who only build + test in VS.

But I'll let others weigh in to settle this. Should API check cause building in VS to fail? Or should it on fail on build.cmd? cc @Eilon @muratg

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Common.props
natemcmaster: Producing an and API check are orthogonal concerns. GenerateNuspecDependsOn is not guaranteed to run as a part of producing an binary.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
natemcmaster: https://en.wikipedia.org/wiki/Separation_of_concerns

Tasks != targets.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
natemcmaster: Pack and api check are separate and shouldn't be intermingled. It's possible to produce a binary can end up in a nuget package even when IsPackable is false. EnableApiCheck should be sufficient to disable the target.

The target can use existence of baseline.json files to avoid running. 
The target can use the `IsTestProject` property to suppress the warning on test projects.

src/Microsoft.AspNetCore.BuildTools.ApiCheck.Task/ApiCheckTask.cs
```diff
@@ -0,0 +1,188 @@
...
+                .SingleOrDefault();
+
+            // If above expression does not find dotnet, fall back to system PATH and hope for the best.
+            return dotnetPath?.Exists == true ? dotnetPath.FullName : ToolExe;
+        }
+
+        /// <inheritdoc />
+        protected override string GenerateCommandLineCommands()
+        {
+            var arguments = string.Empty;
```
dougbu: 🙈 Not the Right Thing:tm: for this task. `ArgumentEscaper` doesn't handle anything but spaces (not all special `bash` and `cmd` characters) and this will run in /every/ Universe build. In addition, MSBuild properties like `$(TargetPath)` are already in reasonable shape and folder names containing double quotes are incredibly unlikely.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Common.props
dougbu: 🙈 We've already had this conversation.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
dougbu: The agreed plan was to do the API Check as part of the Pack target.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Common.props
natemcmaster: I wasn't aware of a design discussion about this. Please re-consider. Making VS builds and build.cmd as similar as possible is a highly desirable goal.

If the only concern is that api check errors might impede a developer's inner loop, there are better solutions. Two come to mind: (1) make api check failures a warning in Debug but an error in Release. (2) detect "BuildingInsideVisualStudio" and demote errors to being non-fatal.

Eilon: It was discussed w/ the engineering team that we don't want this to run when building from VS.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Loader/AssemblyLoader.cs
```diff
@@ -1,33 +1,30 @@
...
-                string assemblyPath,
-                string assetsJson,
-                string framework)
+        public static Assembly LoadAssembly(string assemblyPath, string assetsJson)
         {
-            
 #if NETCOREAPP1_1
             var lockFile = new LockFileFormat().Read(assetsJson);
-            var graph = PackageGraph.Create(lockFile, framework);
+            var graph = PackageGraph.Create(lockFile, "netcoreapp1.1");
```
dougbu: As we discussed offline, moving to `netcoreapp2.0` hit some roadblocks. Would likely need to build this tool against a specific shared library version to ensure consuming projects run it in something newer. For now, leaving this as-is because it's earlier than what KoreBuild installs.

On the other hand, I have preserved the dependency version cleanup I did to get rid of package downgrades after moving every `.exe` to `netcoreapp2.0`.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Microsoft.AspNetCore.BuildTools.ApiCheck.csproj
```diff
@@ -1,40 +1,62 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <Import Project="..\..\build\common.props" />
+
   <PropertyGroup>
     <VersionPrefix>1.0.1</VersionPrefix>
```
dougbu: Agreed. Will talk to @Eilon offline about the Sdk version. That one is much more difficult to change.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/build/Microsoft.AspNetCore.BuildTools.ApiCheck.targets
dougbu: Done, mainly because consuming the Api Check package directly may be useful in some cases e.g. just to run the tool from a command line. Don't want to affect the consuming project then.

dougbu:
:up::date: with numerous changes. Have a look at recent commit titles to get some sense of what's changed.

This should be ready to go. Running another local Universe build to confirm. But, that's been hanging while cloning repos (before these tools are used) for some reason&hellip;

dougbu:
&hellip; hanging was a temporary issue with my machine. This PR is working.

src/Internal.AspNetCore.BuildTools.Tasks/Internal.AspNetCore.BuildTools.Tasks.csproj
```diff
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="NETStandard.Library" Version="1.6.0" PrivateAssets="All" />
+    <PackageReference Include="NETStandard.Library" Version="$(NetStandardImplicitPackageVersion)" PrivateAssets="All" />
```
natemcmaster: Please revert the version bumps. These versions were picked specifically to align with what is in MSBuild for .NET Core 1.0. A dependency version bump in tools packages should result in a tools version bump. This may break patch builds. We'll take care of the upgrade in #200

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Loader/AssemblyLoader.cs
natemcmaster: Let me clarify: I didn't mean we upgrade the tool to _run_ on netcoreapp2.0. Rather I was asking if we can allows this tool to parse the graph for multiple versions of netcoreapp1.0. 

More context: some other tools hardcode netcoreapp1.0 which led to #121. Hardcoding netcoreapp1.1 is repeating the same mistake. A fix for #121 will likely revert the removal of framework if you commit this as is.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Microsoft.AspNetCore.BuildTools.ApiCheck.csproj
```diff
@@ -1,40 +1,62 @@
...
     <PublishDir>$(IntermediatePackDir)$(TargetFramework)\</PublishDir>
-    <NuspecBasePath>$(IntermediatePackDir)</NuspecBasePath>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.0' ">
+    <PackageReference Include="NETStandard.Library" Version="$(NetStandardImplicitPackageVersion)" PrivateAssets="All" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netstandard1.0' ">
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="$(AspNetCoreVersion)" />
```
natemcmaster: This package isn't being produced anymore. Use Microsoft.Extensions.CommandLineUtils.Sources.

src/NugetReferenceResolver/NugetReferenceResolver.csproj
```diff
@@ -9,13 +9,9 @@
     <OutputType>library</OutputType>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <OutputType>library</OutputType>
-    <TargetFrameworks>netstandard1.6;net451</TargetFrameworks>
-  </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="$(AspNetCoreVersion)" />
```
natemcmaster: Can you add a different version variable for this one? We don't produce this package and it may not be the same as AspNetCoreVersion. For example, its [latest version](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.DotNet.PlatformAbstractions) is 2.0.0-\*, not 1.2.0-\*

src/VersionTool/VersionTool.csproj
```diff
@@ -9,9 +9,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.0.0" />
-    <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="1.0.0-preview2-1-003177" />
-    <PackageReference Include="Microsoft.DotNet.ProjectModel" Version="1.0.0-rc3-1-003177" />
+    <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="$(DotNetCliUtilsVersion)" />
+    <PackageReference Include="Microsoft.DotNet.ProjectModel" Version="$(DotNetProjectModelVersion)" />
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="$(AspNetCoreVersion)" />
```
natemcmaster: .Sources

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Microsoft.AspNetCore.BuildTools.ApiCheck.csproj
```diff
@@ -1,40 +1,62 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <Import Project="..\..\build\common.props" />
+
   <PropertyGroup>
-    <VersionPrefix>1.0.1</VersionPrefix>
-    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
+    <VersionPrefix>1.0.2</VersionPrefix>
```
natemcmaster: :+1: thanks for incrementing the version. This makes it safer to bump dependency versions.

natemcmaster:
Separate to the changes requested above, I wanted to re-iterate my concern about picking the concept of ApiCheck and producing a NuGet package. These are orthogonal concerns. While we _currently_ only API Check when we also pack with Microsoft.NET.Sdk, it may not _always_ be the way we want to use it. E.g. in current form, it would not work on https://github.com/aspnet/AspNetKatana or https://github.com/aspnet/EntityFramework6.

There are 2 simple changes I'm proposing to decouple API Check from pack targets:

1.  Chain off of Build instead of GenerateNuSpec
2.  Use `Condition="Exists('baseline.json')"` to determine if a API check should run, not `IsPackable=true`.

dougbu:
> Side note: M.Ex.CommandLineUtils doesn't exist anymore. It's now a .Sources only package cref aspnet/Common#205

That's unexpected. The .Sources package won't work for at least one project. Don't remember which one but we have tested APIs that expose `CommandLineUtils` types.

dougbu:
> These are orthogonal concerns.

I agree the concerns are orthogonal. However this is where we decided to hook the task in by default.

> in current form, it would not work on https://github.com/aspnet/AspNetKatana or https://github.com/aspnet/EntityFramework6.

All unusual repos will be able to hook the task in differently as they wish. The current approach works fine throughout Universe.

> not `IsPackable=true`

As it turns out, this condition is redundant. I'll remove it.

src/Microsoft.AspNetCore.BuildTools.ApiCheck/Loader/AssemblyLoader.cs
dougbu: Thanks for the context. I'll restore the `--framework` command-line option.

natemcmaster:
> where we decided

Was that decision made along with the information and concerns I have expressed? We are not bound to bad decisions.

dougbu:
> this bumps netcoreapp1.0 tools to 4.3.0 CoreFx packages

I agree with this part of your concern. The rest isn't valid because we have no patch branches using MSBuild. So, I'll make sure we're building for netcoreapp1.1 everywhere.

natemcmaster:
> The rest isn't valid because we have no patch branches using MSBuild.

Talked in person with @dougbu. I'm assuming this comment was after that because in person we discussed that we do have patch branches on MSbuild. aspnet/DotNetTools, aspnet/EntityFramework.Tools and aspnet/Scaffolding.

natemcmaster:
I'm dropping my requests to chain off build not pack, and to run API check in VS. I may disagree with the design, but I'm not part of the decision process so @dougbu is correct here in following the design agreed upon by the build engineering team.

dougbu:
:up::date: Rebased on `dev` and added 4 new commits. I restored ApiCheck.exe's `--framework` option, used that in `ApiCheckTask`, reverted many dependency version changes, remove an `$(IsPackable)` check, and used the `Microsoft.Extensions.CommandLineUtils.Sources` package where possible.

natemcmaster:
Nice work :)

build/dependencies.props
```diff
@@ -1,9 +1,12 @@
 ﻿<Project>
   <PropertyGroup>
     <AspNetCoreVersion>1.2.0-*</AspNetCoreVersion>
```
dougbu: I'll see how much this is used now. If it's just for the `CommandLineUtils`, I'll leave it as-is. Due to patch releases, might have to use too many versions for NuGet use to be interesting and controllable from this file.

BTW I didn't change this property or add a source to the NuGet.config file 😺

src/Internal.AspNetCore.Sdk/build/ApiCheck.targets
```diff
@@ -0,0 +1,30 @@
...
+      <_ApiExclusionsFilePath>$(MSBuildProjectDirectory)\exceptions.$(_ApiListingFileSuffix)</_ApiExclusionsFilePath>
+      <_ApiListingFile Condition=" Exists('$(_ApiListingFilePath)') ">$(_ApiListingFilePath)</_ApiListingFile>
+      <_ApiExclusionsFile Condition=" Exists('$(_ApiExclusionsFilePath)') ">$(_ApiExclusionsFilePath)</_ApiExclusionsFile>
+    </PropertyGroup>
+
+    <Message Condition=" '$(_ApiListingFile)' != '' "
+             Importance="normal"
+             Text="Checking for breaking changes in $(MSBuildProjectName), $(TargetFramework)" />
+    <ApiCheckTask Condition=" '$(_ApiListingFile)' != '' "
+                  ApiListingPath="$(_ApiListingFile)"
```
dougbu: This is how VS indents XML by default. But, I agree it won't be easy to maintain this in Notepad and VS Code.

I am _not_ changing existing indentation to match. There's a random mix in this repo.

dougbu:
- d786608
- 4f67377

PR #211, opened 2017-03-24
Add docs for tasks and implement run task

natemcmaster:
Adds a readme showing some examples of how to use build tasks.

Also, adds two tasks. Run and RunDotNet
These are general purpose process runners that are a little more friendly than Exec. The RunDotNet task defaults to running "dotnet foo" commands.


src/Internal.AspNetCore.BuildTools.Tasks/Run.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Build.Framework;
+
+namespace Microsoft.AspNetCore.BuildTools
+{
+    /// <summary>
+    /// A better version of the Exec task. See <seealso cref="RunBase" /> for more arguments.
```
pranavkm: Why are we doing this?

src/Internal.AspNetCore.BuildTools.Tasks/RunBase.cs
```diff
@@ -0,0 +1,197 @@
...
+using Microsoft.Build.Utilities;
+using Microsoft.Build.Framework;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.AspNetCore.BuildTools
+{
+    /// <summary>
+    /// Base implementation of a task that runs a process.
+    /// </summary>
+    public abstract class RunBase : Task
```
dougbu: If we need this, should inherit from `ToolTask`.

src/Internal.AspNetCore.BuildTools.Tasks/Run.cs
natemcmaster: @pranavkm I add a few things here for common gotchas of Exec.

1. Argument escaping. Exec.Command is a string, which means callers have to manually `&quot;` escape args. Run.Arguments will take care of this, and it will do it right. Wrapping in `&quot;` is not always enough to ensure arguments are escaped. For example, an arg on Windows that ends in `\`. We generally workaround by adding a space between `\` and the quote, but that's not an obivous solution to someone new to MSBuild.
2. MaxRetries. Use case: `<RunDotNet Command="nuget push package.nupkg" MaxRetries="5" />`.
3. Running dotnet commands. `RunDotNet` is shorter syntax, a combination of GetDotNetHost + Exec.

src/Internal.AspNetCore.BuildTools.Tasks/RunBase.cs
natemcmaster: Thanks for the suggestion. I'll look into it.

natemcmaster:
I wrote these with _future_ work in mind, such as the inevitable rewrite of Coherence-Final. Bear in mind that, as explained above, these are "general purpose process runners".  See also https://github.com/aspnet/BuildTools/pull/211#discussion_r107932880 for explanation of goals. 

> The new tasks aren't used anywhere.

Hard to use something that doesn't exist yet. :)

natemcmaster:
Had a discussion with Murat. I'm dropping build work to focus on runtime. If someone else wants to carry this forward, I'll leave this work in a branch.

PR #224, opened 2017-03-29
Remove Razor.Host

ryanbrandenburg:


src/NuGetPackageVerifier/already-owned-packages.txt
```diff
@@ -55,7 +55,6 @@ Microsoft.AspNetCore.Mvc.Formatters.Xml
 Microsoft.AspNetCore.Mvc.Localization
 Microsoft.AspNetCore.Mvc.Razor
 Microsoft.AspNetCore.Mvc.Razor.Extensions
-Microsoft.AspNetCore.Mvc.Razor.Host
```
pranavkm: You could leave it in. It's a list of packages we already own on NuGet.org - removing the dev version of the package shouldn't invalidate that fact.

ryanbrandenburg: Ok, will close then.

PR #225, opened 2017-03-30
[WIP] Fix package publisher: release semaphore correctly and increase…

kichalla:
… timeouts

### DIFF
--- a/src/PackagePublisher/Program.cs
+++ b/src/PackagePublisher/Program.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -20,11 +21,14 @@ namespace PackagePublisher
         private const string ApiKeyEnvironmentVariableName = "APIKey";
         private static CommandOption _packagesDirectory;
         private static CommandOption _feedToUploadTo;
-        private static CommandOption _maxRetryCountOption;
 
         // Do NOT log the key
         private static string _apiKey = null;
-        private static int _maxRetryCount = 10;
+        private const int _maxRetryCount = 5;
+        private const int _maxParallelPackagePushes = 4;
+        private static readonly TimeSpan _packagePushTimeout = TimeSpan.FromSeconds(90);
+        private static ConcurrentBag<PackageInfo> _packages = null;
+        private static readonly CancellationTokenSource _packagePushCancellationTokenSource = new CancellationTokenSource();
 
         public static int Main(string[] args)
         {
@@ -36,10 +40,6 @@ namespace PackagePublisher
             _feedToUploadTo = app.Option(
                 "-f|--feed",
                 "Feed to upload the .nupkg file(s) to",
-                CommandOptionType.SingleValue);
-            _maxRetryCountOption = app.Option(
-                "-c|--retryCount",
-                $"Max retry count to publish a package. Defaults to {_maxRetryCount}",
                 CommandOptionType.SingleValue);
             var help = app.Option("-h|--help", "Show help", CommandOptionType.NoValue);
 
@@ -73,7 +73,9 @@ namespace PackagePublisher
                         }
                     });
 
-                if (packages.Count() == 0)
+                _packages = new ConcurrentBag<PackageInfo>(packages);
+
+                if (_packages.Count == 0)
                 {
                     Console.WriteLine("No package files were found to be published!!");
                     return 1;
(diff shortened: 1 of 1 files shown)


src/PackagePublisher/Program.cs
```diff
@@ -24,7 +24,9 @@ public static class Program
 
         // Do NOT log the key
         private static string _apiKey = null;
-        private static int _maxRetryCount = 10;
+        private static int _maxRetryCount = 5;
+        private static int _packagePushTimeoutInSeconds = 90;
```
kichalla: Earlier timeout of 30 seconds for an individual package push seems less considering we have packages of size ~90mb, but I reduced the number of retries.

src/PackagePublisher/Program.cs
```diff
@@ -24,7 +24,9 @@ public static class Program
...
+        private static int _semaphoreTimeoutInSeconds = _maxRetryCount * _packagePushTimeoutInSeconds;
```
kichalla: Earlier the semaphore timeout was 3 mins (or 180 seconds) where as the worst case of a push would have been 10 * 30 seconds resulting in semaphore entry failure which resulted number of parallel pushes to be greater than 4 and also semaphore release to be called incorrectly.

src/PackagePublisher/Program.cs
```diff
@@ -24,7 +24,9 @@ public static class Program
 
         // Do NOT log the key
         private static string _apiKey = null;
-        private static int _maxRetryCount = 10;
+        private static int _maxRetryCount = 5;
```
mikeharder: Change `_packagePushTimeoutInSeconds` and `_semaphoreTimeoutInSeconds` from `private static int` to `private const int`.

src/PackagePublisher/Program.cs
pranavkm: Wouldn't we have to account for the number of packages in here? Since all the pushes queue up at the onset, you give it at most 7.5 minutes which is usually enough, but that's at most it takes for one package to publish, not the 89 other packages also waiting to be pushed.

mikeharder: First, I would change both of these from `int *InSeconds` to `TimeSpan *`.  I agree the value should be at least `_maxRetryCount * _packagePushTimeout * NumPackages`.  Or, just choose a huge value that should always be sufficient for all packages to finish, say 1 hour.

src/PackagePublisher/Program.cs
```diff
@@ -20,13 +20,12 @@ public static class Program
...
         private static CommandOption _feedToUploadTo;
-        private static CommandOption _maxRetryCountOption;
 
         // Do NOT log the key
         private static string _apiKey = null;
         private static int _maxRetryCount = 5;
-        private static int _packagePushTimeoutInSeconds = 90;
-        private static int _semaphoreTimeoutInSeconds = _maxRetryCount * _packagePushTimeoutInSeconds;
+        private static readonly TimeSpan _packagePushTimeout = TimeSpan.FromSeconds(90);
+        private static readonly TimeSpan _semaphoreTimeout = TimeSpan.FromHours(1);
```
pranavkm: 😮

kichalla: You think its too much?

mikeharder: I think this is fine for now, let's try and see what happens.  Another approach would be to keep a queue of packages to be uploaded, then start 4 tasks which each pull a package from the queue, upload in a retry loop, and repeat until the queue is empty.  This avoids the need for a semaphore.

mikeharder: This is how I've written similar code in the past.  To me at least, it's a little simpler to understand than the blocking semaphore.

pranavkm: Yeah, that might be an better solution. Let's go with this for now and we can refactor this code later on.

src/PackagePublisher/Program.cs
```diff
@@ -20,11 +20,12 @@ public static class Program
         private const string ApiKeyEnvironmentVariableName = "APIKey";
         private static CommandOption _packagesDirectory;
         private static CommandOption _feedToUploadTo;
-        private static CommandOption _maxRetryCountOption;
 
         // Do NOT log the key
         private static string _apiKey = null;
-        private static int _maxRetryCount = 10;
+        private static int _maxRetryCount = 5;
```
mikeharder: Change to `private const _maxRetryCount` since the command line param has been removed.

src/PackagePublisher/Program.cs
```diff
@@ -23,9 +23,10 @@ public static class Program
 
         // Do NOT log the key
         private static string _apiKey = null;
-        private static int _maxRetryCount = 5;
+        private const int _maxRetryCount = 5;
+        private const int _maxParallelPackagePushes = 4;
         private static readonly TimeSpan _packagePushTimeout = TimeSpan.FromSeconds(90);
-        private static readonly TimeSpan _semaphoreTimeout = TimeSpan.FromHours(1);
+        private static ConcurrentQueue<PackageInfo> _packageQueue = null;
```
pranavkm: `ConcurrentBag` should suffice.

src/PackagePublisher/Program.cs
```diff
@@ -128,37 +131,38 @@ private static bool AreValidParameters()
...
-                var tasks = packages.Select(async package =>
+            await Task.WhenAll(tasks);
+        }
+
+        private static async Task PushPackagesAsync(PackageUpdateResource packageUpdateResource)
+        {
+            PackageInfo package = null;
+            var exceptions = new List<Exception>();
+
+            while (_packageQueue.TryDequeue(out package))
```
pranavkm: `while (_packageBag.TryTake(out var package` for 💯 points

src/PackagePublisher/Program.cs
```diff
@@ -126,65 +126,75 @@ private static bool AreValidParameters()
...
-                    {
-                        semaphore.Release();
-                    }
-                });
-                await Task.WhenAll(tasks);
+                    await PushPackageAsync(packageUpdateResource, package);
+                }
+                catch (Exception ex)
+                {
+                    // do not block from the next package to be pushed
```
mikeharder: Why keep pushing after one package failed?  Is this useful?  If not, I would just throw after one push fails (after retries).

kichalla: I believe the idea is that we could try pushing it offline in the worst case. @pranavkm can you confirm?

pranavkm: In the past, the first failure would've caused the process to exit. If a push set fails, it just makes sense to push the whole set again (either thru a rebuild or manually) rather than pick and choose some of the packages

mikeharder: If we **did** want to fail the process at the first push failure (after retry), we would let the exception flow through `PushPackagesAsync`, then replace `Task.WhenAll()` in `PublishToFeedAsync()` with the following:

```C#
for (var i=0; i < tasks.Length; i++) {
    await await Task.WhenAny(tasks);
}
```

kichalla: What have we decided? keep the code as it is for now or exit on the first failure?

pranavkm: Could we start with not catching exceptions? Keeps it simple to start with

mikeharder: I would prefer fail fast, but either way I think we should merge this PR by EOD.

mikeharder: Sounds like @pranavkm agreed that fail fast is fine.

mikeharder: If we just don't catch exceptions in `PushPackagesAsync`, we will still not fail fast since `Task.WhenAll()` will wait for the remaining tasks to drain the package queue.

kichalla: Sure, will work on the fixes.

PR #226, opened 2017-04-07
Stop cross compiling splitpackages on net451

kichalla:


src/SplitPackages/SplitPackages.csproj
```diff
@@ -4,9 +4,10 @@
   <PropertyGroup>
     <Description>Copies NuGet packages form a given source folder into a specific set of folders based on a CSV file.</Description>
     <VersionPrefix>1.0.1</VersionPrefix>
-    <TargetFrameworks>netcoreapp1.0;net451</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.0</TargetFrameworks>
     <DebugType>portable</DebugType>
     <OutputType>exe</OutputType>
+    <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
```
pranavkm: Why do we need this?

kichalla: Because we need a executable: https://github.com/aspnet/Coherence/blob/dev/build/repo.targets#L50

kichalla: Ignore my previous comment. We want a standalone executable with all its dependencies packed together otherwise when executing the tool it cannot find the references. (I ran into this issue when working on the package publisher too), but the side effect is that it increases the size of the package.

pranavkm: We aren't doing this with the verifier (https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/NuGetPackageVerifier.csproj#L8). What's different here?

kichalla: The verifier is being used as an msbuild task https://github.com/aspnet/BuildTools/tree/dev/src/NuGetPackageVerifier.Task ?

pranavkm: All it does is launch the binary though - https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier.Task/VerifyPackages.cs#L45-L55.

pranavkm:
We need to bump up the package version for tools that changed before this goes in. I reverted your commit

kichalla:
c11946aa85b477208f7ba836cc4a398a4afea05e

PR #228, opened 2017-04-08
Remove unused variables

smitpatel:


src/Microsoft.AspNetCore.BuildTools.ApiCheck/ApiListing/MemberDescriptor.cs
```diff
@@ -136,7 +136,7 @@ private IEnumerable<string> GetComponents()
             if (Kind != MemberKind.Field)
             {
                 var name = ExplicitInterface != null ? $"{ExplicitInterface}.{Name}" : Name;
-                yield return GetParametersComponent(Name);
+                yield return GetParametersComponent(name);
```
smitpatel: Note for reviewer - What is correct variable to pass here?

smitpatel: @javiercn - Perhaps you know what is the right parameter to pass here.

PR #233, opened 2017-04-13
Added 'Microsoft.Dotnet' prefix to inclusion list of package verifier

kichalla:


src/NuGetPackageVerifier/third-party.json
```diff
@@ -4,6 +4,7 @@
     "Microsoft.AspNetCore",
     "Microsoft.CodeAnalysis.Razor",
     "Microsoft.Data.Sqlite",
+    "Microsoft.DotNet",
```
kichalla: I want to use Microsoft.DotNet.PlatformAbstractions for which I need this

kichalla:
I am not doing this and instead added an exclusion to the package that doesn't ship.

PR #234, opened 2017-04-14
Detect package references containing dll's that were not used during compilation

pakrym:
Note: it does not tell you that you can safely remove package reference it just points out which dll's were not used during compilation. Package that has unused dll could have brought transitive reference that is used, content files, etc. 

Sometimes package references are unused only for one of the targets so users need to pay attention to which target produced the warning.

### DIFF
--- a/src/Internal.AspNetCore.BuildTools.Tasks/FindUnusedReferences.cs
+++ b/src/Internal.AspNetCore.BuildTools.Tasks/FindUnusedReferences.cs
@@ -58,7 +58,7 @@ namespace Microsoft.AspNetCore.BuildTools
                 StringComparer.OrdinalIgnoreCase);
 
             using (var fileStream = File.OpenRead(Assembly))
-            using (PEReader reader = new PEReader(fileStream))
+            using (var reader = new PEReader(fileStream))
             {
                 var metadataReader = reader.GetMetadataReader();
                 foreach (AssemblyReferenceHandle assemblyReferenceHandle in metadataReader.AssemblyReferences)
--- a/src/Internal.AspNetCore.BuildTools.Tasks/FindUnusedReferences.cs
+++ b/src/Internal.AspNetCore.BuildTools.Tasks/FindUnusedReferences.cs
@@ -1,10 +1,12 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection.Metadata;
 using System.Reflection.PortableExecutable;
-using System.Text;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 
--- a/src/Internal.AspNetCore.Sdk/build/FindUnusedReferences.targets
+++ b/src/Internal.AspNetCore.Sdk/build/FindUnusedReferences.targets
@@ -1,0 +1,15 @@
+﻿<!--
+  WARNING: These targets are intended for building Microsoft's ASP.NET Core repos and are not intended
+  for use outside of Microsoft.
+-->
+<Project>
+  <Target Name="FindUnusedReferences"
+          AfterTargets="CoreCompile"
+          DependsOnTargets="RunResolvePackageDependencies"
+          Condition=" '$(EnableFindUnusedReferences)' == 'true' ">
+    <Sdk_FindUnusedReferences Assembly="@(IntermediateAssembly)" References="@(ReferencePath)" Packages="@(PackageDependencies)" Files="@(FileDefinitions)" >
+      <Output TaskParameter="UnusedReferences" ItemName="UnusedReferences" />
+    </Sdk_FindUnusedReferences>
+    <Warning Condition="'@(UnusedReferences)' != ''" Text="Unused reference in $(MSBuildProjectFile)/$(TargetFramework) %(UnusedReferences.Identity) from package %(UnusedReferences.PackageName)" />
+  </Target>
+</Project>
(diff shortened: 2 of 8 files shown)


pakrym:
Sample output:

```
C:\Users\pakrym\.nuget\packages\internal.aspnetcore.sdk\2.0.0-rc2-15234\build\FindUnusedReferences.targets(13,5): warning : Unused reference in Microsoft.Extensions.Configuration.Test.Common.csproj/net46 Microsoft.Extensions.FileProviders.Physical/2.0.0-preview1-24445/lib/netstandard1.3/Microsoft.Extensions.FileProviders.Physical.dll from package Microsoft.Extensions.FileProviders.Physical [D:\github\aspnet\Configuration\test\Microsoft.Extensions.Configuration.Test.Common\Microsoft.Extensions.Configuration.Test.Common.csproj]
  Microsoft.Extensions.Configuration.Test.Common -> D:\github\aspnet\Configuration\test\Microsoft.Extensions.Configuration.Test.Common\bin\Debug\net46\Microsoft.Extensions.Configuration.Test.Common.dll
C:\Users\pakrym\.nuget\packages\internal.aspnetcore.sdk\2.0.0-rc2-15234\build\FindUnusedReferences.targets(13,5): warning : Unused reference in Microsoft.Extensions.Configuration.AzureKeyVault.csproj/net46 System.Reflection.Metadata/1.4.2/lib/portable-net45+win8/System.Reflection.Metadata.dll from package System.Reflection.Metadata [D:\github\aspnet\Configuration\src\Microsoft.Extensions.Configuration.AzureKeyVault\Microsoft.Extensions.Configuration.AzureKeyVault.csproj]
```

src/Internal.AspNetCore.Sdk/build/FindUnusedReferences.targets
```diff
@@ -0,0 +1,15 @@
...
+-->
+<Project>
+  <Target Name="FindUnusedReferences"
+          AfterTargets="CoreCompile"
+          DependsOnTargets="RunResolvePackageDependencies"
+          Condition=" '$(EnableFindUnusedReferences)' == 'true' ">
+    <Sdk_FindUnusedReferences Assembly="@(IntermediateAssembly)" References="@(ReferencePath)" Packages="@(PackageDependencies)" Files="@(FileDefinitions)" >
+      <Output TaskParameter="UnusedReferences" ItemName="UnusedReferences" />
+    </Sdk_FindUnusedReferences>
+    <Warning Condition="'@(UnusedReferences)' != ''" Text="Unused reference in $(MSBuildProjectFile)/$(TargetFramework) %(UnusedReferences.Identity) from package %(UnusedReferences.PackageName)" />
```
pranavkm: Would it be possible to make this a verifier rule? We really don't have a good way to suppress individual warnings here and I rather not spend time every compilation doing extra work.

natemcmaster:
How many warnings do you get when building Universe with this change? I'm concerned this will further clutter our build output with warnings that are not actionable.

I think with a few extra checks, this target could make strong recommendations about whether a package is safe to remove. Those additional checks might include:

 - listed as a PackageReference? 
 - build targets or content files actually in use? 
 - are its transitive references in use?

src/Internal.AspNetCore.Sdk/build/FindUnusedReferences.targets
pakrym: Not really, I need list of things that got passed to compiler and it's not easy to infer from assets.json anymore :(.

pakrym:
@natemcmaster tons of warnings right now, this definitely needs to be disabled for tests and samples. Maybe this should be run only occasionally to cleanup libraries and not on every run, or it need a way to specify exclusions. 

1. It always looks only on direct package reference
2. I think that there are to many ways for package to be involved in build pipeline and it's to hard to track them all, my goal was not to make a tool that gives definite result but just points to a possible problem
3. Sometimes it's good to know that you can lower package reference to it's transitive dependency

natemcmaster:
+1 reduce noise.

 - make it a diagnostic that only shows up if you run `/verbosity:Normal`. Most ppl won't care or won't know what to do with the info
- or, disable by default. Require setting `/p:FindUnusedReferences=true`

kichalla:
> Sometimes it's good to know that you can lower package reference to it's transitive dependency

+1

Does this give warning in the scenario where I have a direct package reference but is not really required to be specified as its going to be brought transitively by another direct package reference that I have? (I feel VS should have built-in support for this)

pakrym:
@kichalla no. It gives a warning when you reference a package but don't use it's dll, sometimes it happens that you at the same time use dll of transitive dependency in brings, so you can replace package reference with it's transitive dependency.

pakrym:
This is disabled by default.

pranavkm:
Is there a lot of utility in this if it's disabled by default?

pakrym:
I will do https://github.com/aspnet/Coherence/issues/175 by enabling it per repo maybe fix more bugs and if we feel it gives useful results I'm going to enable it by default.

src/Internal.AspNetCore.BuildTools.Tasks/FindUnusedReferences.cs
```diff
@@ -0,0 +1,82 @@
...
+            var references = new HashSet<string>(References.Select(item => item.ItemSpec), StringComparer.OrdinalIgnoreCase);
+            var referenceFiles = Files.Where(file => references.Contains(file.GetMetadata("ResolvedPath")))
+                .ToDictionary(item => Path.GetFileNameWithoutExtension(item.GetMetadata("ResolvedPath")), StringComparer.OrdinalIgnoreCase);
+
+            var directReferences = new HashSet<string>(
+                Packages.Where(p => string.IsNullOrEmpty(p.GetMetadata("ParentPackage"))).Select(i => i.ItemSpec),
+                StringComparer.OrdinalIgnoreCase);
+
+            using (var fileStream = File.OpenRead(Assembly))
+            using (PEReader reader = new PEReader(fileStream))
```
pranavkm: `var` all day everyday.

PR #239, opened 2017-04-21
Make location of PackagePublisher availabe as an MSbuild property

natemcmaster:
See https://github.com/aspnet/KoreBuild/issues/216.

### DIFF
--- a/src/PackagePublisher/PackagePublisher.nuspec
+++ b/src/PackagePublisher/PackagePublisher.nuspec
@@ -8,6 +8,5 @@
   </metadata>
   <files>
     <file src="netcoreapp1.0\" target="/" />
-    <file src="netcoreapp1.0\build\*.targets" target="/build" />
   </files>
 </package>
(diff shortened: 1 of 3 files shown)


pakrym:
LGTM, is this the only thing causing Import warnings?

natemcmaster:
The warning is caused by using `<Project Sdk="Microsoft.NET.Sdk">` in the repo.targets file in aspnet/Universe. Microsoft.NET.Sdk was imported to make available the ResolveDependencies task which reasons about project.assets.json and the nuget cache. Adding this property makes all that unnecessary as the package itself makes a property available pointing to the right file location.

https://github.com/aspnet/Universe/pull/503/files#diff-e45ce6afc81adf6a66a1064a7eeb7ca1R1

src/PackagePublisher/PackagePublisher.nuspec
```diff
@@ -8,5 +8,6 @@
   </metadata>
   <files>
     <file src="netcoreapp1.0\" target="/" />
+    <file src="netcoreapp1.0\build\*.targets" target="/build" />
```
Eilon: What files end up in `netcoreapp1.0\build\*.targets` that then end up in the NUPKG for this?

natemcmaster: Nothing. Can be removed. Can't remember why I added it even though I wrote this last night

Issue #140, opened 2017-01-03
Reenable AssemblyIsBuiltInReleaseConfigurationRule


pranavkm:
```
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
  Parameter name: assemblyResolver
     at Mono.Cecil.MetadataResolver..ctor(IAssemblyResolver assemblyResolver)
     at Mono.Cecil.ModuleDefinition.get_MetadataResolver()
     at Mono.Cecil.ModuleDefinition.Resolve(TypeReference type)
     at Mono.Cecil.Mixin.CheckedResolve(TypeReference self)
     at Mono.Cecil.SignatureReader.ReadCustomAttributeEnum(TypeReference enum_type)
     at Mono.Cecil.SignatureReader.ReadCustomAttributeElement(TypeReference type)
     at Mono.Cecil.SignatureReader.ReadCustomAttributeConstructorArguments(CustomAttribute attribute, Collection`1 parameters)
     at Mono.Cecil.MetadataReader.ReadCustomAttributeSignature(CustomAttribute attribute)
     at Mono.Cecil.CustomAttribute.<Resolve>b__35_0(CustomAttribute attribute, MetadataReader reader)
     at Mono.Cecil.ModuleDefinition.Read[TItem,TRet](TItem item, Func`3 read)
     at Mono.Cecil.CustomAttribute.get_ConstructorArguments()
     at NuGetPackageVerifier.Rules.AssemblyIsBuiltInReleaseConfiguraitonRule.HasReleaseConfiguration(Collection`1 assemblyAttributes)
     at NuGetPackageVerifier.Rules.AssemblyIsBuiltInReleaseConfiguraitonRule.<ValidateAttribute>d__0.MoveNext()
     at NuGetPackageVerifier.Rules.SigningVerificationCompositeRule.<Validate>d__1.MoveNext()
     at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
     at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
     at NuGetPackageVerifier.PackageAnalyzer.AnalyzePackage(PackageAnalysisContext context)
     at NuGetPackageVerifier.Program.Main(String[] args)
\dnx.msbuild(178,9): error MSB3073: The command "dotnet \tools\commands\NuGetPackageVerifier\NuGetPackageVerifier.dll "\artifacts\Signed\Packages\ " "\tools\NuGetPackageVerifier.json"" exited with code 255.
```

[...]

PR #242, opened 2017-04-25
Re-enabling AssemblyIsBuiltInReleaseConfigurationRule

ajaybhargavb:
Issue - #140 

@pranavkm @Eilon @pakrym 

Needed to target netcoreapp2.0 here because `ICustomAttributeTypeProvider` was internal for 1.0

### DIFF
--- a/src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
+++ b/src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
@@ -28,6 +28,7 @@ namespace NuGetPackageVerifier.Rules
                         using (var packageFileStream = context.PackageReader.GetStream(currentFile))
                         using (var stream = new MemoryStream())
                         {
+                            // packageFileStream is not a seekable stream. So wrap it in a memory stream for PEReader to consume.
                             packageFileStream.CopyTo(stream);
                             stream.Position = 0;
                             using (var peReader = new PEReader(stream))
--- a/src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
+++ b/src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -68,7 +68,7 @@ namespace NuGetPackageVerifier.Rules
 
         private class NullProvider : ICustomAttributeTypeProvider<object>
         {
-            public static NullProvider Instance => new NullProvider();
+            public static NullProvider Instance { get; } = new NullProvider();
 
             public object GetPrimitiveType(PrimitiveTypeCode typeCode) => null;
             public object GetSystemType() => null;
--- a/src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
+++ b/src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
@@ -16,26 +16,20 @@ namespace NuGetPackageVerifier.Rules
         // TODO: Revert to using Mono.Cecil when https://github.com/jbevain/cecil/issues/306 is fixed.
         public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
         {
+            var issues = new List<PackageVerifierIssue>();
             foreach (var currentFile in context.PackageReader.GetFiles())
             {
                 var extension = Path.GetExtension(currentFile);
                 if (extension.Equals(".dll", StringComparison.OrdinalIgnoreCase) ||
                     extension.Equals(".exe", StringComparison.OrdinalIgnoreCase))
                 {
-                    var assemblyPath = Path.ChangeExtension(
-                        Path.Combine(Path.GetTempPath(), Path.GetTempFileName()), extension);
-
                     try
                     {
                         using (var packageFileStream = context.PackageReader.GetStream(currentFile))
-                        using (var fileStream = new FileStream(assemblyPath, FileMode.Create))
+                        using (var stream = new MemoryStream())
                         {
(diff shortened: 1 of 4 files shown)


src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -1,45 +1,90 @@
...
+
+                                        break;
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    finally
+                    {
+                        if (File.Exists(assemblyPath))
```
pakrym: Do you still need this?

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -1,45 +1,90 @@
...
+                if (extension.Equals(".dll", StringComparison.OrdinalIgnoreCase) ||
+                    extension.Equals(".exe", StringComparison.OrdinalIgnoreCase))
+                {
+                    var assemblyPath = Path.ChangeExtension(
+                        Path.Combine(Path.GetTempPath(), Path.GetTempFileName()), extension);
+
+                    try
+                    {
+                        using (var packageFileStream = context.PackageReader.GetStream(currentFile))
+                        using (var fileStream = new FileStream(assemblyPath, FileMode.Create))
```
pakrym: Why do you need to create actual file?

ajaybhargavb: To check if the assembly is managed. https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/AssemblyHelpers.cs#L17

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -1,45 +1,90 @@
...
+                        {
+                            packageFileStream.CopyTo(fileStream);
+                        }
+
+                        if (AssemblyHelpers.IsAssemblyManaged(assemblyPath))
```
ajaybhargavb: PE reader doesn't throw during creation but throws at `GetMetadataReader()` with System.InvalidOperationException: 'PE image does not have metadata.'

pranavkm: Can we try catch the `InvalidPEAssemblyException` and continue instead of extracting the file?

ajaybhargavb:
🆙 📅

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -1,45 +1,83 @@
...
             {
-                yield return PackageIssueFactory.AssemblyHasIncorrectBuildConfiguration(currentFilePath);
+                var extension = Path.GetExtension(currentFile);
+                if (extension.Equals(".dll", StringComparison.OrdinalIgnoreCase) ||
+                    extension.Equals(".exe", StringComparison.OrdinalIgnoreCase))
+                {
+                    try
+                    {
+                        using (var packageFileStream = context.PackageReader.GetStream(currentFile))
+                        using (var stream = new MemoryStream())
```
ajaybhargavb: Needed this because `packageFileStream` does not support seek operation

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -1,45 +1,83 @@
...
         {
-            // todo remove "Cecil throws ArgumentException when evaluating constructor arguments"
-            context.Logger.Log(LogLevel.Info, $"{nameof(AssemblyIsBuiltInReleaseConfigurationRule)} skipped");
-            return Enumerable.Empty<PackageVerifierIssue>();
-        }
-
-        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(string currentFilePath, AssemblyDefinition assembly, Collection<CustomAttribute> assemblyAttributes)
-        {
-            if (!HasReleaseConfiguration(assemblyAttributes))
+            var issues = new List<PackageVerifierIssue>();
```
ajaybhargavb: Needed this because `yield return` can't be used when there is a catch block.

ajaybhargavb:
Ping @pranavkm @pakrym

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
pranavkm: Does PEReader require a seekable stream? You're not doing seeks in this code

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -1,45 +1,83 @@
...
 
-        private static bool HasReleaseConfiguration(Collection<CustomAttribute> assemblyAttributes)
+        private class NullProvider : ICustomAttributeTypeProvider<object>
         {
-            var foundAttr = assemblyAttributes.SingleOrDefault(attr => attr.AttributeType.FullName == typeof(DebuggableAttribute).FullName);
-            if (foundAttr == null)
-            {
-                return false;
-            }
+            public static NullProvider Instance => new NullProvider();
```
pranavkm: `public static NullProvider Instance { get; } = new NullProvider();` if you actually want this to be a singleton

src/NuGetPackageVerifier/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
ajaybhargavb: >Does PEReader require a seekable stream?

Yes

Eilon: Recommend adding a comment for this in the code

ajaybhargavb:
@Eilon turns out I didn't need to copy any code. They had a public api in `netcoreapp2.0` so I just had to update the target framework 😄

Issue #166, opened 2017-02-09
NuGetPackageVerifier rules are required to be stateless per package, but aren't


Eilon:
E.g. here: https://github.com/aspnet/BuildTools/blob/ddf2ab467171df1d022ded7bfa6c80ccd7dc285b/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs#L14-L15

Rules are singletons, so they must not have per-package state. But some rules are storing state about the package they are verifying.

In practice, this isn't a problem, because everything is single-threaded, and all state is always overwritten. But this code is very fragile and could easily break.

PR #245, opened 2017-05-02
Remove state from Rules

jbagga:
Addresses #166 

cc @Eilon @ajaybhargavb


src/NuGetPackageVerifier/Rules/AssemblyHasAttributeRuleBase.cs
```diff
@@ -57,5 +64,11 @@ public virtual IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext
             string currentFilePath,
             AssemblyDefinition assembly,
             Mono.Collections.Generic.Collection<CustomAttribute> assemblyAttributes);
+
+        public abstract IEnumerable<PackageVerifierIssue> ValidateAttribute(
```
ajaybhargavb: This can be avoided. If `IPackageMetadata` is really needed by the sub classes, you can just add it to the existing `ValidateAttribute` abstract method. But I wouldn't really recommend it unless there are multiple rules that would benefit from this.

src/NuGetPackageVerifier/Rules/PackageOwnershipRule.cs
```diff
@@ -25,7 +25,6 @@ public class PackageOwnershipRule : IPackageVerifierRule
             "microsoft",
         };
         private static readonly IList<string> OwnedPackages = GetOwnedPackages();
-        private readonly HttpClient _httpClient = new HttpClient();
```
ajaybhargavb: This can just be `static`. Having a static `HttpClient` should be fine as long as we don't store any http response.

src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -6,40 +6,33 @@
 using System.Linq;
 using System.Reflection;
 using Mono.Cecil;
+using Mono.Collections.Generic;
+using NuGet.Packaging;
 using NuGet.Versioning;
 
 namespace NuGetPackageVerifier.Rules
 {
     public class PackageVersionMatchesAssemblyVersionRule : AssemblyHasAttributeRuleBase
```
ajaybhargavb: Better to have this implement `IPackageVerifierRule` instead of using `AssemblyHasAttributeRuleBase` and copy the boilerplate code into this class.

jbagga:
@ajaybhargavb Updated

src/NuGetPackageVerifier/Rules/PackageOwnershipRule.cs
```diff
@@ -25,7 +25,7 @@ public class PackageOwnershipRule : IPackageVerifierRule
...
+        private static readonly HttpClient _httpClient = new HttpClient();
```
ajaybhargavb: `_httpClient` => `HttpClient` to keep it consistent with the other `static readonly`s

jbagga:
@Eilon @ajaybhargavb ~40% less time with these changes to run the rules

src/NuGetPackageVerifier/PackageAnalysisContext.cs
```diff
@@ -18,9 +21,18 @@ public class PackageAnalysisContext : IDisposable
...
 
+        public IDictionary<string, AssemblyAttributesData> AssemblyData = new Dictionary<string, AssemblyAttributesData>();
+
         public void Dispose()
         {
             _reader?.Dispose();
         }
     }
+
+    public class AssemblyAttributesData
```
Eilon: Move type to its own file.

src/NuGetPackageVerifier/Rules/AssemblyHasAttributeHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.IO;
+using Mono.Cecil;
+
+namespace NuGetPackageVerifier.Rules
+{
+    public class AssemblyHasAttributeHelper
+    {
+        public static void GetAssemblyAttributesData(PackageAnalysisContext context)
```
Eilon: `GetXyz` methods sound like they return some data, but this one doesn't. Can we come up with a better method name for this?

src/NuGetPackageVerifier/PackageAnalysisContext.cs
```diff
@@ -18,9 +21,18 @@ public class PackageAnalysisContext : IDisposable
...
+    {
+        public AssemblyDefinition Assembly { get; set; }
```
ajaybhargavb: Where is this used?

src/NuGetPackageVerifier/Rules/AssemblyHasAttributeHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            if (context.AssemblyData.Count == 0)
```
Eilon: You can earn up to 5 Pranav Points if you use `!x.Any()` instead of `x.Count == 0`.

src/NuGetPackageVerifier/PackageAnalysisContext.cs
jbagga: To get `AssemblyVersion` [here](https://github.com/aspnet/BuildTools/blob/jbagga/StatelessNPV166/src/NuGetPackageVerifier/Rules/PackageVersionMatchesAssemblyVersionRule.cs#L49) since it isn't a custom attribute

src/NuGetPackageVerifier/Rules/AssemblyHasCommitHashAttributeRule.cs
```diff
@@ -11,18 +11,20 @@
...
-        public override IEnumerable<PackageVerifierIssue> ValidateAttribute(
-            string currentFilePath,
-            AssemblyDefinition assembly,
-            Collection<CustomAttribute> assemblyAttributes)
+        public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context)
         {
-            var fileName = Path.GetFileNameWithoutExtension(currentFilePath);
-            var isSourcesPackage = fileName.EndsWith(".Sources", StringComparison.OrdinalIgnoreCase);
-            if (!isSourcesPackage && !HasCommitHashInMetadataAttribute(assemblyAttributes))
+            AssemblyHasAttributeHelper.GetAssemblyAttributesData(context);
```
Eilon: This method stores its data in a cache on the context, so it does the work only once per package.

ajaybhargavb: Ah, missed see the `if` at the top. This is good as is 👍

jbagga: I was actually in the process of moving it to `Program.cs` [here](https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/Program.cs#L166) and [here](https://github.com/aspnet/BuildTools/blob/dev/src/NuGetPackageVerifier/Program.cs#L229)

src/NuGetPackageVerifier/Rules/AssemblyHasAttributeHelper.cs
ajaybhargavb: You can earn 5 more if you reverse the check and `return` here 😄

src/NuGetPackageVerifier/Rules/AssemblyHasCommitHashAttributeRule.cs
ajaybhargavb: That works too. The only disadvantage I can think of with that is we might populate `AssemblyData` even if no rules actually run on that. But it is a rare case and I don't know if we care enough about it. @Eilon thoughts?

Eilon: Hmm I agree that in practice it would be fine either way, but I kind of prefer it being on-demand because that better reflects the actual intentions of the code...

jbagga:
@Eilon @ajaybhargavb Updated

jbagga:
Running on the latest successful Coherence-Signed packages using the `NuGetPackageVerifier.json` found [here](https://github.com/aspnet/Coherence-Signed/blob/dev/tools/NugetPackageVerifier.json), the `dev` branch takes an average of 104,201 ms for 3 runs while the branch with these changes takes 64,388 ms. This is an improvement of ~39%.

cc @Eilon

src/NuGetPackageVerifier/PackageAnalysisContext.cs
```diff
@@ -18,6 +19,8 @@ public class PackageAnalysisContext : IDisposable
         public IPackageVerifierLogger Logger { get; set; }
         public PackageArchiveReader PackageReader => _reader ?? (_reader = new PackageArchiveReader(PackageFileInfo.FullName));
 
+        public IDictionary<string, AssemblyAttributesData> AssemblyData { get; set; } = new Dictionary<string, AssemblyAttributesData>();
```
ajaybhargavb: Drop the `set;`

PR #246, opened 2017-05-03
Update Microsoft.Azure.KeyVault version to 2.2.1-preview

JunTaoLuo:
This version is being verified by @pakrym now. Once verified and pushed to nuget we can update this version in Configuration.AzureKeyVault.

This will fix the issue https://github.com/Azure/azure-sdk-for-net/issues/3003 and will unblock https://github.com/aspnet/MetaPackages/issues/21 where we want to include this in the metapackage and runtime store.

### DIFF
--- a/src/NuGetPackageVerifier/third-party.json
+++ b/src/NuGetPackageVerifier/third-party.json
@@ -19,7 +19,7 @@
     "Microsoft.ApplicationInsights.AspNetCore": [ "2.0.0" ],
     "Microsoft.AspNet.Identity.EntityFramework": [ "2.2.1" ],
     "Microsoft.AspNet.WebApi.Client": [ "5.2.2" ],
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
@pakrym found another problem where one of the transitive dependencies of M.Azure.KeyVault, Microsoft.Rest.ClientRuntime 2.3.6 also restricts the version of Newtonsoft.Json to less than 10. We may not be able to make this change.

JunTaoLuo:
Merged in https://github.com/aspnet/BuildTools/commit/5b4ff0e5f784f33308a8d6384aa5446b5f94340c

th3morg:
This should be reopened. The issue is happening again in the latest build.

Eilon:
@th3morg can you log a new issue with details of what's happening? Please include any relevant version details, logs, error messages, stack traces, etc.

PR #249, opened 2017-05-05
Add package names to already-owned-packages

jbagga:
Some packages were missing from the already-owned-packages list.

cc @Eilon


src/NuGetPackageVerifier/already-owned-packages.txt
```diff
@@ -68,6 +80,7 @@ Microsoft.AspNetCore.Proxy
 Microsoft.AspNetCore.Razor
 Microsoft.AspNetCore.Razor.Design
 Microsoft.AspNetCore.Razor.Evolution
```
ajaybhargavb: This doesn't exist anymore. `Evolution` was renamed to `Language`. Please make sure there aren't any other non-existent packages in this list.

PR #250, opened 2017-05-05
Update AppInsights Version

pakrym:
React to version change in https://github.com/aspnet/AzureIntegration/pull/46


JunTaoLuo:
Pretty sure we don't but does this need OSS approval cc @Eilon ?

PR #252, opened 2017-05-10
Add some more Razor tooling dependencies on VS

rynowak:
Microsoft.VisualStudio.Editor
Microsoft.VisualStudio.Text.UI

Both of these at 15.0.26201 which is the VS 2017 RTW version.


rynowak:
@Eilon signed off in person

Issue #146, opened 2017-01-13
API Check should catch `abstract` additions to `public` classes


dougbu:
For example, I added the `abstract` [`ShouldValidate`](https://github.com/aspnet/Mvc/pull/5676/files#diff-438d1905c3cf0d136b093a030abe0d41R304) property to [`ModelMetadata`](https://github.com/aspnet/Mvc/blob/90edf480917cfb4d8e67a3cdc0493f97e08f015b/src/Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelMetadata.cs) in aspnet/Mvc#5676. But, the [AppVeyor build](https://ci.appveyor.com/project/aspnetci/mvc/build/1.0.7886) and all other command-line builds succeeded.

Yes, there's a bug in @dougbu as well.

dougbu:
- 3596a1f
- aspnet/EntityFramework@284f1e6
- aspnet/HttpAbstractions@5b9e129
- aspnet/Mvc@9f4ff22

PR #259, opened 2017-05-19
Disallow new `abstract` type members

dougbu:
- #146
- also add tests of field removals

nits:
- whitespace cleanup in `ApiListingGenerator` and `ApiListingComparerTests`
- remove unused `additionalFilters` parameter from `ApiListingComparerTests.CreateApiListingDocument()`

### DIFF
--- a/test/ApiCheck.Test/ApiListingComparerTests.cs
+++ b/test/ApiCheck.Test/ApiListingComparerTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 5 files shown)


test/ApiCheck.Test/ApiListingComparerTests.cs
```diff
@@ -80,6 +79,68 @@ public void Compare_Detects_TypeGenericAritybreakingChanges_as_removal()
...
+        [InlineData("public struct ComparisonScenarios.StructToRemoveFieldsFrom")]
+        public void Compare_DetectsAllFieldRemovals(string typeToCheck)
+        {
+            // Arrange
+            var v1ApiListing = CreateApiListingDocument(V1Assembly);
+            var v2ApiListing = CreateApiListingDocument(V2Assembly);
+            var comparer = new ApiListingComparer(v1ApiListing, v2ApiListing);
+
+            // Oops. The NewInternalProperty addition is a breaking change; makes it impossible to subclass type in
+            // another assembly.
```
dougbu: @Eilon these "oops" comments are about corner cases. Worth filing bugs?

Eilon: Sorry I don't understand the comment in the code. What is "oops" about it? And what is the issue?

dougbu: API check doesn't complain when an `internal abstract` member is added though that is a significant breaking change -- prevents creation of a subclass outside the defining assembly. As I said, it's a corner case but "oops", something is slipping though&hellip;

test/ApiCheck.Test/ApiListingComparerTests.cs
```diff
@@ -80,6 +79,68 @@ public void Compare_Detects_TypeGenericAritybreakingChanges_as_removal()
...
+                new BreakingChange(
+                    typeToCheck,
+                    "public const System.Int32 ConstToChangeValue = 1",
+                    ChangeKind.Removal),
+                // Removing a const doesn't cause a binary incompatibilty but often causes problems.
+                new BreakingChange(
+                    typeToCheck,
+                    "public const System.Int32 ConstToMakeField = 2",
+                    ChangeKind.Removal),
+                // Oops. Making a field writable is not technically a breaking change.
```
dougbu: And the oops here is that API Check is disallowing a non-breaking change to a field. This one is also a corner case and it's also easy to work around (with entries in the breaking change files).

dougbu:
3596a1f

Issue #266, opened 2017-06-01
ThirdParty.json shouldn't be tied to NuGetPackageVerifier


ajaybhargavb:
Right now whenever we have to add an entry to third-party.json, we need get the latest build of NPV for it to work. We should instead have the json at a different location and have NPV read from it on every run. This would make this much simpler.

@ryanbrandenburg @pakrym

pakrym:
Which different location do you have in mind?

ajaybhargavb:
I was thinking Coherence-Signed but I checked and we also run this check as part of individual builds. So not sure what would be the right location. Do you have any suggestions?

natemcmaster:
+1. We've been burned by this a few times now. We have to spin a new build of NGPV every time we need to change third-party dependencies.

PR #275, opened 2017-06-19
Add SQLitePCLRaw 1.1.6 to allowed third-party dependencies

natemcmaster:
This version of SQLitePCLRaw fixes issues with netstandard2.0 and AssetTargetFallback.

cc @Eilon  - OSS approval


src/NuGetPackageVerifier/third-party.json
```diff
@@ -41,11 +41,11 @@
...
-    "SQLitePCLRaw.bundle_green": [ "1.1.3", "1.1.5" ],
-    "SQLitePCLRaw.core": [ "1.1.3", "1.1.5" ],
-    "SQLitePCLRaw.lib.e_sqlite3.v110_xp": [ "1.1.3", "1.1.5" ],
-    "SQLitePCLRaw.lib.e_sqlite3.linux": [ "1.1.3", "1.1.5" ],
-    "SQLitePCLRaw.lib.e_sqlite3.osx": [ "1.1.3", "1.1.5" ],
+    "SQLitePCLRaw.bundle_green": [ "1.1.3", "1.1.5", "1.1.6" ],
+    "SQLitePCLRaw.core": [ "1.1.3", "1.1.5", "1.1.6" ],
+    "SQLitePCLRaw.lib.e_sqlite3.v110_xp": [ "1.1.3", "1.1.5", "1.1.6" ],
+    "SQLitePCLRaw.lib.e_sqlite3.linux": [ "1.1.3", "1.1.5", "1.1.6" ],
+    "SQLitePCLRaw.lib.e_sqlite3.osx": [ "1.1.3", "1.1.5", "1.1.6" ],
```
bricelam: Do we need the closure or can the *.lib.* ones be removed?

natemcmaster: Not sure. @pakrym does this list flatten dependencies? Or does it only check explicit `PackageReference`'s?

pakrym: It only checks direct `PackageReference`'s

natemcmaster: Ok. We can reduced to just .core and .bundle_green after https://github.com/aspnet/MetaPackages/pull/164

JunTaoLuo: Do we need the previous versions? Shouldn't only 1.1.6 be needed?

natemcmaster: If the tool versioned with our code, I would delete old versions. But in theory old builds of the code might be built with new versions of the tool. Ideally this config data won't ship in the tool. #266

natemcmaster:
Closing. This version of SQLite bundled in this update appears to includes a regression in LEFT JOIN. Let's not update until we have a fix or a different bundled version.

PR #278, opened 2017-06-23
Produce KoreBuild as a compiled artifact

natemcmaster:
- Refactored KoreBuild scripts to be reusable functions
- Removed KoreBuild's dependency on Microsoft.Common.targets
- Fixed a minor bug in ZipArchive task
- Add functional tests for a simple KoreBuild repo
- Rewrite the bootstrapper scripts to pull from Azure storage blobs

Part of https://github.com/aspnet/KoreBuild/issues/258


natemcmaster:
See https://github.com/aspnet/specs/blob/master/runtime/design-notes/2017-06-19-build-korebuild.md for the high-level description of what this is about.

.travis.yml
```diff
@@ -0,0 +1,26 @@
+language: csharp
```
natemcmaster: @pranavkm do you have the ability to enable Travis for this repo? It would be good to enable so we can ensure the functional tests pass on Linux

pranavkm: I don't think so. Perhaps @muratg might?

natemcmaster: Hmm, @muratg doesn't either, and @Eilon is OOF.

test/KoreBuild.FunctionalTests/KoreBuild.FunctionalTests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+<Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
```
pranavkm: I'm wary of doing this - if we were to bootstrap from scratch, this ends up being a cycle. Can we just copy this file from source \ use the one that's already here?

test/BuildTools.Tasks.Tests/ZipArchiveTest.cs
```diff
@@ -92,6 +92,40 @@ public void CreatesZip()
             }
         }
 
+        [Fact]
+        public void FailsForEmptyFileName()
```
pranavkm: Separate commit. Seems entirely unrelated to the KoreBuild change.

sdk/KoreBuild/modules/sharedsources/sharedsources.csproj
```diff
@@ -0,0 +1,43 @@
...
+      <Pack>true</Pack>
+      <PackagePath>$(ContentTargetFolders)\any\any\</PackagePath>
+    </EmbeddedResource>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="@(PackageReference)" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>
```
pranavkm: new lines at EOF

test/KoreBuild.FunctionalTests/KoreBuild.FunctionalTests.csproj
natemcmaster: That's a good observation, and one I plan to address in a follow-up as a part of the "version pinning" work. For now, this repo already has a bunch of dependencies on aspnetcore packages, so it will be a big change that will affect existing tools.

test/BuildTools.Tasks.Tests/ZipArchiveTest.cs
natemcmaster: I discovered this bug it while using the ZipArchive task to produce the korebuild artifact.

natemcmaster: but I'll move to separate commit to keep history cleaner

PR #290, opened 2017-07-05
Updated properties to generate nuget package's nuspec properties

kichalla:
Fixes: https://github.com/aspnet/Coherence-Signed/issues/564

- A follow up PR would be to remove this https://github.com/aspnet/Coherence-Signed/blob/dev/build/Legacy-MsBuild.targets#L83

### DIFF
--- a/src/Internal.AspNetCore.Sdk/build/Common.props
+++ b/src/Internal.AspNetCore.Sdk/build/Common.props
@@ -12,12 +12,14 @@ Usage: this should be imported once via NuGet at the top of the file.
   <PropertyGroup>
     <Authors>Microsoft</Authors>
     <Company>Microsoft Corporation.</Company>
-    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <Copyright>Copyright © Microsoft Corporation</Copyright>
     <IncludeSymbols>true</IncludeSymbols>
     <NeutralLanguage>en-US</NeutralLanguage>
     <NoPackageAnalysis>true</NoPackageAnalysis>
-    <ProjectUrl>https://asp.net</ProjectUrl>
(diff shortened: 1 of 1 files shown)


src/Internal.AspNetCore.Sdk/build/Common.props
```diff
@@ -12,12 +12,14 @@ Usage: this should be imported once via NuGet at the top of the file.
...
     <Authors>Microsoft</Authors>
     <Company>Microsoft Corporation.</Company>
-    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <Copyright>Copyright © Microsoft Corporation</Copyright>
     <IncludeSymbols>true</IncludeSymbols>
     <NeutralLanguage>en-US</NeutralLanguage>
     <NoPackageAnalysis>true</NoPackageAnalysis>
-    <ProjectUrl>https://asp.net</ProjectUrl>
-    <RequireLicenseAcceptance>true</RequireLicenseAcceptance>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/aspnet/Home/2.0.0/LICENSE.txt</PackageLicenseUrl>
```
kichalla: All these properties match as per https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/Pack.targets#L151

ajaybhargavb: I get a 404 for the above link.

kichalla: I did not create a 2.0.0 tag in Home repo yet and wanted to wait when we actually create tags for all our repos.

PR #292, opened 2017-07-05
Add push.cmd 

natemcmaster:
Add push command for CI. This keeps all "push" logic separate from "build".

Ensures 
 - we don't mistakenly push an update for the wrong KoreBuild channel
 - adds `-WhatIf` support for local testing
 - sets `-WhatIf:$true` when TeamCity is building a personal build


natemcmaster:
:bell:

PR #299, opened 2017-07-11
Update AzureKeyVaultVersion to 2.3.2

pakrym:
https://github.com/aspnet/Configuration/issues/691


JunTaoLuo:
Maybe I misspoke, I'm not sure if OSS is needed here. I think I was conflating our updates to sqlite third party dependencies. Might as well confirm that though.

Eilon:
Correct, no OSS requirement here.

PR #301, opened 2017-07-11
Allow configuring KoreBuild from the version.props file

natemcmaster:
Currently, when we branch we have to touch our build.sh and build.ps1 scripts to change which version of KoreBuild is acquired. This keeps the command-line flag overrides but moves the default channel into version.props

Expected usage:
```xml
<Project>
  <PropertyGroup>
     <!-- defaults to 'dev' in the script if not specified in version.props -->
     <KoreBuildChannel>rel/2.0.0</KoreBuildChannel>
     <!-- defaults to 'https://aspnetcore.blob.core.windows.net/buildtools' in the script if not specified in version.props -->
     <KoreBuildToolsSource>C:/buildtools/</KoreBuildToolsSource>
     <VersionPrefix>2.0.0</VersionPrefix>
     <VersionSuffix>rtm</VersionSuffix>
   </PropertyGroup>
</Project>
```

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -12,16 +12,19 @@ Downloads korebuild if required. Then builds the repository.
 The folder to build. Defaults to the folder containing this script.
 
 .PARAMETER Channel
-The channel of KoreBuild to download.
+The channel of KoreBuild to download. Overrides the value from the config file.
 
 .PARAMETER DotNetHome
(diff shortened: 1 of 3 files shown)


natemcmaster:
>  I assume the build.ps1 and build.sh are examples for other repos to copy from?

Yes. When we're ready to roll this out, we'll replace their contents to match these files.

PR #303, opened 2017-07-12
Update MSBuild to 15.3.0-*

mikeharder:
- Matches version shipping with Visual Studio 15.3

### DIFF
--- a/sdk/KoreBuild/msbuild/KoreBuild.RepoTasks.Sdk/Sdk/Sdk.props
+++ b/sdk/KoreBuild/msbuild/KoreBuild.RepoTasks.Sdk/Sdk/Sdk.props
@@ -10,7 +10,8 @@
       The dependency versions here don't matter as long as the version is binary compatible with the version of MSBuild and NuGet
       already embedded in Microsoft.NET.Sdk. These are essentially reference assemblies.
     -->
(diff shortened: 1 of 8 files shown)


sdk/KoreBuild/msbuild/KoreBuild.RepoTasks.Sdk/Sdk/Sdk.props
```diff
@@ -10,7 +10,8 @@
       The dependency versions here don't matter as long as the version is binary compatible with the version of MSBuild and NuGet
       already embedded in Microsoft.NET.Sdk. These are essentially reference assemblies.
     -->
-    <MSBuildPackagesVersion>15.3.0-preview-000388-01</MSBuildPackagesVersion>
+    <MicrosoftBuildVersion>15.3.0-*</MicrosoftBuildVersion>
```
natemcmaster: Let's leave this one hardcoded instead of floating. The version doesn't really matter here as they are just reference assemblies.

mikeharder: But why hardcode to a random version instead of using `15.3.0-*` which we use pretty much everywhere else?

PR #304, opened 2017-07-12
Update ApplicationInsights version

JunTaoLuo:
Updating to latest version of ApplicationInsights.

### DIFF
--- a/src/NuGetPackageVerifier/third-party.json
+++ b/src/NuGetPackageVerifier/third-party.json
@@ -16,8 +16,8 @@
   "prefixExclusionList": [ ],
 
   "packages": {
-    "Microsoft.ApplicationInsights": [ "2.4.0-beta3", "2.4.0-beta5" ],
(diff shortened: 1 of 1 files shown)


src/NuGetPackageVerifier/third-party.json
```diff
@@ -17,7 +17,7 @@
 
   "packages": {
     "Microsoft.ApplicationInsights": [ "2.4.0-beta3", "2.4.0-beta5" ],
```
pakrym: Do you have any idea who uses this directly?

JunTaoLuo: I only found one instance in our repos https://github.com/aspnet/AzureIntegration/pull/75.

pakrym: I meant `Microsoft.ApplicationInsights`.

JunTaoLuo: No one, but it's pulled in transitively via M.AI.AspNetCore. I've updated the version as well but I'm not sure if we should include this in the first place. Not too familiar with how this tool works.

PR #305, opened 2017-07-12
Update MSBuild version

ryanbrandenburg:


natemcmaster:
Build fails. Where are these packages?

ryanbrandenburg:
The mirror must pass first.

PR #306, opened 2017-07-13
Port PackagePublisher to be an MSBuild task

natemcmaster:
Moved https://github.com/aspnet/Universe/pull/520 here instead

### DIFF
--- a/src/NuGet.Tasks/PushNuGetPackages.cs
+++ b/src/NuGet.Tasks/PushNuGetPackages.cs
@@ -1,0 +1,152 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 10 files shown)


src/NuGet.Tasks/PushNuGetPackages.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Framework;
+using NuGet.Common;
+using NuGet.Packaging;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using NuGet.Tasks.Internal;
+using BuildTask = Microsoft.Build.Utilities.Task;
```
pranavkm: Inline this? It's used exactly once, isn't it?

natemcmaster: will do.

PR #309, opened 2017-07-13
Use fixed build number for non-CI builds

natemcmaster:
Don't use a time-based build number for local builds to allow for restore noop.

In the 2.0 SDK, NuGet restore will no-op if there have been no changes to the dependency spec. For local builds, this time-based build number invalidates the restore cache each time build.ps1 one. This increases restore time that is often unnecessary.

Comparison on aspnet/Mvc, with no new dependencies to install:
Restore w/ time-based buildnumber: ~44 seconds 
Restore w/ fixed build number: ~20 seconds.

### DIFF
--- a/sdk/KoreBuild/KoreBuild.Common.props
+++ b/sdk/KoreBuild/KoreBuild.Common.props
@@ -39,7 +39,6 @@ Default layout and configuration.
     <IntermediateDir>$(RepositoryRoot)obj\</IntermediateDir>
   </PropertyGroup>
 
-
   <!-- Use build number from CI if available -->
   <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
     <!--
@@ -50,8 +49,19 @@ Default layout and configuration.
   </PropertyGroup>
 
   <!-- Create a temporary build number if not assigned by the CI server -->
+  <PropertyGroup Condition=" '$(BuildNumber)' == '' AND '$(IncrementalVersion)' == 'true' ">
+    <!--
+      Create second-based build number for local builds.
+      635556672000000000 is Jan 1, 2015.
+    -->
+    <_SecondBasedTimeStamp>$([System.DateTime]::UtcNow.Subtract($([System.DateTime]::FromBinary(635556672000000000))).TotalSeconds.ToString("F0"))</_SecondBasedTimeStamp>
+    <_SecondBasedTimeStamp>t$([System.Int64]::Parse($(_SecondBasedTimeStamp)).ToString("x9"))</_SecondBasedTimeStamp>
+    <BuildNumber>$(_SecondBasedTimeStamp)</BuildNumber>
+  </PropertyGroup>
+
   <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
-    <BuildNumber>dev-0000</BuildNumber>
+    <BuildNumber>t000</BuildNumber>
+    <UsingLocalBuildNumber>true</UsingLocalBuildNumber>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/sdk/KoreBuild/KoreBuild.Common.props
+++ b/sdk/KoreBuild/KoreBuild.Common.props
@@ -51,13 +51,7 @@ Default layout and configuration.
 
   <!-- Create a temporary build number if not assigned by the CI server -->
   <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
-    <!--
-      Create second-based build number for local builds.
-      635556672000000000 is Jan 1, 2015.
-    -->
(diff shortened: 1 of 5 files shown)


sdk/KoreBuild/KoreBuild.Common.props
```diff
@@ -51,13 +51,7 @@ Default layout and configuration.
...
   <!-- Create a temporary build number if not assigned by the CI server -->
   <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
-    <!--
-      Create second-based build number for local builds.
-      635556672000000000 is Jan 1, 2015.
-    -->
-    <_SecondBasedTimeStamp>$([System.DateTime]::UtcNow.Subtract($([System.DateTime]::FromBinary(635556672000000000))).TotalSeconds.ToString("F0"))</_SecondBasedTimeStamp>
-    <_SecondBasedTimeStamp>t$([System.Int64]::Parse($(_SecondBasedTimeStamp)).ToString("x9"))</_SecondBasedTimeStamp>
-    <BuildNumber>$(_SecondBasedTimeStamp)</BuildNumber>
+    <BuildNumber>dev-0000</BuildNumber>
```
pranavkm: Does keeping `$(Version)` pinned but auto-incrementing `$(PackageVersion)` in dev builds not invalidate the build graph?

natemcmaster: Nope. `$(PackageVersion)` is actually what they use to invalidate the restore operation. The reason this invalidates restore is that it gets written into the project.assets.json file. `$(PackageVersion)` is computed from `$(Version)`

ryanbrandenburg: To me it looks like what happens here is we set BuildNumber to `dev-0000`, then in repo.props, because it's set to dev-0000 we set it to a time based BuildNumber. What am I missing?

natemcmaster: This uses the time-based build number for local builds of KoreBuild, but doesn't invalidate the restore cache. This build number is in the solution build lifecycle, not the project build lifecycle.

pranavkm: Let's wait to get some feedback from the engineering on this. It'll be a change to our workflows when doing cross repo work and we should find out if the workaround is acceptable to the team at large.

natemcmaster:
I think we should hold off on this change. Although it would be good to add this, we need to improve the cross-repo work story first. This would make it more difficult for devs on our team to test their cross-repo changes.

natemcmaster:
Discussed more internally. Sounds like we want to give this a try.

sdk/KoreBuild/KoreBuild.Common.props
pranavkm: `t-0000` (because it's higher than `preview`, `rc` etc)

sdk/KoreBuild/KoreBuild.Common.props
```diff
@@ -51,13 +51,7 @@ Default layout and configuration.
 
   <!-- Create a temporary build number if not assigned by the CI server -->
   <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
-    <!--
-      Create second-based build number for local builds.
-      635556672000000000 is Jan 1, 2015.
-    -->
-    <_SecondBasedTimeStamp>$([System.DateTime]::UtcNow.Subtract($([System.DateTime]::FromBinary(635556672000000000))).TotalSeconds.ToString("F0"))</_SecondBasedTimeStamp>
-    <_SecondBasedTimeStamp>t$([System.Int64]::Parse($(_SecondBasedTimeStamp)).ToString("x9"))</_SecondBasedTimeStamp>
-    <BuildNumber>$(_SecondBasedTimeStamp)</BuildNumber>
```
pranavkm: Could we make this a property? `/p:IncrementalBuild=true` that lights this up?

natemcmaster: Incremental build means something else in MSBuild. What about `IncrementalVersion`?

pranavkm: Sure. Just as long as we make it easier for users to not have to remember build numbers 👍 .

natemcmaster: Added it.

sdk/KoreBuild/KoreBuild.Common.props
natemcmaster: Updated.

natemcmaster: Though, the build number doesn't override the versionsuffix. Builds will be `2.1.0-preview1-t000`, not `2.1.0-t000`.

PR #324, opened 2017-07-17
Fix bugs in shared sources packaging targets

natemcmaster:
The shared sources packaging targets was writing files into the korebuild installation folder instead of the project folder. This fixes this by changing the obj and bin folders for the shared sources project to be in the project folder instead.

It also ensures that the same behavior of Directory.Build.props and Directory.Build.targets will be preserved when switching from old KoreBuild to compiled KoreBuild.

### DIFF
--- a/sdk/KoreBuild/modules/sharedsources/sharedsources.csproj
+++ b/sdk/KoreBuild/modules/sharedsources/sharedsources.csproj
@@ -4,9 +4,11 @@
   <PropertyGroup>
     <BaseOutputPath>$(NuspecBasePath)bin\</BaseOutputPath>
     <BaseIntermediateOutputPath>$(NuspecBasePath)obj\</BaseIntermediateOutputPath>
+    <DirBuildPropsInRepo>$([MSBuild]::GetDirectoryNameOfFileAbove($(NuspecBasePath), Directory.Build.props))\Directory.Build.props</DirBuildPropsInRepo>
+    <DirBuildTargetsInRepo>$([MSBuild]::GetDirectoryNameOfFileAbove($(NuspecBasePath), Directory.Build.targets))\Directory.Build.targets</DirBuildTargetsInRepo>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


sdk/KoreBuild/modules/sharedsources/sharedsources.csproj
```diff
@@ -27,11 +36,11 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(NuspecBasePath)'!=''">
-    <Compile Include="$(NuspecBasePath)\**\*.cs">
+    <Compile Include="$(NuspecBasePath)\**\*.cs" Exclude="$(BaseOutputPath);$(BaseIntermediateOutputPath);">
```
ryanbrandenburg: If these `Exclude`s showed up a third time I'd say they should be a property like "NonCompilingPaths" for descriptiveness and DRY, but with just two instances it's more of a preference.

sdk/KoreBuild/modules/sharedsources/sharedsources.csproj
```diff
@@ -1,5 +1,13 @@
 <!-- this is a dummy project designed to produce contentFiles packages -->
-<Project Sdk="Microsoft.NET.Sdk">
+<Project>
+
+  <PropertyGroup>
+    <BaseOutputPath>$(NuspecBasePath)bin\</BaseOutputPath>
+    <BaseIntermediateOutputPath>$(NuspecBasePath)obj\</BaseIntermediateOutputPath>
+  </PropertyGroup>
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(NuspecBasePath), Directory.Build.props))\Directory.Build.props" Condition="Exists('$([MSBuild]::GetDirectoryNameOfFileAbove($(NuspecBasePath), Directory.Build.props))\Directory.Build.props')" />
```
ryanbrandenburg: This `GetDirectoryNameOfFileAbove` sequence and the one for `Directory.Build.targets` could benefit from a descriptive property.

natemcmaster: Updated

sdk/KoreBuild/modules/sharedsources/sharedsources.csproj
natemcmaster: updated this too

PR #326, opened 2017-07-18
Add runtime store verifier

JunTaoLuo:
Addresses https://github.com/aspnet/Coherence-Signed/issues/443 and verifies all managed binaries in the store are signed.

For 2.0.0 I propose we create a standalone tool called RuntimeStoreVerifier to reduce the size of the change. I would like to consolidate the functionality with the NugetPackageVerifier and rename it to PackageVerifier post 2.0 to simplify usage. Since this is planned to be consolidated, I didn't create corresponding MSBuild tasks for this tool.

### DIFF
--- a/BuildTools.sln
+++ b/BuildTools.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26703.0
+VisualStudioVersion = 15.0.26711.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{A4F4353B-C3D2-40B0-909A-5B48A748EA76}"
 EndProject
@@ -49,7 +49,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{76ABA507
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RepoTasks", "build\tasks\RepoTasks.csproj", "{F3CAD1F9-68DB-49EA-88A7-573AE7FFE047}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Tasks", "src\NuGet.Tasks\NuGet.Tasks.csproj", "{020ED083-4076-4711-A52B-2F89EA884F9B}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NuGet.Tasks", "src\NuGet.Tasks\NuGet.Tasks.csproj", "{020ED083-4076-4711-A52B-2F89EA884F9B}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "RuntimeStoreVerifier", "src\RuntimeStoreVerifier\RuntimeStoreVerifier.csproj", "{1B5856CC-E597-4E97-9C95-DEAC80CE724A}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "WinTrust.Sources", "WinTrust.Sources", "{B99861A9-F594-4FCF-A09F-9730A3FF9436}"
+	ProjectSection(SolutionItems) = preProject
+		src\WinTrust.Sources\WinTrust.cs = src\WinTrust.Sources\WinTrust.cs
+	EndProjectSection
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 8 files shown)


natemcmaster:
Where will this be used? And have you considered making this a RepoTask instead? I'm hesitant to add a new tool if it is only used in one place.

JunTaoLuo:
It will be used in Coherence-Signed. The reason I'm not adding it as a repo task is that this will probably be integrated with NugetPackageVerifier. The reason I'm favouring putting it here is that we avoid duplicating the WinTrust logic.

natemcmaster:
I'm not sure NugetPackageVerifier is the right tool either. That tool is designed for Nuget packages. The runtime store isn't a nuget package. I'd rather have some duplicated code than tools with only one use-case in one repo.

src/RuntimeStoreVerifier/RuntimeStoreVerifier.nuspec
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<package>
+  <metadata>
+    <id>RuntimeStoreVerifier</id>
```
natemcmaster: FYI - one of the near-team goals of the "Build KoreBuild" work I'm doing is that we will not longer install these kinds of tools via NuGet. Instead, they will be bundled as tasks in KoreBuild. But this is fine for now as the 2.0.0 Coherence-Signed branch still uses pre-flight restore.

BuildTools.sln
```diff
@@ -49,7 +49,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{76ABA507
...
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RepoTasks", "build\tasks\RepoTasks.csproj", "{F3CAD1F9-68DB-49EA-88A7-573AE7FFE047}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NuGet.Tasks", "src\NuGet.Tasks\NuGet.Tasks.csproj", "{020ED083-4076-4711-A52B-2F89EA884F9B}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NuGet.Tasks", "src\NuGet.Tasks\NuGet.Tasks.csproj", "{020ED083-4076-4711-A52B-2F89EA884F9B}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "RuntimeStoreVerifier", "src\RuntimeStoreVerifier\RuntimeStoreVerifier.csproj", "{1B5856CC-E597-4E97-9C95-DEAC80CE724A}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "WinTrust.Sources", "WinTrust.Sources", "{B99861A9-F594-4FCF-A09F-9730A3FF9436}"
+	ProjectSection(SolutionItems) = preProject
+		src\WinTrust.Sources\WinTrust.cs = src\WinTrust.Sources\WinTrust.cs
```
natemcmaster: Nit: can you put this `shared/` instead of `src/`? WinTrust.Sources isn't a csproj, right?

src/RuntimeStoreVerifier/Program.cs
```diff
@@ -0,0 +1,26 @@
+using Microsoft.Extensions.CommandLineUtils;
```
natemcmaster: Copyright headers

src/RuntimeStoreVerifier/RuntimeStoreVerifierCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace RuntimeStoreVerifier
+{
+    public class RuntimeStoreVerifierCommand
+    {
+        internal static int Execute(
+            CommandOption exclusionFileOption,
+            CommandOption verboseOption,
+            CommandArgument pathArgument)
+        {
+            var exclusions = exclusionFileOption.HasValue() ? File.ReadAllLines(exclusionFileOption.Value()) : null;
```
natemcmaster: Any reason this tool can't be a task instead? MSBuild item groups already have powerful include/exclude filtering.

JunTaoLuo:
We were considering repurposing NuGetPackageVerifier to verify more than just nuget packages and renaming it to reflect that. However, I didn't want to do that as part of this change since I'm not sure how extensively that impacts our CI. The repo tasks solution was considered but we decided to see how big of a change it is to build this as a tool. A repo task in Coherence-Signed sounds like a good alternative though.

src/RuntimeStoreVerifier/RuntimeStoreVerifierCommand.cs
JunTaoLuo: This would be an argument for doing this as a repo task

natemcmaster: This can still be an MSBuild task, regardless of where it lives (repotask or here). To make it an MSBuild task, we just need to make a few tweaks to the project and add a UsingTask somewhere.

JunTaoLuo:
Closing in favour of https://github.com/aspnet/Coherence-Signed/pull/615

PR #328, opened 2017-07-18
Rel/2.0.0

ryanbrandenburg:


ryanbrandenburg:
I notice that' it's trying to take along a commit from the PushBot, @natemcmaster is it expected that that commit wouldn't already be in dev?

natemcmaster:
I've seen this before. I think its a quirk in how GitHub displays the diff. Try merging locally first and look at what actually changed.

ryanbrandenburg:
If I do it locally I still get the extra commit, but it's on top of one with an identical name that was made 2 minutes before it. Is your new PushBot stuff making commits as part of the build without merging them forward?

ryanbrandenburg:
Talked in person, we're ok with this because for now there's no good solution that doesn't take a lot of engineering work, and it's a no-op merge anyway.

PR #331, opened 2017-07-19
Add Overwrite property to UnzipArchive

ryanbrandenburg:
Currently if you try to extract a file over an existing file it failes with an IOException.

### DIFF
--- a/test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
+++ b/test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -40,7 +40,7 @@ namespace BuildTools.Tasks.Tests
                 BuildEngine = new MockEngine(),
             };
 
-            Assert.True(task.Execute(), "Task failed");
+            Assert.True(task.Execute(), "The task failed but should have passed.");
             Assert.True(Directory.Exists(outDir), outDir + " does not exist");
             Assert.Equal(files.Length, task.OutputFiles.Length);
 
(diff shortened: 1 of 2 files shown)


test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
```diff
@@ -54,6 +54,30 @@ public void UnzipsFile()
...
+
+            var task = new UnzipArchive
+            {
+                File = dest,
+                Destination = outDir,
+                BuildEngine = new MockEngine(),
+                Overwrite = true
+            };
+
+            Assert.True(task.Execute(), "Task failed");
```
JunTaoLuo: Failed?

natemcmaster: This is the message that appears when the assertion fails. A better message might be "The task failed but should have passed".

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
```diff
@@ -54,6 +54,30 @@ public void UnzipsFile()
...
+        [Fact]
+        public void Overwrites()
+        {
+            var files = new[]
+            {
+                "a.txt",
+                "dir/b.txt"
+            };
+
+            var dest =CreateZip(files);
```
natemcmaster: Nit: formatting

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
```diff
@@ -54,6 +54,30 @@ public void UnzipsFile()
...
+            var task = new UnzipArchive
+            {
+                File = dest,
+                Destination = outDir,
+                BuildEngine = new MockEngine(),
+                Overwrite = true
+            };
+
+            Assert.True(task.Execute(), "Task failed");
+            Assert.True(task.Execute(), "Task failed on second run.");
```
natemcmaster: Would be good to assert it actually wrote to the file. Before you run the task, create `$outDir/a.txt` with contents. Assert it is overwritten as an empty file after the task runs.

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
JunTaoLuo: My bad, when I see two values in an `Assert` call, I automatically assume it's `Assert.Equals`. Yea this makes sense now.

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
```diff
@@ -54,6 +54,30 @@ public void UnzipsFile()
             }
         }
 
+        [Fact]
+        public void Overwrites()
```
JunTaoLuo: Add a test that make sure no overwrite occurs when it's set to false?

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
ryanbrandenburg: I just copied the message from the other tests here. I'll update them too.

ryanbrandenburg:
This isn't needed for rel/2.0.0 work, it's part of work to clean up some of our more obtuse MSBuild files.

ryanbrandenburg:
:up::date:

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
```diff
@@ -54,6 +54,68 @@ public void UnzipsFile()
...
+                File = dest,
+                Destination = outDir,
+                BuildEngine = new MockEngine(),
+                Overwrite = true
+            };
+
+            Directory.CreateDirectory(outDir);
+
+            // Create a.txt before trying to unzip
+            using (var file = File.CreateText(Path.Combine(outDir, "a.txt")))
```
natemcmaster: This still isn't exactly verifying the file is overwritten. Try this:
```c#
var path = Path.Combine(outDir, "a.txt");
File.WriteAllText("contents!", path);
Assert.True(task.Execute(), "The task failed but should have passed.");
Assert.Empty(File.ReadAllText(path));
```

ryanbrandenburg:
:up::date:

PR #338, opened 2017-07-24
Rename version.{props => xml} and minor bootstrapper updates

natemcmaster:
Port feedback from @pranavkm on https://github.com/aspnet/Universe/pull/521


build/common.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <Import Project="dependencies.props" />
-  <Import Project="..\version.props" />
+  <Import Project="..\version.xml" />
```
ryanbrandenburg: Is `version.props`->`version.xml` something we want to do in every repo?

natemcmaster: That's the plan. I think this helps make it clearer that the file can be used as raw XML, not just an MSBuild file.

PR #339, opened 2017-07-25
Use wget and fallback to curl

natemcmaster:
Workaround more SSL issues with Azure + Travis-CI

Example: https://travis-ci.org/aspnet/Common/jobs/257500738


pranavkm:
As long as it works :shipit:

PR #340, opened 2017-07-26
Fix all warnings and errors found by the shellscript linter

natemcmaster:
Changes:
 - fix a syntax error that will be problematic for old versions of bash
 - surround any parameter with a path in double quotes. Prevents word splitting
 - fix for undefined variables when invoking some scripts directly
 - fix --verbose flag to flow through into sub-scripts


natemcmaster:
Merging to unblock build.

PR #341, opened 2017-07-27
Make default dotnet dir consistent on Windows and Linux: ~/.dotnet

natemcmaster:
On Windows, we used to install to ~/.dotnet/x64. We did this because we were anticipating a dotnet feature that got cut. 

To simplify build tools and scripts, let's remove <arch> from the default dotnet path on Windows.

Also, I added some warning messages when the shell environment may lead to issues in which dotnet is used.


natemcmaster:
So we'll leave as is.

Issue #297, opened 2017-07-10
Ensure Internal.AspNetCore.Sdk aligns with the version of KoreBuild


natemcmaster:
Part of https://github.com/aspnet/KoreBuild/issues/239. cc @mikeharder 

We should ensure the targets and tasks imported from Internal.AspNetCore.Sdk align with the version of KoreBuild. This means we shouldn't use a floating PackageReference version.

Furthermore, this will be required to allow us to pin PackageReference versions. Because Internal.AspNetCore.Sdk itself is installed as a PackageRef, it cannot define the versions of other PackageRef's that are used.

Possible solutions:

 - use a custom SdkResolver and change the project reference to `<Sdk Name="Internal.AspNetCore.Sdk" />` in each project.
 - pin the package reference version
 - have Korebuild generate a \*.props/target file that each csproj must conditionally import

natemcmaster:
If we use a SdkResolver, we would need to write some kind of script to write into the VS folder with elevated permissions. There is no simple way to install a custom SDK resolver. https://github.com/Microsoft/msbuild/issues/2278. I'd be okay with this, though, because our devs would only need to run this once per VS installation. The SdkResolver itself doesn't need to change from build to build, or even branch to branch (as long as we don't break it).

natemcmaster:
> - pin the package reference version

This is the approach implemented in #367

PR #343, opened 2017-07-28
Remove deprecated build tools

natemcmaster:
This PR deletes the following packages that are no longer in use (on dev).

 - SplitPackages. This became a RepoTask here: https://github.com/aspnet/Coherence/blob/dev/build/tasks/CopyPackagesToSplitFolders.cs
 - PackageClassifier: a class-library project that was only used by SplitPackages
 - PackagePublisher: this because a NuGet task that is bundled in compiled korebuild and doesn't need to be installed as a PackageReference. https://github.com/aspnet/BuildTools/pull/306


src/Internal.AspNetCore.BuildTools.Tasks/module.props
```diff
@@ -2,8 +2,9 @@
 
   <PropertyGroup>
     <_BuildToolsAssembly>$(MSBuildThisFileDirectory)Internal.AspNetCore.BuildTools.Tasks.dll</_BuildToolsAssembly>
+    <_BuildTasksPrefix></_BuildTasksPrefix>
```
ryanbrandenburg: What is this property for it it's not populated?

natemcmaster: Until we get https://github.com/aspnet/BuildTools/issues/297, we have to ensure we don't get assembly conflicts between Internal.AspNetCore.Sdk and KoreBuild. In the former this is set to `<_BuildTasksPrefix>Sdk_</__BuildTasksPrefix>`

PR #344, opened 2017-07-28
Add support for /t:Noop

natemcmaster:
Per discussion we had last week.

This adds a /t:Noop target that will attempt to do as little as possible. This is useful when you want to update tools without invoking the full build, or even part of the build.

### DIFF
--- a/sdk/KoreBuild/modules/noop/module.targets
+++ b/sdk/KoreBuild/modules/noop/module.targets
@@ -1,0 +1,10 @@
+<!-- Don't do anything. Useful for scripts that need to initialize KoreBuild without building anything. -->
+<Project>
+
+  <Target Name="Noop" />
+
+  <Target Name="Cow" DependsOnTargets="Noop">
+    <Message Text="Moo!" Importance="High" />
+  </Target>
+
+</Project>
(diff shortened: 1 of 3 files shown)


smitpatel:
What exactly this target do?

natemcmaster:
This is designed to initialize build tools, but nothing else. When you run `build.cmd /t:noop`, it will

 - download KoreBuild (if required)
 - download dotnet.exe (if required)
 - download .NET Framework reference assemblies (on Linux) (if required)
 - execute msbuild /t:Noop (which does nothing)

sdk/KoreBuild/modules/noop/module.targets
```diff
@@ -0,0 +1,10 @@
+<!-- Don't do anything. Useful for scripts that need to initialize KoreBuild without building anything. -->
+<Project>
+
+  <Target Name="Noop" />
+
+  <Target Name="Cow" DependsOnTargets="Noop">
+    <Message Text="Moo!" Importance="High" />
```
anurse: Should check for existence of [`cowsay`](https://en.wikipedia.org/wiki/Cowsay) and use it. :trollface:

natemcmaster: Updated :)

anurse: This is what happens when I troll. I should just stop trolling ;P

PR #346, opened 2017-07-31
Update version of Remotion.Linq

smitpatel:
Reaction PR of https://github.com/aspnet/EntityFramework/pull/9305

### DIFF
--- a/src/NuGetPackageVerifier/third-party.json
+++ b/src/NuGetPackageVerifier/third-party.json
@@ -37,7 +37,7 @@
     "Newtonsoft.Json": [ "10.0.1" ],
     "Newtonsoft.Json.Bson": [ "1.0.1" ],
     "NuGet.Frameworks": [ "4.0.0" ],
-    "Remotion.Linq": [ "2.1.1" ],
+    "Remotion.Linq": [ "2.2.0-alpha-002" ],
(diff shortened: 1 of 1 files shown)


src/NuGetPackageVerifier/third-party.json
```diff
@@ -37,7 +37,7 @@
     "Newtonsoft.Json": [ "10.0.1" ],
     "Newtonsoft.Json.Bson": [ "1.0.1" ],
     "NuGet.Frameworks": [ "4.0.0" ],
-    "Remotion.Linq": [ "2.1.1" ],
+    "Remotion.Linq": [ "2.2.0-aplha-002" ],
```
ryanbrandenburg: Pretty sure you mean `alpha`. It sounds like the fixes an important bug for you guys, and that the maintainer says 2.2.0-RTM will be basically identical to the alpha? Just double checking because generally we don't like taking beta/alpha dependencies.

smitpatel:
The alpha version is making all tests in EF passes, which is sufficient bar for quality. cc: @anpete @divega @ajcvickers 

We want to take this dependency on alpha (or any other version it upgrades later) before releasing next preview so that we can verify stability with customer scenarios too. The owner of ReLinq is ready to release RTM version for us whenever needed (it was ready for 2.0 release too but we decided not to take it in)

PR #349, opened 2017-07-31
Output to the correct location

ryanbrandenburg:


natemcmaster:
Still hoping this will get merged upstream too. https://github.com/dotnet/cli/pull/6981

PR #352, opened 2017-08-02
Fix fallback to curl when wget is not available

JunTaoLuo:
I know I know, I should install wget. But this logic is wrong and I wasn't able to build on my machine.

### DIFF
--- a/sdk/KoreBuild/scripts/common.sh
+++ b/sdk/KoreBuild/scripts/common.sh
@@ -90,7 +90,7 @@ __fetch() {
 
     __verbose "Downloading $remote_path"
 
-    failed=false
+    local failed=false
     if __machine_has 'wget'; then
         # Try wget first as this has been more reliable than curl.
         # Travis CI frequently has TLS issues with curl on macOS
@@ -98,6 +98,8 @@ __fetch() {
         progress_bar='--quiet'
         [ "$__is_verbose" = true ] && [ -z "${PS1:-}" ] && progress_bar='--progress=bar --show-progress'
         __exec wget $progress_bar --tries 10 -O "$local_path" "$remote_path" || failed=true
+    else
+        failed=true
     fi
 
     if [ "$failed" = true ] && __machine_has 'curl'; then
--- a/scripts/bootstrapper/build.sh
+++ b/scripts/bootstrapper/build.sh
@@ -99,9 +99,11 @@ __get_remote_file() {
         return 0
     fi
 
-    failed=false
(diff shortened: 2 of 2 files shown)


natemcmaster:
Aha, I see the logic error. Can you also update https://github.com/aspnet/BuildTools/blob/dev/scripts/bootstrapper/build.sh?

JunTaoLuo:
Certainly

JunTaoLuo:
:up::date:

sdk/KoreBuild/scripts/common.sh
```diff
@@ -90,24 +90,23 @@ __fetch() {
...
-    failed=false
+    local succeeded=false
     if __machine_has 'wget'; then
         # Try wget first as this has been more reliable than curl.
         # Travis CI frequently has TLS issues with curl on macOS
         # Only show progress bar if shell is interactive
         progress_bar='--quiet'
         [ "$__is_verbose" = true ] && [ -z "${PS1:-}" ] && progress_bar='--progress=bar --show-progress'
-        __exec wget $progress_bar --tries 10 -O "$local_path" "$remote_path" || failed=true
+        __exec wget $progress_bar --tries 10 -O "$local_path" "$remote_path" && succeeded=true
```
natemcmaster: The problem with && is that if wget fails, it won't actually fallback to curl. The first part of this check will fail and cause the script to exit because we have `set -e` at the top.

sdk/KoreBuild/scripts/common.sh
```diff
@@ -90,24 +90,23 @@ __fetch() {
...
     fi
 
-    if [ "$failed" = true ] && __machine_has 'curl'; then
-        failed=false
+    if [ "$succeeded" = false ] && __machine_has 'curl'; then
```
natemcmaster: Another way to solve this is to add an "else" to the wget block.
```sh
if __machine_has 'wget'; then
   # ...
else
    failed=true
fi
```

sdk/KoreBuild/scripts/common.sh
JunTaoLuo: uh oh. I already updated all the other repos after you approved the MetaPackage one. I didn't notice the set -e at the top of the file.

natemcmaster: Yeah, my bad. Pre-mature sign off.

JunTaoLuo:
Tested this and it works building locally. I'll make the same changes in all other repos once this is approved

PR #362, opened 2017-08-09
Implement version pinning from PackageLineup's

natemcmaster:
PackageLineup can be specified in repo.props/targets as a way to pin a set of package references. This adds support to KoreBuild for using a nuget package as a lineup, and to pin the versions.

Details: https://github.com/aspnet/BuildTools/wiki/%5Bspec%5D-Package-version-lineups

Part of https://github.com/aspnet/KoreBuild/issues/239.

**Example usage:**
To use this feature, our repos can make the following changes:
1. Remove the `Version` attribute from PackageReference
```xml
<ItemGroup>
  <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstraction" />
</ItemGroup>
```
2. Add a package lineup to the repo.
```xml
<!-- build/repo.props -->
<ItemGroup>
  <PackageLineup Include="Internal.AspNetCore.Coherence.Lineup" Version="2.1.0-*" />
</ItemGroup>
```
3. To make the VS experience better, add the following target to the project.
```xml
<Project InitialTargets="EnsureKoreBuildRestored">
  <Target Name="EnsureKoreBuildRestored" Condition=" '$(KoreBuildRestoreTargetsImported)' != 'true'">
    <Error Text="Package references have not been pinned. Run 'build.cmd /t:Restore'." />
  </Target>
</Project>
```

cc @mikeharder @davidfowl

### DIFF
--- a/modules/NuGet.Tasks/ApplyNuGetPolicies.cs
+++ b/modules/NuGet.Tasks/ApplyNuGetPolicies.cs
@@ -1,0 +1,249 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Framework;
+using Newtonsoft.Json;
+using NuGet.Tasks.Policies;
+using NuGet.Tasks.ProjectModel;
+using NuGet.Tasks.Utilties;
+
+namespace NuGet.Tasks
+{
+    /// <summary>
+    /// Applies policies to a repository on how NuGet can be used, such as restricting which PackageReference's are allowed,
+    /// which versions of packages can be used, which feeds are available, etc.
+    /// </summary>
+    public class ApplyNuGetPolicies : Microsoft.Build.Utilities.Task, ICancelableTask
+    {
+        private readonly CancellationTokenSource _cts;
+        private readonly MSBuildPolicyFactory _policyFactory;
+
+        public ApplyNuGetPolicies()
+        {
+            _policyFactory = new MSBuildPolicyFactory();
+            _cts = new CancellationTokenSource();
+        }
+
+        [Required]
+        public string SolutionDirectory { get; set; }
+
+        /// <summary>
+        /// The policies to be applied to the repository.
+        /// </summary>
+        [Required]
+        public ITaskItem[] Policies { get; set; }
+
+        /// <summary>
+        /// The MSBuild projects or solutions to which the policy should be applied.
+        /// </summary>
+        [Required]
+        public ITaskItem[] Projects { get; set; }
+
+        /// <summary>
+        /// Key-value base list of properties to be applied to <see cref="Projects" /> during project evaluation.
+        /// e.g. "Configuration=Debug;BuildNumber=1234"
+        /// </summary>
+        public string ProjectProperties { get; set; }
+
+        /// <summary>
+        /// NuGet sources, ; delimited. When set, they override sources from NuGet.config.
+        /// </summary>
+        public string RestoreSources { get; set; }
+
+        /// <summary>
+        /// NuGet sources, ; delimited.
+        /// When set they are in addition to source from NuGet.config and/or <see cref="RestoreSources"/>.
+        /// </summary>
+        public string RestoreAdditionalSources { get; set; }
+
+        /// <summary>
+        /// User packages folder
+        /// </summary>
+        public string RestorePackagesPath { get; set; }
+
+        /// <summary>
+        /// Disable parallel project restores and downloads
+        /// </summary>
+        public bool RestoreDisableParallel { get; set; }
+
+        /// <summary>
+        /// NuGet.Config path
+        /// </summary>
+        public string RestoreConfigFile { get; set; }
+
+        /// <summary>
+        /// Disable the web cache
+        /// </summary>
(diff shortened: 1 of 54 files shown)


davidfowl:
@natemcmaster How does the tooling deal with this?

natemcmaster:
If you open up the solution before pinning, you'll see this:
image
image

Once you execute `build.cmd /t:Restore`, package versions and are pinned. In 15.3, VS should auto-detect the update without needing to close/re-open.
image

Because this sets "IsImplicitlyDefined", upgrades do not appear. The NuGet UI will show the current version, but upgrades are locked.

image

modules/NuGet.Tasks/ApplyNuGetPolicies.cs
```diff
@@ -0,0 +1,248 @@
...
+        /// <summary>
+        /// The MSBuild projects or solutions to which the policy should be applied.
+        /// </summary>
+        [Required]
+        public ITaskItem[] Projects { get; set; }
+
+        /// <summary>
+        /// e.g. "Configuration=Debug;BuildNumber=1234"
+        /// </summary>
+        public string Properties { get; set; }
```
ryanbrandenburg: I assume that these properties could be different depending on the policies we try to enforce?

modules/NuGet.Tasks/ApplyNuGetPolicies.cs
```diff
@@ -0,0 +1,248 @@
...
+using NuGet.Tasks.ProjectModel;
+using NuGet.Tasks.Utilties;
+
+namespace NuGet.Tasks
+{
+    /// <summary>
+    /// Applies policies to a repository on how NuGet can be used, such as restricting which PackageReference's are allowed,
+    /// which versions of packages can be used, which feeds are available, etc.
+    /// </summary>
+    public class ApplyNuGetPolicies : Microsoft.Build.Utilities.Task, ICancelableTask
```
ryanbrandenburg: Why do we want this to be an `ICancelableTask`?

modules/NuGet.Tasks/Internal/MSBuildLogger.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using NuGet.Common;
+
+namespace NuGet.Build
+{
+    /// <summary>
+    /// TaskLoggingHelper -> ILogger
+    /// </summary>
+    internal class MSBuildLogger : LoggerBase, Common.ILogger
```
ryanbrandenburg: Do we already have this somewhere else or do we imagine this will be the only place that needs it?

modules/NuGet.Tasks/NuGet.Tasks.csproj
```diff
@@ -6,15 +6,16 @@
...
+    <PackageReference Include="Microsoft.Build" Version="$(MicrosoftBuildVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildVersion)" PrivateAssets="All" />
-    <PackageReference Include="NuGet.Packaging" Version="$(NuGetInMSBuildVersion)" PrivateAssets="All" />
-    <PackageReference Include="NuGet.Protocol" Version="$(NuGetInMSBuildVersion)" PrivateAssets="All" />
+    <PackageReference Include="NuGet.Build.Tasks" Version="$(NuGetInMSBuildVersion)" PrivateAssets="All" />
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetInMSBuildVersion)" PrivateAssets="All" />
   </ItemGroup>
 
-</Project>
+</Project>
```
ryanbrandenburg: Newline.

modules/NuGet.Tasks/module.targets
```diff
@@ -0,0 +1,41 @@
...
+    <RestoreDependsOn>ApplyNuGetPolicies;$(RestoreDependsOn)</RestoreDependsOn>
+    <ApplyNuGetPoliciesDependsOn>
+      $(ApplyNuGetPoliciesDependsOn);
+      ResolveSolutions;
+      $(PrepareDependsOn);
+    </ApplyNuGetPoliciesDependsOn>
+  </PropertyGroup>
+
+  <ItemGroup Condition="@(PackageLineup->Count()) != 0 AND '$(RestrictVersionOnPackageReference)' != 'false' ">
+    <!-- Automatically warn about package reference with versions this when lineups are used. -->
```
ryanbrandenburg: "with versions this when..."? Not sure what you're getting at here.

natemcmaster: grammar error. it said something different at first and i failed to delete "this"

modules/NuGet.Tasks/Internal/MSBuildLogger.cs
natemcmaster: This implements NuGet's `ILogger` to forward to MSBuild. At the moment, I don't think we use any other NuGet APIs that require ILogger.

modules/NuGet.Tasks/ApplyNuGetPolicies.cs
natemcmaster: 'Properties' is a little vague. I'm going to rename to 'ProjectProperties'. This are properties meant to be applied to projects and solutions when they are loaded. This task executes a design-time build in order to examine target frameworks, package references, and a few other details.

modules/NuGet.Tasks/ApplyNuGetPolicies.cs
natemcmaster: So Ctrl+C doesn't orphan a potentially long-running network request.

PR #363, opened 2017-08-09
[WIP]Modularize KoreBuild commands

ryanbrandenburg:
Modularize commands in KoreBuild and add a docker command.

Right now I'm thinking you would call this like `./run.ps1 --Command docker-build --platform Ubuntu /t:Publish` or `./run.ps1` (which would be the same as calling `./build.cmd` now)


files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -15,200 +15,46 @@ function Join-Paths($path, $childPaths) {
 Set-Variable 'IS_WINDOWS' -Scope Script -Option Constant -Value $((Get-Variable -Name IsWindows -ValueOnly -ErrorAction Ignore) -or !(Get-Variable -Name IsCoreClr -ValueOnly -ErrorAction Ignore))
 Set-Variable 'EXE_EXT' -Scope Script -Option Constant -Value $(if ($IS_WINDOWS) { '.exe' } else { '' })
 
-<#
```
ryanbrandenburg: I'll need to add commends to all of these functions once they're more set in stone.

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -15,200 +15,46 @@ function Join-Paths($path, $childPaths) {
...
+function Set-Settings
+{
+    [CmdletBinding()]
+    param (
+        [string]$ToolsSource,
+        [string]$DotNetHome,
+        [string]$Path,
+        [string]$ConfigFile 
+    )
+    $global:ToolsSource = $ToolsSource
```
ryanbrandenburg: @natemcmaster is this what you meant about Set-Settings?

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -15,200 +15,46 @@ function Join-Paths($path, $childPaths) {
...
-        $msBuildResponseFile = Join-Path $msbuildArtifactsDir msbuild.rsp
-
-        $msBuildLogArgument = ""
-
-        if ($VerbosePreference -eq 'Continue' -or $env:KOREBUILD_ENABLE_BINARY_LOG -eq "1") {
-            Write-Verbose 'Enabling binary logging'
-            $msbuildLogFilePath = Join-Path $msbuildArtifactsDir msbuild.binlog
-            $msBuildLogArgument = "/bl:$msbuildLogFilePath"
+        docker-build {
+            Invoke-DockerBuild --Args $Args
```
ryanbrandenburg: I should really be calling `Invoke-DockerBuild --platform $platform --Args $ArgsMinusPlatform` but I'm not sure what a good way to split off defined parameters from passthrough params is. Thoughts/advice?

files/KoreBuild/scripts/commands/Commands.psd1
```diff
@@ -0,0 +1,11 @@
...
+    GUID = '0764e878-8885-4923-8d04-2d3d902712a5'
+    RootModule = 'Commands.psd1'
+    Author = 'ASP.NET Core'
+    CompanyName = 'Microsoft'
+    ModuleVersion = '0.1'
+    PowerShellVersion = '4.0'
+    FunctionsToExport = @('Invoke-DockerBuild','Install-Tools','Invoke-MSBuild')
+    AliasesToExport = @('')
+    VariablesToExport = @('')
+}
```
ryanbrandenburg: Newline.

files/KoreBuild/scripts/commands/Commands.psm1
```diff
@@ -0,0 +1,8 @@
+#requires -version 4
+
+Set-StrictMode -Version 2
+$ErrorActionPreference = 'Stop'
+
+. "$PSScriptRoot\Invoke-DockerBuild.ps1"
```
ryanbrandenburg: Is there some way that I'll only have to include either these `. "$script"` imports or the `FunctionsToExport` stuff in the .psd1?

files/KoreBuild/scripts/commands/Install-Tools.ps1
```diff
@@ -0,0 +1,106 @@
...
+.PARAMETER DotNetHome
+The directory where tools will be stored on the local machine.
+#>
+function Install-Tools(
+    [Parameter(Mandatory = $true)]
+    [string]$ToolsSource,
+    [Parameter(Mandatory = $true)]
+    [string]$DotNetHome) {
+
+    $ErrorActionPreference = 'Stop'
```
ryanbrandenburg: Hopefully I understood the intent and functions well enough to split this off properly here.

scripts/bootstrapper/build.ps1
```diff
@@ -1,177 +1 @@
...
-Import-Module -Force -Scope Local (Join-Path $korebuildPath 'KoreBuild.psd1')
-
-try {
-    Install-Tools $ToolsSource $DotNetHome
-    Invoke-RepositoryBuild $Path @MSBuildArgs
-}
-finally {
-    Remove-Module 'KoreBuild' -ErrorAction Ignore
-}
+& .\run.ps1
```
ryanbrandenburg: Newline

scripts/bootstrapper/run.cmd
```diff
@@ -0,0 +1,2 @@
+@ECHO OFF
+PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0run.ps1' %*; exit $LASTEXITCODE"
```
ryanbrandenburg: Newline

scripts/bootstrapper/run.ps1
```diff
@@ -0,0 +1,166 @@
...
+        finally {
+            Remove-Item $tmpfile -ErrorAction Ignore
+        }
+    }
+
+    return $korebuildPath
+}
+
+function Get-LocalKoreBuild {
+    return "$PSScriptRoot/../../files/KoreBuild"
```
ryanbrandenburg: I added this temporarily to make it easier to test changes in KoreBuild. I'd like to leave it in so long as the logic is correct.

scripts/bootstrapper/run.ps1
```diff
@@ -0,0 +1,166 @@
...
+        elseif ($env:USERPROFILE) { Join-Path $env:USERPROFILE '.dotnet'} `
+        elseif ($env:HOME) {Join-Path $env:HOME '.dotnet'}`
+        else { Join-Path $PSScriptRoot '.dotnet'}
+}
+
+if (!$Channel) { $Channel = 'dev' }
+if (!$ToolsSource) { $ToolsSource = 'https://aspnetcore.blob.core.windows.net/buildtools' }
+
+# Execute
+$korebuildPath = Get-LocalKoreBuild
```
ryanbrandenburg: I'll have to revert this to `Get-KoreBuild`

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -15,200 +15,46 @@ function Join-Paths($path, $childPaths) {
...
-        Write-Verbose "Building $Path"
-        Write-Verbose "dotnet = ${global:dotnet}"
-
-        # Generate global.json to ensure the repo uses the right SDK version
-        $sdkVersion = __get_dotnet_sdk_version
-        if ($sdkVersion -ne 'latest') {
-            "{ `"sdk`": { `"version`": `"$sdkVersion`" } }" | Out-File (Join-Path $Path 'global.json') -Encoding ascii
-        } else {
-            Write-Verbose "Skipping global.json generation because the `$sdkVersion = $sdkVersion"
+function Invoke-Command(
```
natemcmaster: You'll need to choose a different name. This one is taken :) 

https://docs.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Core/Invoke-Command?view=powershell-6

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -15,200 +15,46 @@ function Join-Paths($path, $childPaths) {
...
-    [string] $Path,
-    [Parameter(ValueFromRemainingArguments = $true)]
-    [string[]] $MSBuildArgs) {
-
-    $ErrorActionPreference = 'Stop'
-
-    if (-not $PSBoundParameters.ContainsKey('Verbose')) {
-        $VerbosePreference = $PSCmdlet.GetVariableValue('VerbosePreference')
-    }
+function Set-Settings
```
natemcmaster: `Set-KoreBuildSettings` would be better to avoid possible conflicts with other modules that may use this name.

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -15,200 +15,46 @@ function Join-Paths($path, $childPaths) {
...
+function Invoke-Command(
+    [Parameter(Mandatory=$true)]
+    [string]$Command,
+    [string[]]$Args
+)
+{
+    # call the command
+    
+    $commandModule = "$PSScriptRoot\commands\Commands.psm1"
+    Import-Module -Force -Scope Local $commandModule
```
natemcmaster: You don't really need the `Commands.psm1` file at all. Normally you only write a .psd1 file when you want to control what is "internal" vs global. You could simply add this to KoreBuild.psm1:
```ps1
Get-ChildItems "$PSScriptRoot/commands/*.psm1" | % {
   Import-Module $_
}
```

files/KoreBuild/scripts/commands/Invoke-DockerBuild.ps1
```diff
@@ -0,0 +1,31 @@
...
+
+    $dockerFileName = "$platform.dockerfile"
+    $dockerFile = "$PSScriptRoot\docker\$dockerFileName"
+    $dfDestination = "$global:Path\$dockerFileName"
+
+    # dockerfile must be inside the docker context
+    Copy-Item -Path $dockerFile -Destination $dfDestination -Force
+
+    Write-Host "Building '$dfDestination' as '$containerName'"
+    & docker build --build-arg BUILD_ARGS=$Args -t $containerName -f $dfDestination $global:Path
```
natemcmaster: Nit: I'd recommend use the `__exec` function so the console will log what is being executed.

files/KoreBuild/scripts/commands/Invoke-MSBuild.ps1
```diff
@@ -0,0 +1,137 @@
...
+        Remove-Item $publishFolder -Recurse -Force
+    }
+
+    $sdkPath = "/p:RepoTasksSdkPath=$(Join-Paths $PSScriptRoot ('..', 'msbuild', 'KoreBuild.RepoTasks.Sdk', 'Sdk'))"
+
+    __exec $global:dotnet restore $taskProj $sdkPath
+    __exec $global:dotnet publish $taskProj --configuration Release --output $publishFolder /nologo $sdkPath
+}
+
+function __exec($cmd) {
```
natemcmaster: We will likely need this in other places. It might be good to create a `common.psm1` file that individual commands can import.
```
- scripts/
    + common.psm1
    + KoreBuild.psm1
- commands/
     + Example.psm1
```
```ps1
# commands/Example.psm1

Import-Module "$PSScriptRoot/../common.psm1"
```

scripts/bootstrapper/run.ps1
natemcmaster: This is how i usually test the bootstrapper:
```ps1
cd C:\dev\aspnet\BuildTools
./build.ps1 /t:Package
./scripts/bootstrapper/build.ps1 -ToolsSource ./artifacts/ -Path C:\dev\aspnet\Razor
```

scripts/bootstrapper/build.ps1
natemcmaster: We can delete build.ps1 altogether. Our automation (like TC) invoke build.cmd, not build.ps1. We should, however, change `build.cmd` to invoke `run.ps1 build`

files/KoreBuild/scripts/commands/docker/Ubuntu.dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893 && \
+    apt-get update && \
+    apt-get install -y dotnet-dev-1.0.4
+
+RUN npm install -g bower && \
+    npm install -g grunt && \
+    npm install -g gulp && \
+    npm install -g typescript
+
+ADD ./ ./
```
ryanbrandenburg: @natemcmaster do you know if this will just happen automatically?

files/KoreBuild/scripts/commands/docker/Ubuntu.dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:16.04
+
+ARG BUILD_ARGS
+
+RUN apt-get update && \
```
natemcmaster: Does this need to be Ubuntu 16? Or can we use existing `microsoft/dotnet` images that already install most of this?

e.g.: https://github.com/dotnet/dotnet-docker/blob/master/2.0/runtime-deps/stretch/Dockerfile

files/KoreBuild/scripts/commands/docker/Ubuntu.dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    apt-get install -y unzip && \
+    apt-get install -y sudo && \
+    apt-get install -y apt-transport-https
+
+RUN sudo sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ xenial main" > /etc/apt/sources.list.d/dotnetdev.list' && \
+    sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 417A0893 && \
+    apt-get update && \
+    apt-get install -y dotnet-dev-1.0.4
+
+RUN npm install -g bower && \
```
natemcmaster: Which repos need NPM?

files/KoreBuild/scripts/KoreBuild.psm1
natemcmaster: You can try to use splatting: https://docs.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Core/about_Splatting.

Or, you may find it will be easier to implement `Invoke-KoreBuildCommand` like this

```ps1
function Invoke-KoreBuildCommand {
   param( 
    [string]$CommandName,
    [Parameter(ValueFromRemainingArguments = $true)]
    [string[]] $MoreArgs
   )

   $commandFile = Join-Path $PSScriptRoot ('..', 'commands', "$CommandName.ps1")
   if (!(Test-Path $commandFile)) {
       Write-Error "Unrecognized command: $CommandName"
   }
   & $commandFile -Config $global:KoreBuildSettings @MoreArgs
}

File: scripts/bootstrapper/run.ps1 (hunk 10 above)
Comment (ryanbrandenburg) [inline scripts/bootstrapper/run.ps1]: I'll add that to the README.md.

File: files/KoreBuild/scripts/KoreBuild.psm1 (hunk 3 above)
Comment (natemcmaster) [inline files/KoreBuild/scripts/KoreBuild.psm1]: Ok, splatting isn't working as I expected. We might need something more complex. This looks like a promising answer: https://stackoverflow.com/a/30795074/2526265

Comment (natemcmaster) [inline files/KoreBuild/scripts/KoreBuild.psm1]: I played around with this a little last night and it seems splatting works better if you using the automatic variable `@args`.

**Comment by ryanbrandenburg** [comment]
:up::date:

**Comment by natemcmaster** [comment]
Currently fails on Linux/macOS
```
./build.sh: line 5: /home/travis/build/aspnet/BuildTools/files/KoreBuild/KoreBuild.sh: No such file or directory
```

File: scripts/UploadBlobs.ps1 [hunk 19]
```diff
@@ -95,3 +78,5 @@ $globs | ForEach-Object {
         }
     }
 }
+
+Import-Module "$PSScriptRoot/common.psm1"
```
natemcmaster: Put this at the top of the file and use `Join-Path` where possible.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+
+namespace KoreBuild.Console.Commands
+{
+    internal class DockerBuildCommand : SubCommandBase
+    {
+        private CommandArgument Platform { get; set; }
```
natemcmaster: Either make this public, or make it a private field instead of private property.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+            var dockerFileSource = Path.Combine(Directory.GetCurrentDirectory(), "Commands", "DockerFiles" , dockerFileName);
+
+            File.Copy(dockerFileSource, dockerFileDestination, true);
+
+            var buildArgs = new List<string> { "build" };
+
+            buildArgs.AddRange(new string[] { "-t", ContainerName, "-f", dockerFileDestination, RepoPath });
+            var buildResult = RunDockerCommand(buildArgs);
+
+            if(buildResult != 0)
```
natemcmaster: Nit: spacing

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                return buildResult;
+            }
+
+            var runArgs = new List<string> { "run", "--rm", "-it", "--name", ContainerName, ContainerName };
+
+            if (Arguments != null && Arguments.Count > 0)
+            {
+                var argString = String.Join(" ", Arguments);
```
natemcmaster: Nit: lowercase primitive types. `string`

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+            var buildResult = RunDockerCommand(buildArgs);
+
+            if(buildResult != 0)
+            {
+                return buildResult;
+            }
+
+            var runArgs = new List<string> { "run", "--rm", "-it", "--name", ContainerName, ContainerName };
+
+            if (Arguments != null && Arguments.Count > 0)
```
natemcmaster: Nit: you can use the elvis operator. `Arguments?.Count > 0`

tools/KoreBuild.Console/Commands/DockerFiles/ubuntu.dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    apt-get install -y dotnet-dev-1.0.4
+
+RUN npm install -g bower && \
+    npm install -g grunt && \
+    npm install -g gulp && \
+    npm install -g typescript
+
+ADD ./ ./
+
+RUN dos2unix ./build.sh
```
natemcmaster: Who is use CRLF in build.sh? Ideally, this shouldn't be necessary because we've explicitly set `eol=lf` in .gitattributes

tools/KoreBuild.Console/Commands/InstallToolsCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+
+                var process = Process.Start(psi);
+                process.WaitForExit();
+            }
+
+            var channel = GetChannel();
+            var runtimeChannel = GetRuntimeChannel();
+            var runtimeVersion = GetRuntimeVersion();
+
+            var runtimesToInstall = new List<Tuple<string, string>>
```
natemcmaster: Honor `KOREBUILD_SKIP_OLD_RUNTIME_INSTALL`.

build.ps1
```diff
@@ -13,16 +13,12 @@ param(
...
-    Import-Module -Force -Scope Local $PSScriptRoot/files/KoreBuild/KoreBuild.psd1
-    Install-Tools $ToolsSource $DotNetHome
-    Invoke-RepositoryBuild $Path @MSBuildArgs
-}
-finally {
-    Remove-Module 'KoreBuild' -ErrorAction Ignore
-}
+$korebuildConsoleproj = "tools/KoreBuild.Console/KoreBuild.Console.csproj"
+
+& dotnet run -p $korebuildConsoleproj install-tools --toolsSource $ToolsSource --dotNetHome $DotNetHome $Arguments
```
natemcmaster: Regrettably, this won't work. It requires a pre-existing installation of dotnet, which we are not guaranteed to have. Something needs to execute via PowerShell/bash to install dotnet before you can run KoreBuild.Console.

files/KoreBuild/README.md
```diff
@@ -9,8 +9,9 @@ Layout
...
 ----------------------------|--------------
 .version                    | Contains the current version of korebuild. For diagnostics and logging.
-KoreBuild.psd1              | PowerShell module for executing korebuild.
-KoreBuild.sh                | Bash module for executing korebuild. (Should be `source`'d).
 config/                     | Contains configuration data.
 scripts/                    | Bash/PowerShell scripts
 modules/                    | Extensions to the KoreBuild lifecycle.
+
+
+./build.ps1 docker-build -platform ubuntu "/t:package"
```
natemcmaster: Can you add more description here to what is happening?

scripts/bootstrapper/run.ps1
```diff
@@ -164,14 +169,8 @@ if (!$Channel) { $Channel = 'dev' }
...
-try {
-    Install-Tools $ToolsSource $DotNetHome
-    Invoke-RepositoryBuild $Path @MSBuildArgs
-}
-finally {
-    Remove-Module 'KoreBuild' -ErrorAction Ignore
-}
+$koreBuildproj = Join-Paths $korebuildPath ("tools", "KoreBuild.Console.dll")
+
+& dotnet $koreBuildproj $Command --toolsSource=$ToolsSource --dotNetHome=$DotNetHome --repoPath=$Path $Arguments
```
natemcmaster: As above, this isn't going to work. We don't have a gurantee that dotnet is available, or that the right shared runtime is installed. This needs to happen first.

scripts/bootstrapper/run.ps1
```diff
@@ -164,14 +169,8 @@ if (!$Channel) { $Channel = 'dev' }
...
-Import-Module -Force -Scope Local (Join-Path $korebuildPath 'KoreBuild.psd1')
 
-try {
-    Install-Tools $ToolsSource $DotNetHome
-    Invoke-RepositoryBuild $Path @MSBuildArgs
-}
-finally {
-    Remove-Module 'KoreBuild' -ErrorAction Ignore
-}
+$koreBuildproj = Join-Paths $korebuildPath ("tools", "KoreBuild.Console.dll")
```
natemcmaster: IMO this is an internal implementation detail. I would prefer we keep `KoreBuild.psd1` and add a function called `Invoke-KoreBuildCommand` which can then invoke KoreBuild.Console.dll. It gives us the flexibility to implement commands in whatever we want (python, shell, C#) without needing to change the bootstrapper.

tools/KoreBuild.Console/Commands/MSBuildCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+
+            var msBuildArguments = string.Empty;
+
+            foreach (var arg in Arguments)
+            {
+                msBuildArguments += Environment.NewLine + arg;
+            }
+
+            msBuildArguments += $@"
```
natemcmaster: This is may be problematic for Linux...we should test. Literal multiline strings in C# will capture `\r\n` in its value. `dotnet msbuild @file.rsp` may be expecting only `\n` on Linux.

To test this, build KoreBuild on Windows and run the artifact on Linux.

tools/KoreBuild.Console/Commands/MSBuildCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+""""
+";
+
+            if(!Directory.Exists(msBuildArtifactsDir))
+            {
+                Directory.CreateDirectory(msBuildArtifactsDir);
+            }
+
+            File.WriteAllText(msBuildResponseFile, msBuildArguments, System.Text.Encoding.ASCII);
+            var noop = msBuildArguments.Contains("/t:Noop") || msBuildArguments.Contains("/t:Cow");
```
natemcmaster: Use case-insensitive comparisons.

tools/KoreBuild.Console/Commands/PushToNugetCommand.cs
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.CommandLineUtils;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace KoreBuild.Console.Commands
+{
+    internal class PushToNugetCommand : SubCommandBase
```
natemcmaster: We probably don't need this command anymore. It's regrettable you've ported this from PS to C# because I was planning to remove it altogether. We don't use it anywhere.

tools/KoreBuild.Console/Commands/RootCommand.cs
```diff
@@ -0,0 +1,29 @@
...
+            application.Command("install-tools", new InstallToolsCommand().Configure, throwOnUnexpectedArg:false);
+            application.Command("msbuild", new MSBuildCommand().Configure, throwOnUnexpectedArg:false);
+            application.Command("push-nuget", new PushToNugetCommand().Configure);
+
+            application.VersionOption("--version", GetVersion);
+            base.Configure(application);
+        }
+
+        private static string GetVersion()
+                => typeof(RootCommand).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()
```
natemcmaster: Nit: formatting

tools/KoreBuild.Console/Commands/SubCommandBase.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace KoreBuild.Console.Commands
+{
+    internal class SubCommandBase : CommandBase
+    {
+        private string _defaultToolsSource = "https://aspnetcore.blob.core.windows.net/buildtools";
```
natemcmaster: `const`

tools/KoreBuild.Console/Commands/SubCommandBase.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+
+        private string GetDotNetHome()
+        {
+            var dotnetHome = Environment.GetEnvironmentVariable("DOTNET_HOME");
+            var userProfile = Environment.GetEnvironmentVariable("USERPROFILE");
+            var home = Environment.GetEnvironmentVariable("HOME");
+
+            var dotnetFolderName = ".dotnet";
```
natemcmaster: `const`

tools/KoreBuild.Console/Commands/SubCommandBase.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            return result;
+        }
+
+        private string FindKoreBuildDirectory()
+        {
+            var executingDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var root = Directory.GetDirectoryRoot(executingDir);
+            while (true)
```
natemcmaster: It's usually better to put some kind of condition here. `while(true)` is a dangerous pattern.

files/KoreBuild/scripts/common.sh
```diff
@@ -1,116 +0,0 @@
...
-    set -e
-    if [ $exit_code -ne 0 ]; then
-        __error "<<< $cmdname failed with exit code $exit_code"
-    elif [ "$__is_verbose" = true ]; then
-        echo -e "${GREEN}<<< $cmdname [$exit_code]${RESET}"
-    fi
-    return $exit_code
-}
-
-__ensure_osx_version() {
```
natemcmaster: We need this. Keep this check.

files/KoreBuild/scripts/get-netfx.sh
```diff
@@ -1,46 +0,0 @@
-#!/usr/bin/env bash
```
natemcmaster: This is required for Linux build to work.

tools/KoreBuild.Console/Commands/InstallToolsCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+            var dotnet = GetDotNetExecutable();
+            var dotnetOnPath = GetCommandFromPath("dotnet");
+
+            if (dotnetOnPath != null && (dotnetOnPath != dotnet))
+            {
+                Log($"dotnet found on the system PATH is '{dotnetOnPath}' but KoreBuild will use '{dotnet}'");
+            }
+
+            var pathPrefix = Directory.GetParent(dotnet);
+            if (PathENV.StartsWith($"{pathPrefix};"))
```
natemcmaster: `StringComparison.OrdinalIgnoreCase`

Also, you'll need to make this x-plat friendly by also using `System.IO.Path.PathSeparator` instead of hard-coding the semicolon.

tools/KoreBuild.Console/Commands/InstallToolsCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                Log($"installDir = {installDir}");
+                Log($"DOTNET_INSTALL_DIR = {DotNetInstallDir}");
+                Log("The environment variable DOTNET_INSTALL_DIR is deprecated. The recommended alternative is DOTNET_HOME.");
+            }
+
+            var dotnet = GetDotNetExecutable();
+            var dotnetOnPath = GetCommandFromPath("dotnet");
+
+            if (dotnetOnPath != null && (dotnetOnPath != dotnet))
```
natemcmaster: This may not work. String comparison in PowerShell is case-insensitive. Also, in Linux this comparison should account for symbolic links.

files/KoreBuild/scripts/get-dotnet.sh
```diff
@@ -1,101 +0,0 @@
...
-install_dir=$1
-
-if [ ! -z "${DOTNET_INSTALL_DIR:-}" ] && [ "${DOTNET_INSTALL_DIR:-}" != "$install_dir" ]; then
-    __verbose "install_dir = $install_dir"
-    __verbose "DOTNET_INSTALL_DIR = $DOTNET_INSTALL_DIR"
-    __warn 'The environment variable DOTNET_INSTALL_DIR is deprecated. The recommended alternative is DOTNET_HOME.'
-fi
-
-dotnet_in_path="$(which dotnet 2>/dev/null || true )"
-# The '-ef' condition tests if files are the same inode. This avoids showing the warning if users symlink dotnet into path
```
natemcmaster: This will be hard to do right in C#. It's only mildly important we do it, but it's one of the pieces of functionality we've lost in this port.

tools/KoreBuild.Console/Commands/InstallToolsCommand.cs
```diff
@@ -0,0 +1,203 @@
...
+        {
+            var installDir = GetDotNetInstallDir();
+
+            Log($"Installing tools to '{installDir}'");
+            
+            if(DotNetInstallDir != null && DotNetInstallDir != installDir)
+            {
+                Log($"installDir = {installDir}");
+                Log($"DOTNET_INSTALL_DIR = {DotNetInstallDir}");
+                Log("The environment variable DOTNET_INSTALL_DIR is deprecated. The recommended alternative is DOTNET_HOME.");
```
natemcmaster: Preserve existing logging levels. This means we need warn (yellow), verbose (darkgray), error (red), and normal output.

tools/KoreBuild.Console/Commands/SubCommandBase.cs
```diff
@@ -0,0 +1,175 @@
...
+            _toolsSourceOption = application.Option("--toolsSource", "The source to draw tools from.", CommandOptionType.SingleValue);
+            _repoPathOption = application.Option("--repoPath", "The path to the repo to work on.", CommandOptionType.SingleValue);
+            _dotNetHomeOption = application.Option("--dotNetHome", "The place where dotnet lives", CommandOptionType.SingleValue);
+        }
+
+        protected override bool IsValid()
+        {
+            if(!Directory.Exists(RepoPath))
+            {
+                System.Console.WriteLine($"The RepoPath '{RepoPath}' doesn't exist.");
```
natemcmaster: Use logging.

tools/KoreBuild.Console/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            new RootCommand().Configure(application);
+
+            try
+            {
+                return application.Execute(args);
+            }
+            catch (Exception ex)
+            {
+                System.Console.WriteLine($"Exception thrown: '{ex.ToString()}'");
```
natemcmaster: `Console.Error.WriteLine`

tools/KoreBuild.Console/Commands/CommandBase.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        protected virtual bool IsValid()
+        {
+            return true;
+        }
+
+        protected void Log(string message)
+        {
+            System.Console.WriteLine(message);
```
natemcmaster: Take a look at this: https://github.com/aspnet/DotNetTools/blob/dev/shared/ConsoleReporter.cs

files/KoreBuild/scripts/common.sh
ryanbrandenburg: How is that check called?

natemcmaster: Oh snap, it isn't. Looks like this got removed in our refactoring.

tools/KoreBuild.Console/Commands/DockerFiles/ubuntu.dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:16.04
```
natemcmaster: By the way, consider using this as your base: https://github.com/dotnet/dotnet-docker/blob/master/2.0/runtime-deps/stretch/amd64/Dockerfile

```
FROM microsoft/dotnet:2.0-runtime-deps
```

natemcmaster: It'll save you from having to remember all the required dependencies of dotnet.

ryanbrandenburg:
:up::date: fixed most comments and made note of the issues we're not fixing quite yet, added a script to let us quickly test KoreBuild commands.

scripts/bootstrapper/run.ps1
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env powershell
+#requires -version 4
+
+<#
+.SYNOPSIS
+Build this repository
```
natemcmaster: "Executes KoreBuild commands"

scripts/bootstrapper/run.ps1
```diff
@@ -0,0 +1,181 @@
...
+
+.DESCRIPTION
+Downloads korebuild if required. Then builds the repository.
```
natemcmaster: "Downloads KoreBuild if required. Then executes the KoreBuild command. To see available commands, execute with \`-Command help\`"

scripts/bootstrapper/run.ps1
```diff
@@ -0,0 +1,181 @@
...
+    <KoreBuildToolsSource>https://aspnetcore.blob.core.windows.net/buildtools</KoreBuildToolsSource>
+  </PropertyGroup>
+</Project>
+```
+#>
+[CmdletBinding(PositionalBinding = $false)]
+param(
+    [Parameter(Mandatory=$true, Position = 0)]
+    [string]$Command,
+    [Alias('p')]
```
natemcmaster: Can you remove this alias? It's intentionally missing from `build.ps1` b/c I got feedback that `-p` was too ambiguous.

scripts/bootstrapper/run.ps1
```diff
@@ -0,0 +1,181 @@
...
+if (!$Channel) { $Channel = 'dev' }
+if (!$ToolsSource) { $ToolsSource = 'https://aspnetcore.blob.core.windows.net/buildtools' }
+
+# Execute
+$korebuildPath = Get-KoreBuild
+
+Import-Module -Force -Scope Local (Join-Path $korebuildPath 'KoreBuild.psd1')
+
+try {
+    Invoke-KoreBuildCommand $Command $ToolsSource $DotNetHome $Path @Arguments
```
natemcmaster: At one point, we talked about having a `Set-KoreBuildSettings` command. I still think that would be a useful way to break apart the functions.
```ps1
Set-KoreBuildSettings -ToolsSource $ToolsSource -DotNetHome $DotNetHome -RepoPath $Path
Invoke-KoreBuildCommand $Command @Arguments
```

shared/Microsoft.Extensions.CommandLineUtils.Sources/Output/ConsoleReporter.cs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.Extensions.Tools.Internal
```
natemcmaster: Nit: can you move these copied files into tools/KoreBuild.Console/ ? We don't need them in shared/, and also they're not part of Microsoft.Extensions.CommandLineUtils.Sources

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.CommandLineUtils;
+using System.Collections.Generic;
```
natemcmaster: Nit: sort System.* usings first

ryanbrandenburg:
:up::date: I also did the broad strokes on the bash scripts, but I haven't actually tested them yet because I'm waiting on a VM to spin up.

natemcmaster:
> /home/travis/build/aspnet/BuildTools/files/KoreBuild/KoreBuild.sh: line 22: kore_build_settings: unbound variable

files/KoreBuild/KoreBuild.sh
```diff
@@ -6,6 +6,43 @@ source "$__korebuild_dir/scripts/common.sh"
...
+
+    [ -z "${dot_net_home:-}" ] && dot_net_home="$HOME/.dotnet"
+    [ -z "${tools_source:-}" ] && tools_source='https://aspnetcore.blob.core.windows.net/buildtools'
+
+    declare -A kore_build_settings
+    kore_build_settings["tools_source"]=$tools_source
+    kore_build_settings["dot_net_home"]=$dot_net_home
+    kore_build_settings["repo_path"]=$repo_path
+
+    export $kore_build_settings
```
natemcmaster: remove the dollar sign

tools/KoreBuild.Console/Commands/SubCommandBase.cs
```diff
@@ -0,0 +1,180 @@
...
+        public IReporter Reporter => new ConsoleReporter(PhysicalConsole.Singleton, _verbose != null, false);
+
+        public override void Configure(CommandLineApplication application)
+        {
+            _koreBuildDir = FindKoreBuildDirectory();
+
+            base.Configure(application);
+
+            _verbose = application.Option("-v|--verbose", "Show verbose output", CommandOptionType.NoValue);
+            _toolsSourceOption = application.Option("--toolsSource", "The source to draw tools from.", CommandOptionType.SingleValue);
```
natemcmaster: Nit: command-line typically use kebab-case, so `--tools-source` and `--dotnet-home`. See build.sh for example.

https://github.com/aspnet/BuildTools/blob/a8447f5472f2e0b96c85322edccc10d6e84a01f9/scripts/bootstrapper/build.sh#L31-L37

ryanbrandenburg:
:up::date: I fixed and clean up the bootstrappers in addition to responding to your comments.

tools/KoreBuild.Console/Commands/InstallToolsCommand.cs
```diff
@@ -0,0 +1,205 @@
...
+                var process = Process.Start(psi);
+                process.WaitForExit();
+            }
+
+            var channel = GetChannel();
+            var runtimeChannel = GetRuntimeChannel();
+            var runtimeVersion = GetRuntimeVersion();
+
+            // TODO: Honor KOREBUILD_SKIP_OLD_RUNTIME_INSTALL
+            var runtimesToInstall = new List<Tuple<string, string>>
```
natemcmaster: You can remove this now. We don't install these anymore.

tools/KoreBuild.Console/Commands/MSBuildCommand.cs
```diff
@@ -0,0 +1,122 @@
...
+            // TODO: naturalize newlines
+            msBuildArguments += $@"
+/nologo
+/m
+/p:RepositoryRoot={RepoPath}\
+""""
+{msBuildLogArgument}
+/clp:Summary
+""{makeFileProj}""
+""""
```
natemcmaster: Why are there two arguments that are just empty strings? (this and line 64)

tools/KoreBuild.Console/Commands/DockerFiles/debian.dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM microsoft/dotnet:2.0-runtime-deps-jessie
```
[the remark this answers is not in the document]
ryanbrandenburg: Stretch doesn't work yet due to [this](https://github.com/dotnet/cli/blob/master/scripts/obtain/dotnet-install.sh#L67), so it can't be our default, but I'll rename it for future-proofing.

natemcmaster:
Ok, good to know. Is there an issue tracking this, or should just fix it ourselves and make a PR to the CLI team?

Side note: we could also fix this by passing in a rid explicitly with `--runtime-id`

ryanbrandenburg:
There are no active issues that I can find, but it doesn't seem to be as simple as just adding it to that list because I didn't see anything indicating that dotnet/cli#6537 is fixed.

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -119,17 +185,15 @@ The base url where build tools can be downloaded.
 .PARAMETER DotNetHome
```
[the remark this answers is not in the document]
ryanbrandenburg: I've fixed this now, but it didn't update for some reason.

ryanbrandenburg:
:up::date:

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -111,25 +101,145 @@ function Invoke-RepositoryBuild(
...
+}
+
+function Get-DotnetMajorVersion() {
+    if(Get-Variable "dotnet" -Scope Global -ErrorAction SilentlyContinue)
+    {
+        $infoOutput = dotnet --version
+
+        $version = $infoOutput.SubString(0, $infoOutput.IndexOf('.'))
+        $versionInt = [convert]::ToInt32($version, 10)
+        Write-Host "Dotnet version was $versionInt"
```
natemcmaster: `write-verbose` or `write-debug`

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -22,21 +20,13 @@ Builds a repository
...
 
 .EXAMPLE
-Invoke-RepositoryBuild $PSScriptRoot /p:Configuration=Release /t:Verify
-
-.NOTES
-This is the main function used by most repos.
+Invoke-RepositoryBuild /p:Configuration=Release /t:Verify
 #>
 function Invoke-RepositoryBuild(
-    [Parameter(Mandatory = $true)]
```
natemcmaster: We can make life easier on ourselves if we keep this and `Install-Tools` backwards compatible. Make the parameters that were mandator optional, and fallback to `$global:KoreBuildSettings`. 

Otherwise, you have to update every single build.ps1/sh script out there are the same time to prevent breaking.

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -111,25 +101,145 @@ function Invoke-RepositoryBuild(
...
+function Get-KoreBuildConsole()
+{
+    return Join-Paths $PSScriptRoot ("..", "tools", "KoreBuild.Console.dll")
+}
+
+<#
+.SYNOPSIS
+Installs tools if required.
+#>
+function Install-Tools()
```
natemcmaster: Same here. Keep `$DotNetHome` as a parameter and fallback to `$global:KoreBuildSettings`.

natemcmaster:
cc @pranavkm @anurse @pakrym  you may also want to take a look before this is merged. The essential change here is that we want to move `build.sh/ps1` => `run.ps1/sh` so that you can call `run docker-build` or `run something-else`. The current build behavior is `run.sh default-build`. We'll leave build.cmd and build.sh files in repos that are just abbreivations for calling `run.sh default-build`.

ryanbrandenburg:
For those new to the PR, once we've moved all the repo's to use run.sh/.ps1 we'll be able to do some more cleanup, like not exposing the Install-Tools and Invoke-RepositoryBuild functions in KoreBuild.psd1, eventually removing them entirely in favor of their more cross-platform friendly console command versions.

ryanbrandenburg:
I talked with @natemcmaster and we realized we need to do a little bit more work on the Docker portion of this (ToolsSource need to flow through to the docker image) so that people get their expected results, but there's no reason that should delay the rest of this PR. I'll remove the Docker portions for now and submit them as a separate PR.

anurse:
Yeah, I had some concerns about the docker stuff as well that will need to be considered. The main one was that we seem to be using the same names for Images and Containers, which is a bad idea. We should try to use properly qualified Image names so that they don't interfere with images from other sources. I'd suggest names like `aspnetbuild/debian:jessie` and `aspnetbuild/ubuntu:14.04`.

PR #367, opened 2017-08-10
Bundle Internal.AspNetCore.Sdk in KoreBuild

### LINKED ISSUE #297: Ensure Internal.AspNetCore.Sdk aligns with the version of KoreBuild

natemcmaster:
- Adds API check and Internal.AspNetCore.SDK into the KoreBuild artifact.
 - Add a "FolderLineup" policy which will (1) pin the version of Internal.AspNetCore.Sdk to match KoreBuild and (2) add a local feed to the extracted location of these bundled packages

Resolves #297


test/NuGet.Tasks.Tests/PackageLineupPolicyTests.cs
```diff
@@ -22,8 +22,18 @@
 
 namespace NuGet.Tasks.Tests
 {
-    public class PackageLineupPolicyTests
+    public class PackageLineupPolicyTests : IDisposable
```
ryanbrandenburg: I didn't know xunit knew how deal with IDisposable stuff, neat.

natemcmaster:
I've been testing this against Kestrel. Once this is in, I'm opening a PR to kestrel to start using lineups.

test/NuGet.Tasks.Tests/PackageLineupPolicyTests.cs
natemcmaster: Yup. xunit will create an instance of the test class for each test method and call .Dispose to cleanup after, regardless of test outcome. The same is true of test fixtures and test collections.

PR #368, opened 2017-08-11
Make push timeout configurable

pakrym:
Site Extension is a large package and needs bigger timeout.


modules/NuGet.Tasks/PushNuGetPackages.cs
```diff
@@ -34,6 +33,8 @@ public class PushNuGetPackages : Microsoft.Build.Utilities.Task, ICancelableTask
         // not be required if pushing to the filesystem
         public string ApiKey { get; set; }
 
+        public int Timeout { get; set; } = 90;
```
natemcmaster: Can you rename to TimeoutSeconds? Makes it clearer which units this requires.

PR #370, opened 2017-08-14
Remove dependency on aspnetcore-ci-dev NuGet feed

natemcmaster:
Solves the 🐔 and 🍳 problem of building BuildTools if aspnetcore-ci-dev were to be cleared.

This required duplicating Microsoft.Extensions.CommandLineUtils.Sources into this repo.

cc @mikeharder - build from source

### DIFF
--- a/modules/BuildTools.Tasks/BuildTools.Tasks.csproj
+++ b/modules/BuildTools.Tasks/BuildTools.Tasks.csproj
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\shared\Microsoft.Extensions.CommandLineUtils.Sources\**\*.cs" />
     <None Include="module.props" CopyToPublishDirectory="PreserveNewest" />
     <None Include="BuildTools.Tasks.props" CopyToPublishDirectory="PreserveNewest" />
     <None Include="*.targets" CopyToPublishDirectory="PreserveNewest" />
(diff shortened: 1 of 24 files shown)


modules/BuildTools.Tasks/BuildTools.Tasks.csproj
```diff
@@ -9,16 +9,15 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\shared\Microsoft.Extensions.CommandLineUtils.Sources\**\*.cs" />
```
pranavkm: The 2.0.0 version of this package exists on aspnetcore-master (which isn't afflicted by the rooster and scrambled omelette problem).

natemcmaster: We'd still run into this in the future: https://github.com/aspnet/Coherence-Signed/issues/558

pranavkm: Isn't that solved the same way we'd solve all our other 3rd party dependency issues?

natemcmaster: Not exactly. 3rd party deps are on NuGet.org, but Ms.Ext.CommandLineUtils.Sources won't be.

pranavkm: Why does the feed location make a difference? I'm not trying to disavow checking in sources, I"m just trying to figure out how any of this is meant to work.

natemcmaster: The hypothetical scenario is that there is no feed at all. If we don't have any dotnet.myget.org feed available, we're stuck.

FWIW I wish we didn't have to copy source either, but it seems better than the alternative: git submodules.

PR #381, opened 2017-08-24
Add InstallDotNet task and stop installing 1.x by default

### LINKED ISSUE #347: Stop installing .NET Core 1.0 and 1.1 by default
Most repos only need .NET Core 2.0. A few need 1.0 and 1.1 for their build tools. Very few require a runtime for tests. We can stop installing these by default once we update the tools, but should find a way for the repos that still need it for testing to express their dependency on those runtimes.

natemcmaster:
Resolves #347. Apps that need 1.x runtimes must add these explicitly to their repo.props files.

Depends on #378 

cc @smitpatel @pakrym

### DIFF
--- a/modules/KoreBuild.Tasks/InstallDotNet.cs
+++ b/modules/KoreBuild.Tasks/InstallDotNet.cs
@@ -251,7 +251,7 @@ namespace KoreBuild.Tasks
                 all.Add(request);
             }
 
-            // installs SDKs first, which often bundle a shared runtime...making the shared runtime download unecessary.
+            // installs SDKs first, which often bundle a shared runtime...making the shared runtime download unnecessary.
             return all.Where(r => !r.IsSharedRuntime)
                 .Concat(all.Where(r => r.IsSharedRuntime));
         }
--- a/modules/KoreBuild.Tasks/InstallDotNet.cs
+++ b/modules/KoreBuild.Tasks/InstallDotNet.cs
@@ -1,0 +1,259 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Framework;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace KoreBuild.Tasks
+{
+    /// <summary>
+    /// Shim for executing dotnet-install.{sh,ps1}, allowing repos to specify the required .NET Core runtimes or to programmatically add new ones.
+    /// </summary>
+    public class InstallDotNet : Microsoft.Build.Utilities.Task, ICancelableTask
+    {
+        public readonly CancellationTokenSource _cts = new CancellationTokenSource();
+
+        /// <summary>
+        /// The .NET Core runtimes or SDKs to be installed.
+        ///
+        /// Supported metadata:
+        ///   Channel: default is ''
+        ///   Arch: default is 'x64'
+        ///   SharedRuntime: default is 'false'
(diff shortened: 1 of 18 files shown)


modules/KoreBuild.Tasks/InstallDotNet.cs
```diff
@@ -0,0 +1,269 @@
...
+            var ext = Path.GetExtension(InstallScript).ToLowerInvariant();
+            if (ext == ".sh")
+            {
+                defaultArgs.Add("bash");
+                defaultArgs.Add(InstallScript);
+            }
+            else if (ext == ".ps1")
+            {
+                defaultArgs.Add("powershell");
+                defaultArgs.Add("-NoProfile");
```
pakrym: `ExecutionPolicy Unrestricted`?

natemcmaster: Yup, that would help.

modules/KoreBuild.Tasks/InstallDotNet.cs
```diff
@@ -0,0 +1,269 @@
...
+
+                    if (ReferenceEquals(finished, timeout))
+                    {
+                        Log.LogError($"dotnet-install of {assetName} timed out after {TimeoutSeconds} seconds");
+                        return false;
+                    }
+
+                    if (process.ExitCode != 0)
+                    {
+                        Log.LogError($"dotnet-install failed on {assetName}. Arguments: {process.StartInfo.FileName} {process.StartInfo.Arguments}");
```
pakrym: There was a bug with powershell where it returned 0 exit code even when script failed. So we have to do `-Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0build.ps1' %*; exit $LASTEXITCODE"` in `build.cmd`

modules/KoreBuild.Tasks/InstallDotNet.cs
```diff
@@ -0,0 +1,259 @@
...
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace KoreBuild.Tasks
+{
+    /// <summary>
+    /// Shim for executing dotnet-install.{sh,ps1}, allowing repos to specify the required .NET Core runtimes or to programmatically add new ones.
+    /// </summary>
+    public class InstallDotNet : Microsoft.Build.Utilities.Task, ICancelableTask
+    {
+        public readonly CancellationTokenSource _cts = new CancellationTokenSource();
```
ryanbrandenburg: This kind of duplicates some of what I'm doing in #363 but having it as a task is nice, and we could change this to call the KoreBuild command later on.

modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
```diff
@@ -8,6 +8,7 @@
   <ItemGroup>
     <Content Include="*.props" CopyToPublishDirectory="PreserveNewest" />
     <Content Include="*.targets" CopyToPublishDirectory="PreserveNewest" />
+    <Compile Include="..\..\shared\Microsoft.Extensions.CommandLineUtils.Sources\Utilities\*.cs" />
```
ryanbrandenburg: We have near identical code for this in a couple of repos now, should we just start publishing it? I'm not a fan of the ".\shared\" pattern.

natemcmaster: It is published as nuget package from aspnet/Common, but because aspnet/BuildTools can't depend on aspnetcore-ci-dev, we made a copy of it in this repo.

ryanbrandenburg: Ah makes sense.

modules/KoreBuild.Tasks/InstallDotNet.cs
natemcmaster: We can take a look at that later on.

ryanbrandenburg: Yeah, more of a note than anything.

PR #382, opened 2017-08-25
Add support for lineup pinning based on the project's target framework

natemcmaster:
Adds support for specifying dependencies whose version vary based on target framework.

For example, a lineup could look like this:
```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
  <metadata>
    <id>TestLineup</id>
    <version>1.0.0</version>
    <authors>Test</authors>
    <description>Test</description>
    <packageTypes>
      <packageType name="lineup" />
    </packageTypes>
    <dependencies>
      <group targetFramework=".NETCoreApp1.1">
        <dependency id="Microsoft.NETCore.App" version="1.1.2" />
      </group>
      <group targetFramework=".NETCoreApp2.0">
        <dependency id="Microsoft.NETCore.App" version="2.0.0" />
      </group>
    </dependencies>
  </metadata>
</package>
```

cc @mikeharder @pranavkm

### DIFF
--- a/modules/KoreBuild.Tasks/Policies/Lineup/PackageVersionSource.cs
+++ b/modules/KoreBuild.Tasks/Policies/Lineup/PackageVersionSource.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using NuGet.Common;
+using NuGet.Frameworks;
 using NuGet.Packaging;
 using NuGet.Packaging.Core;
 using NuGet.Versioning;
@@ -14,25 +15,25 @@ namespace KoreBuild.Tasks.Lineup
     internal class PackageVersionSource
     {
         private readonly ILogger _logger;
-        private Dictionary<string, (string lineup, PackageIdentity identity)> _packageLineupMap;
+        private IDictionary<string, IDictionary<NuGetFramework, LineupVersion>> _packageLineupMap;
         private List<PackageIdentity> _lineups;
 
         public PackageVersionSource(ILogger logger)
         {
             _logger = logger;
-            _packageLineupMap = new Dictionary<string, (string, PackageIdentity)>(StringComparer.OrdinalIgnoreCase);
+            _packageLineupMap = new Dictionary<string, IDictionary<NuGetFramework, LineupVersion>>(StringComparer.OrdinalIgnoreCase);
             _lineups = new List<PackageIdentity>();
         }
 
         public IEnumerable<PackageIdentity> Lineups => _lineups;
 
-        public void AddPackagesFromLineup(string lineupId, NuGetVersion lineupVersion, List<PackageDependencyGroup> dependencies)
+        public void AddPackagesFromLineup(string lineupId, NuGetVersion lineupVersion, IReadOnlyCollection<PackageDependencyGroup> dependencyGroups)
         {
             _lineups.Add(new PackageIdentity(lineupId, lineupVersion));
 
-            foreach (var dep in dependencies)
+            foreach (var depGroup in dependencyGroups)
             {
-                foreach (var package in dep.Packages)
+                foreach (var package in depGroup.Packages)
                 {
                     if (!package.VersionRange.HasLowerBound)
                     {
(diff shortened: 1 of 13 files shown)


modules/KoreBuild.Tasks/Policies/Lineup/PackageVersionSource.cs
```diff
@@ -70,33 +71,51 @@ public void AddPackagesFromFolder(string folder)
...
+            if (!_packageLineupMap.TryGetValue(id, out var versions))
             {
-                version = info.identity.Version.ToNormalizedString();
-                return true;
+                version = null;
+                return false;
             }
 
-            return false;
+            var lineupFramework = NuGetFrameworkUtility.GetNearest(versions.Keys, targetFramework, f => f);
```
pranavkm: Is `GetNearest` the right thing to use here? I'd expect we would have exact TFMs. For instance, we wouldn't want an app targeting `netcoreapp2.2` to pick up `M.N.A 2.1.0` simply because it's the closest available TFM.

natemcmaster: You bring up a good point. My thinking was this: there are two ways to handle TFMs from the lineup package. As an example, let's say we want PackageA 2.0.0 for .NET Framework projects, but otherwise use PackageA 1.0.0.

(1) exact TFM must match
Pros: tighter control, more predictable behavior. 
Cons: you must explicitly specify every package for every possible TFM that will ever be used. The lineup file becomes harder to author and contains lots of duplicate info.
```xml
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="PackageA" version="1.0.0" />
      </group>
      <group targetFramework=".NETCore2.0">
        <dependency id="PackageA" version="1.0.0" />
      </group>
      <group targetFramework=".NETCore2.1">
        <dependency id="PackageA" version="1.0.0" />
      </group>
      <group targetFramework=".NETFramework4.6">
        <dependency id="PackageA" version="2.0.0" />
      </group>
      <group targetFramework=".NETFramework4.6.1">
        <dependency id="PackageA" version="2.0.0" />
      </group>
      <group targetFramework=".NETFramework4.6.2">
        <dependency id="PackageA" version="2.0.0" />
      </group>
      <group targetFramework=".NETFramework4.7">
        <dependency id="PackageA" version="2.0.0" />
      </group>
    </dependencies>
```

(2) `GetNearest`
**Pros:** we can use `AnyFramework` for most dependencies. Future TFMs follow existing compatibility fallback.
**Cons:** As you identified, there is a possibilty that when new TFMs are introduced, that fallback is wrong.
```xml
    <dependencies>
      <group>
        <dependency id="PackageA" version="1.0.0" />
      </group>
      <group targetFramework=".NETFramework4.6">
        <dependency id="PackageA" version="2.0.0" />
      </group>
    </dependencies>
```

natemcmaster: So far, I've chosen `GetNearest` as I think the `netcoreapp2.2` problem is easier to solve than the authoring problem of exact TFMs only.

pranavkm: I guess if you're optimizing for authoring ease, sure. I'm not entirely sure why you chose nuspec as a file format which boxes you in to this sort of design.

natemcmaster: > I'm not entirely sure why you chose nuspec as a file format which boxes you in to this sort of design.

We considered other formats briefly in design meeting, but none of them offered obvious benefit over using NuGet's package metadata. That said, I'm not dismissing the idea. Lineups are not set in stone yet. I'd love to hear new ideas for how to implement this better.

pranavkm: Maybe a mapping of TFM ranges per package? But I don't have enough context so I'll let trust your choice here 👍

natemcmaster: That's a good point. We don't have a way to express TFM ranges in nuspec metadata. If we had a new format, we could add those kinds of rules.

I'm trying to 🔮 a little and think if there are other behaviors we might want that we can't do with nuget metadata. I can't think of anything, but I'll add this to a new issue so we can discuss more.

natemcmaster: #384

PR #391, opened 2017-09-01
Show a warning when there is a duplicate package reference instead of failing

natemcmaster:
Make it easier to diagnose when multiple PackageReference's for the same package exist


natemcmaster:
> Consider adding a test

Will do, after vacation of course.

PR #393, opened 2017-09-07
Call "sync" between "chmod" and execution

mikeharder:
- Prevents "text file busy" error in Docker (aufs)


mikeharder:
@ryanbrandenburg: It might be a bug in the Docker "aufs" storage driver (https://github.com/moby/moby/issues/9547), but this is the default storage driver on many platforms so I think it's worth adding the workaround to our scripts to prevent issues in Docker, for both ourselves and external contributors.

PR #394, opened 2017-09-07
Add docker support

ryanbrandenburg:
Add the docker-build command to KoreBuild. Fixes aspnet/Coherence-Signed#647.


files/KoreBuild/KoreBuild.sh
```diff
@@ -13,7 +14,7 @@ set_korebuildsettings() {
     local config_file="${4:-}" # optional. Not used yet.
 
     [ -z "${dot_net_home:-}" ] && dot_net_home="$HOME/.dotnet"
-    [ -z "${tools_source:-}" ] && tools_source='https://aspnetcore.blob.core.windows.net/buildtools'
+    [ -z "${tools_source:-}" ] && tools_source=default_tools_source
```
anurse: Missing a `$`? `tools_source=$default_tools_source`

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,172 @@
...
+                dockerToolsSource = "ToolsSource";
+                toolsSourceDestination = Path.Combine(RepoPath, dockerToolsSource);
+                DirectoryCopy(ToolsSource, toolsSourceDestination);
+            }
+
+            try
+            {
+                var buildArgs = new List<string> { "build" };
+
+                var tag = $"{ImagePrefix}:{ContainerName}";
```
anurse: First off: we shouldn't mix the container name into the image name, it can add confusion.

Docker image tag names are generally in the form: `[owner]/[name]:[variant]`. Here, `owner` should map to an account on Docker Hub (eventually, doesn't have to right now), I think "aspnetbuild" is a good name. `name` generally refers to the content of the image (regardless of base OS, etc.), so `korebuild` or `build` would be good names. `variant` is used to distinguish between versions and OSes usually. Take a look at the `microsoft/dotnet` image for some examples: https://hub.docker.com/r/microsoft/dotnet/

I'd suggest something like `aspnetbuild/korebuild:jessie`. We may even want to get in to version number tags at some point, but I think we can start with this, since we're always building the container "fresh" (using caches) rather than pulling from Docker Hub.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        public CommandArgument Platform { get; set; }
+
+        public List<string> Arguments {get; set; }
+        
+        public string ContainerName { get; set; } = "testcontainer";
+
+        public override void Configure(CommandLineApplication application)
+        {
+            Platform = application.Argument("platform", "The docker platform to run on.");
```
anurse: I'd suggest using a term like `Image` for this (I realize it makes up only part of the image name, see comment below, but I think it's a clearer name regardless). It's a little more aligned with Docker terminology. Platform makes me thing of Docker for Windows vs Docker for Linux.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
anurse: Combined with the above comment re:"Platform", this string would become something like `var tag = $"{ImagePrefix}:{Image}` and `ImagePrefix` becomes `aspnetbuild/korebuild`

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,172 @@
...
+                var tag = $"{ImagePrefix}:{ContainerName}";
+                buildArgs.AddRange(new string[] { "-t", tag, "-f", dockerFileDestination, RepoPath });
+                var buildResult = RunDockerCommand(buildArgs);
+
+                if (buildResult != 0)
+                {
+                    return buildResult;
+                }
+
+                var runArgs = new List<string> { "run", "--rm", "-it", "--name", ContainerName, tag };
```
anurse: If we're using `--name` we should use a uniquely-generated (possibly timestamp-based?) name, perhaps something like `aspnetbuild_[timestamp]`. If there's already a container with this name, it will conflict.

Also, we should make it possible to disable `--rm`. If my build fails, it would be nice to have the container stick around.

Finally, make sure the container name is dumped to the console so I know which containers come from which builds.

tools/KoreBuild.Console/Commands/DockerFiles/jessie.dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM microsoft/dotnet:2.0-runtime-deps-jessie
+
+RUN apt-get update && \
+    apt-get install -y git && \
+    apt-get install -y curl && \
+    apt-get install -y unzip && \
+    apt-get install -y apt-transport-https
```
anurse: Add `rm -rf /var/lib/apt/lists/*` to the end of this chain. See https://github.com/anurse/ConferencePlanner/blob/735d5ef3352d18c7a258267a1f18aad359479ec2/docker/base.Dockerfile#L9

That deletes the files `apt-get update` downloaded, which are not insignificant. You need to chain it in to the `RUN` command so that it all ends up being baked as one layer in the image. That ensures that the layer includes only the files added by the packages you installed, and no other cruft.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
mikeharder: `--name` is most useful for long-running containers.  In this scenario, I would just use the random name assigned by Docker.

tools/KoreBuild.Console/Commands/DockerFiles/jessie.dockerfile
mikeharder: You also only need one call to `apt-get install`.  Example:

https://github.com/aspnet/benchmarks/blob/dev/docker/benchmarks/Dockerfile#L3

tools/KoreBuild.Console/Commands/DockerFiles/jessie.dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ADD ./ ./
```
mikeharder: I've typically see this as `ADD . .`, though I think the two are equivalent.

tools/KoreBuild.Console/Commands/DockerFiles/jessie.dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN rm -f ./korebuild-lock.txt
```
mikeharder: I think it would be better to exclude this via a `.dockerignore` file, rather than copying and then removing.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
ryanbrandenburg: I figured you wouldn't ever run two at the same time due to `--rm`, but I guess it's still possible, and as below I would like to let you run interactive in the future, so I'll do that.

I had intended to add disabling `--rm` as an option later on. I was intending to get this running in an actual scenario (one of the CI builds) before complicating things with more options. I just filed #395 for that. Is that good or do you think that option is more important that that?

Good point about the container name.

anurse: I think `--name` becomes relevant when `--rm` is removable. @mikeharder is definitely right that if the container is just going to be removed after the build, there's no need to know the name.

Perhaps the best solution is to remove `--name` for now, and when we make `--rm` removable, we can also add an option to the `docker-build` command to allow providing a container name.

tools/KoreBuild.Console/Commands/DockerFiles/jessie.dockerfile
anurse: 👍 more a performance that space saving issue, but definitely a good one.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
ryanbrandenburg: I'm actually partial to supplying the name ourselves because:
1. We can easily output the name in case it persists for some reason (including future interactivity).
2. It makes it easy to tell at a glance which containers belong to KoreBuild.
3. It doesn't hurt anything (unless there's something I don't know).

I do agree that allowing people to specify the container name in the future would be a nice feature. I created #396 to enumerate all the future improvements we think of while doing this work.

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM microsoft/aspnet:4.6.2
+
+ADD https://download.microsoft.com/download/F/1/D/F1DEB8DB-D277-4EF9-9F48-3A65D4D8F965/NDP461-DevPack-KB3105179-ENU.exe /net461dev.exe
+
+RUN /net461dev.exe /Passive /NoRestart
+
+RUN rm /net461dev.exe 
+
+ADD . c:/repo/
```
ryanbrandenburg: I decided not to dump this in c:\ because it's cleaner, but also because I ran into a problem with running KoreBuild from the drive root on windows.

ryanbrandenburg:
:up::date:

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
mikeharder: I've typically seen Dockerfiles use the following order of statements instead:

```
WORKDIR c:/repo/

ADD . .
```

mikeharder: Also, our docs (https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-docker/manage-windows-dockerfile) recommend escaped backslashes rather than forward slashes, though both may work:

```
WORKDIR c:\\windows
```

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM microsoft/aspnet:4.6.2
+
+ADD https://download.microsoft.com/download/F/1/D/F1DEB8DB-D277-4EF9-9F48-3A65D4D8F965/NDP461-DevPack-KB3105179-ENU.exe /net461dev.exe
```
mikeharder: You should be able to download, install, and remove in a single command (which is more efficient) like so:

https://github.com/Microsoft/aspnet-docker/blob/master/4.6.2/Dockerfile#L12-L16

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN /net461dev.exe /Passive /NoRestart
+
+RUN rm /net461dev.exe 
+
+WORKDIR c:\\repo\\
```
mikeharder: You need to remove the trailing backslash, otherwise it seems Docker treats it as a line continuation:

```
Step 5/7 : WORKDIR c:\\repo\ADD . .
```

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
mikeharder: This is actually somewhat complicated since `NDP461-DevPack-KB3105179-ENU.exe /Passive /NoRestart` immediately exits with error code 0, and the install runs in the background.  So, the delete will fail unless you wait a bit after starting the install.  In your original Dockerfile, I think there's a small delay between the two run commands which usually causes the delete to succeed, but it's not guaranteed.  And technically, I don't think it's guaranteed the install will be complete by the time `build.cmd` tries to use it.  The following should work, but there's no way to know the correct length of time to sleep:

```
RUN Invoke-WebRequest https://download.microsoft.com/download/F/1/D/F1DEB8DB-D277-4EF9-9F48-3A65D4D8F965/NDP461-DevPack-KB3105179-ENU.exe \
      -OutFile C:\NDP461-DevPack-KB3105179-ENU.exe; \
    C:\NDP461-DevPack-KB3105179-ENU.exe /Passive /NoRestart; \
    Start-Sleep -s 10; \
    Remove-Item C:\NDP461-DevPack-KB3105179-ENU.exe -Force;
```

The most reliable way would be to block until we know the install is complete, perhaps using the existence of a file or registry key.

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM microsoft/aspnet:4.6.2
+
+WORKDIR c:\\repo
```
mikeharder: Move `WORKDIR` after `RUN` to improve caching, so if you decide to change the `WORKDIR` you don't need to re-execute the `RUN` command.

files/KoreBuild/KoreBuild.sh
```diff
@@ -71,10 +72,11 @@ install_tools() {
 
     # Instructs MSBuild where to find .NET Framework reference assemblies
     export ReferenceAssemblyRoot="$tools_home/netfx/$netfx_version"
-    
+
     # Call "sync" between "chmod" and execution to prevent "text file busy" error in Docker (aufs)
     chmod +x "$__korebuild_dir/scripts/get-netfx.sh"; sync
-    "$__korebuild_dir/scripts/get-netfx.sh" $verbose_flag $netfx_version "$tools_source" "$ReferenceAssemblyRoot" \
+    # we don't include netfx in the BuildTools artifacts currently, it ends up on the blob store through other means, so we'll only look for it in the default_tools_source
+    "$__korebuild_dir/scripts/get-netfx.sh" $verbose_flag $netfx_version "$default_tools_source" "$ReferenceAssemblyRoot" \
```
natemcmaster: Why this change? This should be reverted as it makes it impossible to override `--tools-source` for the netfx asset.

test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
```diff
@@ -55,5 +56,23 @@ public SimpleRepoTests(ITestOutputHelper output, RepoTestFixture fixture)
             Assert.Same(task, build);
             Assert.NotEqual(0, build.Result);
         }
+
+        [Fact]
+        public async Task DockerSuccessful()
+        {
+            var app = _fixture.CreateTestApp("SimpleRepo");
+            var platform = "jessie";
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
natemcmaster: You should check which platform the docker host is using, not which platform is running this test. Windows machines can use Linux containers.

Also, I would be good to make this test skip if docker is not installed.

```
docker version -f "{{ .Server.Os }}"
```

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Reflection;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace KoreBuild.Console.Commands
+{
+    internal class DockerBuildCommand : SubCommandBase
+    {
+        private const string DockerIgnore = ".dockerignore";
+        private const string DockerfileExtension = ".dockerfile";
+        private const string Owner = "aspnetbuild";
```
natemcmaster: Do we already own this docker hub account? Or are you planning on using a private container registry?

tools/KoreBuild.Console/Commands/DockerFiles/winservercore.dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# DevPack returns exit 0 immediately, but it's not done, so we wait.
+# A more correct thing would be to block on a registry key existing or similar.
+RUN \
+        Invoke-WebRequest https://download.microsoft.com/download/F/1/D/F1DEB8DB-D277-4EF9-9F48-3A65D4D8F965/NDP461-DevPack-KB3105179-ENU.exe -OutFile .\\net461dev.exe ; \
```
natemcmaster: Hopefully this will be unnecessary once we fix https://github.com/aspnet/BuildTools/issues/399

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
```diff
@@ -0,0 +1,182 @@
...
+            File.Copy(dockerIgnoreSource, dockerIgnoreDestination, overwrite: true);
+
+            // If our ToolSource isn't http copy it to the docker context
+            var dockerToolsSource = ToolsSource;
+            string toolsSourceDestination = null;
+            if (!ToolsSource.StartsWith("http"))
+            {
+                dockerToolsSource = "ToolsSource";
+                toolsSourceDestination = Path.Combine(RepoPath, dockerToolsSource);
+                DirectoryCopy(ToolsSource, toolsSourceDestination);
```
natemcmaster: An easier and faster way to do this to volume mount this folder into the container.

test/KoreBuild.FunctionalTests/Utilities/TestApp.cs
```diff
@@ -41,22 +51,22 @@ void Write(object sender, DataReceivedEventArgs e)
...
                 cmd = "bash";
-                arguments.Add("./build.sh");
+                arguments.Add($"./{script}.sh");
             }
 
             arguments.AddRange(new[]
             {
-                "-ToolsSource", _toolsSource,
-                "-Update"
+                "-s", _toolsSource,
```
natemcmaster: Nit: I personally prefer the readability of the long option name. The short option is good for command-line shorthand, but harder to interpret from logs and code.

files/KoreBuild/KoreBuild.sh
ryanbrandenburg: Like the comment says, overriding --tools-source for the netfx asset doesn't work right now. Currently the buildtools artifact does not contain netfx, it gets added to the blob store by some other process. If we want to fix that I feel it's better dealt with in another issue/PR.

tools/KoreBuild.Console/Commands/DockerBuildCommand.cs
anurse: We should probably claim it if we can. I think it would be useful to take the simple steps to prepare to use a registry of some kind.

ryanbrandenburg: Squated.

ryanbrandenburg:
:up::date:

ryanbrandenburg:
ping @natemcmaster.

files/KoreBuild/KoreBuild.sh
natemcmaster: Ok, we'll track that here: https://github.com/aspnet/BuildTools/issues/399

scripts/bootstrapper/run.sh
```diff
@@ -218,6 +218,7 @@ fi
 [ -z "$channel" ] && channel='dev'
 [ -z "$tools_source" ] && tools_source='https://aspnetcore.blob.core.windows.net/buildtools'
 
+__warn "ts: $tools_source"
```
natemcmaster: Warn?

test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
```diff
@@ -55,5 +58,85 @@ public SimpleRepoTests(ITestOutputHelper output, RepoTestFixture fixture)
             Assert.Same(task, build);
             Assert.NotEqual(0, build.Result);
         }
+        
+        [DockerExistsFact]
```
natemcmaster: 👏

PR #401, opened 2017-09-14
Stub out 'framework build'

natemcmaster:
This doesn't work build yet. Just stubbing out the flow of information.

### DIFF
--- a/sdk/KoreBuild/KoreBuild.Common.props
+++ b/sdk/KoreBuild/KoreBuild.Common.props
@@ -37,6 +37,7 @@ Default layout and configuration.
     <ArtifactsDir>$(RepositoryRoot)artifacts\</ArtifactsDir>
     <BuildDir>$(ArtifactsDir)build\</BuildDir>
     <IntermediateDir>$(RepositoryRoot)obj\</IntermediateDir>
+    <ProjectType>Solution</ProjectType>
   </PropertyGroup>
 
 
--- a/sdk/KoreBuild/modules/frameworkbuild/1-prepare.targets
+++ b/sdk/KoreBuild/modules/frameworkbuild/1-prepare.targets
@@ -1,0 +1,56 @@
+<Project>
+  <PropertyGroup>
+    <!-- CleanArtifacts is defined in KoreBuild already  -->
+    <PrepareDependsOn>
+      $(PrepareDependsOn);
+      CleanArtifacts;
+      PrepareSourceCode;
+      PrepareProjects;
+    </PrepareDependsOn>
+  </PropertyGroup>
+
+  <Target Name="PrepareProjects" DependsOnTargets="PrepareSourceCode">
+    <!-- TODO implement task -->
+    <ResolveBuildGraph
+      Repositories="@(Repository)"
+      Properties="$(RepoBuildProps)">
+
+      <!--
+      Identity = RepoDir
+      Metadata:
+        - Name
+        - RepoDir
+        - CommitHash
+      -->
+      <Output TaskParameter="KoreBuildProjects" ItemGroup="KoreBuildProject" />
+
+      <!-- A list of the individual csproj files that will be compiled. -->
+      <Output TaskParameter="MSBuildProjects" ItemGroup="MSBuildProjects" />
+
+      <!-- A list of any PackageReferences that are not being produced by these other projects -->
+      <Output TaskParameter="ExternalPackageReferences" ItemGroup="ExternalPackageReferences" />
+
+      <!-- Packages to be produced (Id/Version) -->
+      <!-- TODO later verify these are actually produced -->
+      <Output TaskParameter="OutputPackages" ItemGroup="ProducedPackages" />
+    </ResolveBuildGraph>
+
+    <Error
+      Text="Currently we only build korebuild projects."
+      Condition="@(Repository->Count()) != @(KoreBuildProject->Count())" />
+
+    <!-- Pins floating PackageReference versions to versions matching the projects that will build them. -->
+    <LinkPackageReferences
+      Projects="@(MSBuildProjects)"
+      Packages="@(ProducedPackages)" />
+  </Target>
+
+  <Target Name="PrepareSourceCode">
+    <!-- Does the needful git clone, checkout, clean, etc. to ensure we have source code ready. -->
+    <GetSourceCode
+      Repositories="@(Repository)"
+      BaseDir="$(IntermediateDir)src\"
+      EnsureCleanCheckout="false" />
+  </Target>
+
+</Project>
--- a/sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
+++ b/sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
@@ -1,0 +1,75 @@
+<Project>
+  <PropertyGroup>
+  	<CompileDependsOn>$(CompileDependsOn);CompileProjects;CompileRuntimeStore</CompileDependsOn>
+  </PropertyGroup>
+
+  <Target Name="CompileProjects" DependsOnTargets="PrepareProjects">
+    <KoreBuild
+      Projects="@(KoreBuildProject)"
+      Targets="Package"
+      BuildInParallel="true"
+      Properties="$(RepoBuildProps)" />
+
+    <PropertyGroup>
+      <!-- compilation done! Skip in future invocations -->
+      <RepoBuildProps>$(RepoBuildProps);NoBuild=true</RepoBuildProps>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <SymbolsPackageArtifactFile Include="%(KoreBuildProject.RootDir)artifacts\build\*.symbols.nupkg" />
+      <PackageArtifactFile Include="%(KoreBuildProject.RootDir)artifacts\build\*.nupkg" Exclude="@(SymbolsPackageArtifactFile)" />
+      <AdditionalArtifactFiles Include="%(KoreBuildProject.RootDir)artifacts\**\*" Exclude="@(PackageArtifactFile)" Repository="%(KoreBuildProject.Name)" />
+    </ItemGroup>
+
+    <!-- TODO do we still need this? -->
+    <Copy SourceFiles="@(PackageArtifactFile)" DestinationFolder="$(BuildDir)" />
+    <Copy SourceFiles="@(SymbolsPackageArtifactFile)" DestinationFolder="$(ArtifactsDir)symbols\" />
+
+    <Copy
+      SourceFiles="%(AdditionalArtifactFiles.Identity)"
+      DestinationFolder="$(ArtifactsDir)%(Repository)\%(RecursiveDir)%(FileName)%(Extension)" />
+
+    <!-- TODO implement task. Reads a nupkg file and creates and item with metadata about its id, version, and file location -->
+    <GetPackageIdentity Files="@(PackageArtifactFile)">
+
+      <Output TaskParameter="PackageIdentity" ItemName="PackageReference" />
+    </GetPackageIdentity>
+
+    <!-- TODO maybe we can make this more generic? e.g. just assign a Category metadata? -->
+    <!-- TODO implement task -->
+    <!-- Categories packages -->
+    <ClassifyPackages
+      CsvFile="$(RepositoryRoot)packages.csv"
+      Packages="@(PackageReference)">
+
+      <!-- Translate the package file into Id/Version/Source (directory) -->
+      <Output TaskParameter="PartnerPackages" ItemName="PartnerPackage" />
+      <Output TaskParameter="ShippingPackages" ItemName="ShippingPackage" />
+      <Output TaskParameter="ShipOobPackages" ItemName="ShipOobPackage" />
+      <Output TaskParameter="NonshippingPackages" ItemName="NonshippingPackage" />
+      <Output TaskParameter="RuntimeStorePackages" ItemName="RuntimeStorePackage" />
+      <Output TaskParameter="SymbolsPackages" ItemName="SymbolsPackage" />
+    </ClassifyPackages>
+
+    <!-- Also copy other files based on their category-->
+    <Copy SourceFiles="@(SymbolsPackageArtifactFile->'%(FullPath)')" DestinationFolder="$(ArtifactsDir)symbols\" />
+    <Copy SourceFiles="@(ShippingPackage->'%(FullPath)')" DestinationFolder="$(ArtifactsDir)ship\" />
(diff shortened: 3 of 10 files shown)


sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
```diff
@@ -0,0 +1,75 @@
...
+    <GetPackageIdentity Files="@(PackageArtifactFile)">
+
+      <Output TaskParameter="PackageIdentity" ItemName="PackageReference" />
+    </GetPackageIdentity>
+
+    <!-- TODO maybe we can make this more generic? e.g. just assign a Category metadata? -->
+    <!-- TODO implement task -->
+    <!-- Categories packages -->
+    <ClassifyPackages
+      CsvFile="$(RepositoryRoot)packages.csv"
```
natemcmaster: Should we keep packages.csv or use another format? i.e. MSBuild items

JunTaoLuo: packages.csv for now I'd assume. There's parsing logic in coherence that assumes that format. Unless we want to change the format there now as well. https://github.com/aspnet/Coherence/blob/f7fcc555ffa897ef1c115f3d2c9fd4220752e6be/build/tasks/CopyPackagesToSplitFolders.cs

JunTaoLuo: Oops that's the dev version. I think rel/2.0.0 is here: https://github.com/aspnet/BuildTools/tree/rel/2.0.0/src/SplitPackages

sdk/KoreBuild/KoreBuild.Common.props
```diff
@@ -37,6 +37,7 @@ Default layout and configuration.
     <ArtifactsDir>$(RepositoryRoot)artifacts\</ArtifactsDir>
     <BuildDir>$(ArtifactsDir)build\</BuildDir>
     <IntermediateDir>$(RepositoryRoot)obj\</IntermediateDir>
+    <ProjectType>Solution</ProjectType>
```
natemcmaster: Just an idea. Another way to do this is to follow the MSBuild convention of \*.\*proj extension being used to determine project type.

e.g., Universe might instead look like this:
```
- src/
    - Microsoft.AspNetCore.All/
        + Microsoft.AspNetCore.All.fxproj
```
```xml
<!-- Microsoft.AspNetCore.All.fxproj -->
<Project>
  <Import Project="KoreBuild.Common.props" />

  <ItemGroup>
     <Repository Id="Antiforgery" Branch="rel/2.0.2" />
     <!-- the rest of the stuff. -->
  </ItemGroup>

  <Import Project="KoreBuild.Common.targets" />
</Project>
```

Example of this kind of layout: https://github.com/aspnet/Coherence-Final/blob/dev/packaging/1.0.6/AspNetCore.PackageCache.1.0.6.pkgproj

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
natemcmaster: Let's copy the CopyPackagesToSplitFolders.cs task. SplitPackages did a whole bunch of other stuff that it didn't need to.

sdk/KoreBuild/modules/frameworkbuild/module.targets
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <ImportGroup Condition=" '$(ProjectType)' == 'Framework' ">
+    <Import Project="1-prepare.targets" />
```
JunTaoLuo: Weird skipping step 2. Let's add it back in even if it's a placeholder.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
JunTaoLuo: Will also need to update the CopyPackagesToSplitFolders.cs task. It currently parses the csv but it should be trivial to take an ItemGroup as input instead and operate on the individual metadata.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
```diff
@@ -0,0 +1,75 @@
...
+      SourceFiles="%(AdditionalArtifactFiles.Identity)"
+      DestinationFolder="$(ArtifactsDir)%(Repository)\%(RecursiveDir)%(FileName)%(Extension)" />
+
+    <!-- TODO implement task. Reads a nupkg file and creates and item with metadata about its id, version, and file location -->
+    <GetPackageIdentity Files="@(PackageArtifactFile)">
+
+      <Output TaskParameter="PackageIdentity" ItemName="PackageReference" />
+    </GetPackageIdentity>
+
+    <!-- TODO maybe we can make this more generic? e.g. just assign a Category metadata? -->
```
JunTaoLuo: This might not be a great idea since categories combinations might not scale with the number of columns. With the 2 columns right now you already have 6 possibilities {ship,noship,shipoob}X{include,exclude}. I like the idea of using a set of MSBuild items with metadata and then filter on those instead: e.g.
```
<ItemGroup>
  <RuntimeStore Include="@(Packages)" Condition="'%(OptimizedCache)' == 'include'" />
  <Shipping Include="@(Packages)" Condition="'%(Category)' == 'ship'" />
  <ImpossiblePackages Include="@(Packages)" Condition="'%(Category)' == 'noship' AND '%(OptimizedCache)' == 'include'" />
</ItemGroup>
```

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
```diff
@@ -0,0 +1,75 @@
+<Project>
+  <PropertyGroup>
+  	<CompileDependsOn>$(CompileDependsOn);CompileProjects;CompileRuntimeStore</CompileDependsOn>
```
JunTaoLuo: I'm assuming `CompileRuntimeStore` builds the individual runtime stores for each RID of the platforms we support. I wouldn't include it in the default compile targets of `CompileProjects` as this will need to be run on 4 different agents. Keep it as a separate target that's not in the default CompileDependsOn property.

sdk/KoreBuild/modules/frameworkbuild/4-package.targets
```diff
@@ -0,0 +1,64 @@
+<Project>
+  <PropertyGroup>
+    <PackageDependsOn>$(PackageDependsOn);PackageRuntimeStore;PackageMetaPackage</PackageDependsOn>
```
JunTaoLuo: Again, not sure how these package steps are supposed to be invoked. The PackageRuntimeStore target will need to be run 4 times whereas PackageMetaPackage only need to be run once. They don't belong in the same list of targets.

sdk/KoreBuild/modules/frameworkbuild/4-package.targets
```diff
@@ -0,0 +1,64 @@
...
+    <!-- TODO implement in buildtools. -->
+    <TarballArchive
+      File="$(ArtifactsDir)runtimestore\$(RuntimeStoreFileName).tar.gz"
+      WorkingDirectory="$(IntermediateDir).rs\"
+      Gzip="true"
+      SourceFiles="@(RuntimeStoreFiles)"
+      Condition="'$(IsLinux)' == 'true' OR '$(IsMacOs)' == 'true'" />
+
+    <!-- TODO add other properties to this task for the package manifest. -->
+    <CreateDebianPackage
```
JunTaoLuo: Not sure we can build the debian packages right after compilation. We need the dlls in the runtime store to be signed before packaging up the installers.

sdk/KoreBuild/modules/frameworkbuild/5-test.targets
```diff
@@ -0,0 +1,18 @@
...
+    <!-- TODO Parallel? NoBuild? -->
+    <KoreBuild
+      Projects="@(KoreBuildProject)"
+      Targets="TestProjects"
+      BuildInParallel="false"
+      Properties="$(RepoBuildProps);NoBuild=true" />
+  </Target>
+
+  <Target Name="TestRuntimeStore">
+    <!-- TODO add tests for the runtime store here -->
```
JunTaoLuo: @kichalla added tests for this by using MusicStore. We can take a look at the current CI configuration and mimic it here.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
natemcmaster: That's a good point. We should decouple CompileRuntimeStore from CompileProjects. You should be able to copmile the runtime store without re-compiling aspnetcore from sourc.e

FYI - my thinking was that you should be able to call `build /t:Compile` and it should compile all aspnetcore source and the runtime store. Likewise, `build /t:Package` would produce all nupkgs, runtime stores, installers, and any other kind of artifact that we ship as a "package" of one form or another.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
natemcmaster: starting to seem like the packages.csv is more effort than its worth. Maybe we should make these an ItemGroup with well-known metadata. i.e.

```xml
<ItemGroup>
  <ICanHazPackage Include="Microsoft.AspNetCore" OutputFolder="ship" Metapackage="true" />
</ItemGroup>
```

sdk/KoreBuild/modules/frameworkbuild/4-package.targets
natemcmaster: Would this tweak work?
```xml
<PropertyGroup>
   <PackageRuntimeStoreDependsOn>$(PackageRuntimeStoreDependsOn);CompileRuntimeStore;</PackageRuntimeStoreDependsOn>
</PropertyGroup>

<Target Name="PackageRuntimeStore" DependsOnTargets="$(PackageRuntimeStoreDependsOn)">
<!- ... -->
```

Coherence-signed could then do this:
```xml
<PropertyGroup>
   <PackageRuntimeStoreDependsOn>$(PackageRuntimeStoreDependsOn);SignRuntimeStoreBinaries;</PackageRuntimeStoreDependsOn>
</PropertyGroup>

<Target Name="SignRuntimeStoreBinaries" DependsOnTargets="CompileRuntimeStore">
<!- ... -->
```

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
JunTaoLuo: Yup that's the idea I was getting to in https://github.com/aspnet/BuildTools/pull/401#discussion_r138785937. The metadata names here are better though.

sdk/KoreBuild/modules/frameworkbuild/4-package.targets
JunTaoLuo: Worth a shot. I'm a little confused with the build steps though, so if you didn't have Coherence-Signed's targets, you'll end up packaging installers with unsigned dlls? How do we guarantee that the coherence-signed targets are always invoked?

natemcmaster: Yeah, I was just talking about this with Murat. I need to think this piece through more. We should whiteboard it.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
natemcmaster: No clue what to name all these groups of packages though.

sdk/KoreBuild/modules/frameworkbuild/1-prepare.targets
```diff
@@ -0,0 +1,56 @@
...
+      Identity = RepoDir
+      Metadata:
+        - Name
+        - RepoDir
+        - CommitHash
+      -->
+      <Output TaskParameter="KoreBuildProjects" ItemGroup="KoreBuildProject" />
+
+      <!-- A list of the individual csproj files that will be compiled. -->
+      <Output TaskParameter="MSBuildProjects" ItemGroup="MSBuildProjects" />
```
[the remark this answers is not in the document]
natemcmaster: We need to modify the PackageReference's for all MSBuild projects to ensure consistent versions. Currently called "pinversions" on rel/2.0.0 and "lineups" in dev. In this draft, I called it "LinkPackageReferences".

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
```diff
@@ -0,0 +1,75 @@
...
+
+    <ItemGroup>
+      <SymbolsPackageArtifactFile Include="%(KoreBuildProject.RootDir)artifacts\build\*.symbols.nupkg" />
+      <PackageArtifactFile Include="%(KoreBuildProject.RootDir)artifacts\build\*.nupkg" Exclude="@(SymbolsPackageArtifactFile)" />
+      <AdditionalArtifactFiles Include="%(KoreBuildProject.RootDir)artifacts\**\*" Exclude="@(PackageArtifactFile)" Repository="%(KoreBuildProject.Name)" />
+    </ItemGroup>
+
+    <!-- TODO do we still need this? -->
+    <Copy SourceFiles="@(PackageArtifactFile)" DestinationFolder="$(BuildDir)" />
+    <Copy SourceFiles="@(SymbolsPackageArtifactFile)" DestinationFolder="$(ArtifactsDir)symbols\" />
```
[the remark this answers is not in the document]
natemcmaster: I haven't been paying attention to that. Does this mean we stop producing .symbols packages altogether? Or that the way we produce them is new?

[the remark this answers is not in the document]
pranavkm: No symbol packages and pdbs go in the actual nupkg. I'm not sure if the intent was to back port to the patch builds.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
JunTaoLuo: ShipPackage, NoShipPackages, ShipOOBPackages, RuntimeStorePackages. These should be sufficient for now.

natemcmaster:
Discussed more in triage. Re-organizing the build like this is likely going to take more work than we have time for. Let's instead try to do the minimal possible work to get the patch building.

sdk/KoreBuild/modules/frameworkbuild/3-compile.targets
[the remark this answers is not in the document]
Eilon: We want to see if we can get source link into the patch as well. @ryanbrandenburg is working on it for `dev` in the meantime.

PR #407, opened 2017-09-20
Add SourceLink support

ryanbrandenburg:


src/Internal.AspNetCore.Sdk/build/Common.props
```diff
@@ -22,6 +22,8 @@ Usage: this should be imported once via NuGet at the top of the file.
     <PackageProjectUrl>https://asp.net</PackageProjectUrl>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
     <Serviceable Condition="'$(Configuration)' == 'Release'">true</Serviceable>
+    <DebugType>Portable</DebugType>
+    <SourceLinkCreate>true</SourceLinkCreate>
```
ryanbrandenburg: Does anyone know a good way to tell if a project is a src project at this point? Right now we're doing SourceLink for test projects, which shouldn't HURT anything, but is kind of a waste.

src/Internal.AspNetCore.Sdk/build/Common.props
```diff
@@ -22,6 +22,8 @@ Usage: this should be imported once via NuGet at the top of the file.
     <PackageProjectUrl>https://asp.net</PackageProjectUrl>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
     <Serviceable Condition="'$(Configuration)' == 'Release'">true</Serviceable>
+    <DebugType>Portable</DebugType>
```
natemcmaster: This is the default setting already.

https://github.com/dotnet/sdk/blob/cd25d6bbfc2aac61cc790c511900427ac64fdaa8/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.Sdk.props#L78

src/Internal.AspNetCore.Sdk/build/Common.props
natemcmaster: Move this to Common.targets so this is added after the csproj is evaluated, and condition on IsPackable.
```xml
<SourceLinkCreate Condition="'$(SourceLinkCreate )' == '' AND  '$(IsPackable)' == 'true'">true</SourceLinkCreate>
```

src/Internal.AspNetCore.Sdk/Internal.AspNetCore.Sdk.nuspec
```diff
@@ -9,6 +9,7 @@
     <copyright>$copyright$</copyright>
     <dependencies>
       <dependency id="Microsoft.AspNetCore.BuildTools.ApiCheck" version="$apicheckversion$" />
+      <dependency id="SourceLink.Create.CommandLine" version="2.2.1" />
```
natemcmaster: I'm hesitant to add new third-party dependencies to our build unless we need to. The requirements of build-from-source will mean we need to either (1) clone, compile and package SourceLink.Create.CommandLine  or (2) conditionally exclude this tool when building from source.

As an alternative, could source link be done as a post-processing step with existing symbols packages? It would avoid these issues.

cc @mikeharder

ryanbrandenburg: I used this package because of https://github.com/aspnet/Universe/issues/553#issuecomment-328742192. The code required here isn't super complicated, we could do it ourselves without having to edit PDB's afterwards.

natemcmaster: I think we should look into that. Putting this in Internal.AspNetCore.Sdk.nuspec makes it hard to conditionally exclude or skip it.

src/Internal.AspNetCore.Sdk/build/Common.props
ryanbrandenburg: As discussed offline, if we put it in Common.targets it doesn't get executed before we compile, and therefore the json doesn't end up in the PDBs.

ctaggart:
@ryanbrandenburg this looks great. As [mentioned](https://github.com/aspnet/Universe/issues/553#issuecomment-328774479), you may wish to add either a `SourceLinkOriginUrl` or a `SourceLinkUrl` to avoid a `git config --get remote.origin.url` call for each build. The SourceLinkOriginUrl is any git url such as [these](https://github.com/ctaggart/SourceLink/blob/master/Tests/When_parsing_links.cs#L8-L11). It is usd to calculate the SourceLinkUrl such as [this](https://github.com/ctaggart/SourceLink/blob/master/Tests/When_parsing_links.cs#L15).

ryanbrandenburg:
We're going to try a different route, closing this.

rgl:
@ryanbrandenburg care to share which route?

ryanbrandenburg:
I discussed it with @natemcmaster and we came to the conclusion that between the build from source effort, the desire to only SourceLink "src" projects, and our simple requirements, it would be simpler to write a SourceLink task ourselves and include it in Internal.AspNetCore.Sdk.

Issue #270, opened 2017-06-12
Update VersionTool to work on csproj


natemcmaster:
The "VersionTool" project in this repo only works with project.json. We'll need to updated it to work with csproj.

PR #411, opened 2017-09-22
Delete VersionTool

natemcmaster:
Resolves https://github.com/aspnet/BuildTools/issues/270

Removes our dependency on packages that were designed for project.json.

cc @JunTaoLuo


BuildTools.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26808.3004
+VisualStudioVersion = 15.0.26730.16
```
JunTaoLuo: Looks fine but why did this version go down?

natemcmaster: RTM vs internal previews

PR #416, opened 2017-10-07
Add a GenerateFileFromTemplate MSBuild task

natemcmaster:
We need this task to generate some files at build time using MSBuild properties as the values.


modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            var values = MSBuildListSplitter.GetNamedProperties(Properties);
+            var template = File.ReadAllText(TemplateFile);
+
+            var result = Replace(template, values);
+            Directory.CreateDirectory(Path.GetDirectoryName(outputPath));
+            File.WriteAllText(outputPath, result);
+
+            FileWrites = Path.GetFullPath(outputPath);
```
JunTaoLuo: Redundant `GetFullPath`? looks like `outputPath` is already the result of a `Path.GetFullPath`

shared/Utilities/MSBuildListSplitter.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.BuildTools
+{
+    internal static class MSBuildListSplitter
+    {
+        public static IDictionary<string, string> GetNamedProperties(string[] input)
```
JunTaoLuo: Curious why we changed the input type?

shared/Utilities/MSBuildListSplitter.cs
```diff
@@ -0,0 +1,40 @@
...
+            foreach (var item in input)
+            {
+                var splitIdx = item.IndexOf('=');
+                if (splitIdx < 0)
+                {
+                    continue;
+                }
+
+                var key = item.Substring(0, splitIdx).Trim();
+                if (string.IsNullOrWhiteSpace(key))
```
JunTaoLuo: Nit: After the `Trim` above, `IsNullOrEmpty` should be sufficient.

modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        internal string Replace(string template, IDictionary<string, string> values)
+        {
+            var sb = new StringBuilder();
+            var line = 1;
+            for (var i = 0; i < template.Length; i++)
+            {
+                var ch = template[i];
+                if (ch == '\n')
```
JunTaoLuo: The logic here is a little convoluted, any chance we can simplify this with a case statement instead? Also is the reason we are iterating through the template instead of doing a simple `.Replace` that we want to produce useful warning messages?

modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
```diff
@@ -0,0 +1,179 @@
...
+                        }
+                        else
+                        {
+                            Log.LogWarning(null, null, null, TemplateFile,
+                                line, 0, 0, 0,
+                                message: "No property value is available for '{0}'",
+                                messageArgs: new[] { varName });
+                        }
+
+                        varNameSb.Clear();
```
JunTaoLuo: No need to clear if you aren't reusing the SB? For perf you should reuse but I'm not sure how much of a concern it is here.

modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
```diff
@@ -0,0 +1,179 @@
...
+                        var varName = varNameSb.ToString();
+                        if (values.TryGetValue(varName, out var value))
+                        {
+                            sb.Append(value);
+                        }
+                        else
+                        {
+                            Log.LogWarning(null, null, null, TemplateFile,
+                                line, 0, 0, 0,
+                                message: "No property value is available for '{0}'",
```
JunTaoLuo: string interpolate instead

modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
natemcmaster: Calling .Replace would be ok if we only ever had one variable. Calling it with multiple variables would iterate over the document multiple times

shared/Utilities/MSBuildListSplitter.cs
natemcmaster: I learned that MSBuild can bind to `string[]`, providing automatic splitting on `;`. This is necessary if you want to pass in multiple strings, some which may contain semicolons. For example, `$(RestoreSources)` has semicolons, so in Scaffolding I can pass this in as `RestoreSources=$([MSBuild]::Escape($(RestoreSources.Trim())));`

natemcmaster:
:up: :date:. Cleaned up the parsing loop a bit.

modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
JunTaoLuo: Ah so it's for perf reasons.

modules/BuildTools.Tasks/GenerateFileFromTemplate.cs
```diff
@@ -0,0 +1,154 @@
...
+                {
+                    ch = template[i];
+                    if (ch != '}')
+                    {
+                        varNameSb.Append(ch);
+                    }
+                    else
+                    {
+                        // Found the end of the variable substitution
+
```
JunTaoLuo: nit: extra line

Issue #375, opened 2017-08-18
Parallelize test assemblies


natemcmaster:
dotnet-vstest supports running tests in parallel. 

Example: on the aspnet/FileSystem repo, you could execute dotnet-vstest twice, once per target framework, and run everything in parallel. This cuts test time in about half (41 seconds -> 22 seconds).

```ps1
dotnet vstest `
    /parallel `
    --framework:".NETCoreApp,Version=v2.0.0" `
    .\test\FS.Embedded.Tests\bin\Debug\netcoreapp2.0\Microsoft.Extensions.FileProviders.Embedded.Tests.dll `
    .\test\FS.Composite.Tests\bin\Debug\netcoreapp2.0\Microsoft.Extensions.FileProviders.Composite.Tests.dll `
    .\test\FS.Globbing.Tests\bin\Debug\netcoreapp2.0\Microsoft.Extensions.FileSystemGlobbing.Tests.dll `
    .\test\FS.Physical.Tests\bin\Debug\netcoreapp2.0\Microsoft.Extensions.FileProviders.Physical.Tests.dll

dotnet vstest `
    /parallel `
    --framework:".NETFramework,Version=v4.6.1" `
    --testAdapterPath:"$(Get-Location)\test\FS.Embedded.Tests\bin\Debug\net461\" `
    .\test\FS.Embedded.Tests\bin\Debug\net461\Microsoft.Extensions.FileProviders.Embedded.Tests.dll `
    .\test\FS.Composite.Tests\bin\Debug\net461\Microsoft.Extensions.FileProviders.Composite.Tests.dll `
    .\test\FS.Globbing.Tests\bin\Debug\net461\Microsoft.Extensions.FileSystemGlobbing.Tests.dll `
    .\test\FS.Physical.Tests\bin\Debug\net461\Microsoft.Extensions.FileProviders.Physical.Tests.dll
```

We just need to figure out how do to this right in MSBuild.

natemcmaster:
In the MVC repo, this approach reduced test time from 187 to 81 seconds, and all tests passed.

natemcmaster:
Added in #417

PR #417, opened 2017-10-10
Execute test assemblies in parallel

natemcmaster:
This invokes `dotnet vstest` directly instead of invoking /t:VSTest on projects.

For projects that do not want test assemblies to run in parallel, they can put projects into custom test groups like this:

```xml
<!-- MyTests.csproj-->
<PropertyGroup>
  <TestGroupName>IntegrationTests</TestGroupName>
</PropertyGroup>
```

The default value of TestGroupName is 'UnitTests'.

If you want all tests in a solution to run in serial, you could to this in test/Directory.Build.props so all test/\*/\*.csproj files pick it up.
```xml
<!-- test/Directory.Build.props -->
<PropertyGroup>
  <TestGroupName>$(MSBuildProjectName)</TestGroupName>
</PropertyGroup>
```

Part of #375

### DIFF
--- a/build/tasks/RepoTasks.csproj
+++ b/build/tasks/RepoTasks.csproj
@@ -9,10 +9,13 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\shared\Microsoft.Extensions.CommandLineUtils.Sources\Utilities\*.cs" />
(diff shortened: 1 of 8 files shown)


natemcmaster:
It's been a month since I first opened. I've updated and re-targeted to dev. I actually want to merge this now. Re-ping for review.

PR #418, opened 2017-10-11
Add benchmark validation

pakrym:
https://github.com/aspnet/Universe/issues/590

It's opt in and requires benchmarks to use https://github.com/aspnet/Common/pull/265

### DIFF
--- a/files/KoreBuild/modules/vstest/module.targets
+++ b/files/KoreBuild/modules/vstest/module.targets
@@ -62,6 +62,10 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
       Importance="High"
       Condition="'@(BenchmarksToValidate)' != ''" />
 
-    <Exec Command="dotnet run --no-build --project %(BenchmarksToValidate.Identity) -- * --validate-fast" IgnoreStandardErrorWarningFormat="true" />
+    <MSBuild Projects="%(BenchmarksToValidate.Identity)" Targets="GetTargetPath">
+      <Output TaskParameter="TargetOutputs" PropertyName="ExecPath" />
+    </MSBuild>
+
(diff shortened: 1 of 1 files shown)


files/KoreBuild/modules/vstest/module.targets
```diff
@@ -54,4 +55,13 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
...
       RemoveProperties="$(_BuildPropertiesToRemove)" />
   </Target>
+
+  <Target Name="ValidateBenchmarks" DependsOnTargets="ResolveSolutions" Condition=" '$(EnableBenchmarkValidation)' == 'true' ">
+
+    <Message Text="%0ARunning benchmark validation for:%0A@(BenchmarksToValidate -> '%(FileName)','%0A')%0A"
+      Importance="High"
+      Condition="'@(BenchmarksToValidate)' != ''" />
+
+    <Exec Command="dotnet run --no-build --project %(BenchmarksToValidate.Identity) -- * --validate-fast" IgnoreStandardErrorWarningFormat="true" />
```
natemcmaster: Does this require Configuration=Release?

pakrym: `JitOptimizationsValidator.FailOnError` is only added in non validation run.
https://github.com/aspnet/DependencyInjection/pull/594/files#diff-45f9a437086310818894446a2b7079fdR21

natemcmaster: By the way, dotnet-run has historically been problematic from build targets. A more reliable way to invoke an executable is like this:
```xml
<MSBuild Projects="Proj.csproj" Targets="GetTargetPath">
  <Output TaskParameter="TargetOutputs" PropertyName="ExecPath" />
</MSBuild>
<Exec Command="dotnet $(ExecPath)" />
```

PR #419, opened 2017-10-11
Use SourceLink

ryanbrandenburg:


src/Internal.AspNetCore.Sdk/build/Git.targets
```diff
@@ -19,6 +21,34 @@
     </Sdk_GetGitCommitInfo>
   </Target>
 
+  <Target Name="CreateSourceLink"
+          BeforeTargets="CoreCompile"
+          DependsOnTargets="GenerateCommitHashAttribute" 
+          Condition="'$(SourceLink)'=='' AND '$(GenerateSourceLinkFile)' != 'false' ">
+
+    <Exec Command="git config --get remote.origin.url"
```
natemcmaster: This invocation should be removed. We sometimes use private forks of repos, and we don't want to disclose this in build artifacts.

You should be able to use `RepositoryUrl` instead, which is pre-defined in our repos. Example: https://github.com/aspnet/Common/blob/dev/Directory.Build.props#L6

src/Internal.AspNetCore.Sdk/build/Git.targets
```diff
@@ -19,6 +21,34 @@
     </Sdk_GetGitCommitInfo>
   </Target>
 
+  <Target Name="CreateSourceLink"
+          BeforeTargets="CoreCompile"
+          DependsOnTargets="GenerateCommitHashAttribute" 
+          Condition="'$(SourceLink)'=='' AND '$(GenerateSourceLinkFile)' != 'false' ">
```
natemcmaster: Can you make this target incremental by defining its Inputs/Outputs?

modules/BuildTools.Tasks/CreateSourceLink.cs
```diff
@@ -0,0 +1,65 @@
...
+            File.WriteAllText(DestinationFile, $"{{\"documents\":{{\"{rootDirectory}\":\"{codeSource}\"}}}}");
+
+            SourceLinkFile = DestinationFile;
+
+            return true;
+        }
+
+        private string ConvertUrl()
+        {
+            string prefix;
```
natemcmaster: Nit: ternary :)
```c#
var prefix = OriginUrl.StartsWith("git@github.com:")
  ? "git@github.com:"
  : "https://github.com/";
```

modules/BuildTools.Tasks/CreateSourceLink.cs
```diff
@@ -0,0 +1,65 @@
...
+        [Required]
+        public string Commit { get; set; }
+
+        [Output]
+        public string SourceLinkFile { get; set; }
+
+        public override bool Execute()
+        {
+            var rootDirectory = Path.Combine(Path.GetFullPath(RootDirectory), "*");
+            rootDirectory = rootDirectory.Replace("\\", "\\\\");
```
natemcmaster: I would prefer we normalize to `/`. MSBuild may give you paths that are a mix.

modules/BuildTools.Tasks/CreateSourceLink.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            var codeSource = ConvertUrl();
```
natemcmaster: Add an error if the OriginUrl does not contain github.com. We only need to support github links for now, but it would be good to have this fail if we every decided to mirror to something else.

ryanbrandenburg:
:up::date:

I know we talked about merging `CreateSourceLink` into one of the existing targets in the `GenerateCommitHashAttribute` chain, but I figured adding the target to the chain clarified things while still keeping a good separation of concerns.

src/Internal.AspNetCore.Sdk/build/Git.targets
```diff
@@ -19,6 +21,29 @@
     </Sdk_GetGitCommitInfo>
   </Target>
 
+  <Target Name="CreateSourceLink"
+          Condition="'$(SourceLink)'=='' AND '$(GenerateSourceLinkFile)' != 'false' "
+          Inputs="$(IntermediateCommitHash);$(MSBuildAllProjects)"
+          Outputs="$(SourceLinkDestination)" >
+
+    <PropertyGroup>
+      <IsSourceProject>$(MSBuildProjectDirectory.StartsWith($(SourceDirectory)))</IsSourceProject>
```
natemcmaster: Can we remove this restriction, or does it cause problems? I know of at least one project that we ship that isn't in `src/`.

src/Internal.AspNetCore.Sdk/build/Git.targets
```diff
@@ -19,6 +21,29 @@
...
+          RootDirectory="$(RepositoryRoot)"
+          OriginUrl="$(RepositoryUrl)"
+          Commit="$(CommitHash)"
+          DestinationFile="$(SourceLinkDestination)"
+          ContinueOnError="WarnAndContinue"
+          Condition="'$(CommitHash)' != '' AND '$(RepositoryUrl)' != '' AND '$(IsSourceProject)' == 'True' ">
+      <Output TaskParameter="SourceLinkFile" PropertyName="SourceLink" />
+    </Sdk_CreateSourceLink>
+
+    <Warning Text="SourceLink not enabled because this is not a git repo."
```
natemcmaster: Also add a warning if `RepositoryUrl` is empty. We should be setting this everywhere sourcelink it used.

src/Internal.AspNetCore.Sdk/build/Git.targets
ryanbrandenburg: This is my attempt to prevent us from needlessly sourcelinking test/sample/etc projects. Should I blacklist the `test` dir instead of whitelisting `src`, or do we not care about spending a little extra time generating these in places they won't get used?

natemcmaster: Ok. Easier to start small first. One change though: instead of using existence under `src/` use `IsPackable`. This property is define as `true` for any csproj that produces a nupkg.

ryanbrandenburg:
:up::date:

src/Internal.AspNetCore.Sdk/build/Git.targets
natemcmaster: So, add this:
```xml
<GenerateSourceLinkFile Condition="'$(GenerateSourceLinkFile)' == '' AND '$(IsPackable)' != 'false'>true</GenerateSourceLinkFile>
```

natemcmaster:
Obivously, fix the tests first...
```
Failed   CreatesFile
Error Message:
 Assert.Equal() Failure
                            (pos 17)
Expected: {"documents":{"C:\*":"https://raw.githubusercontent.com/as···
Actual:   {"documents":{"C:/*":"https://raw.githubusercontent.com/as···
                            (pos 17)
Stack Trace:
   at BuildTools.Tasks.Tests.CreateSourceLinkTest.CreatesFile() in C:\projects\buildtools\test\BuildTools.Tasks.Tests\CreateSourceLinkTest.cs:line 41
Total tests: 20. Passed: 18. Failed: 2. Skipped: 0.
Test Run Failed.
```

natemcmaster:
Windows passes, but Linux still has issues.
```
Failed   CreatesFile
Error Message:
 Assert.Equal() Failure
                         ↓ (pos 15)
Expected: {"documents":{"C:/*":"https://raw.githubusercontent.com/···
Actual:   {"documents":{"/home/travis/build/aspnet/BuildTools/test···
                         ↑ (pos 15)
```

ryanbrandenburg:
Fixed.

natemcmaster:
:shipit:

PR #423, opened 2017-10-13
Use SourceLink

ryanbrandenburg:
Cherry-picking SourceLink back into 2.0.2 as part of aspnet/Universe#553.


natemcmaster:
un-:shipit: we need to make this change first. https://github.com/aspnet/Routing/pull/470

ryanbrandenburg:
Is it enough to make the task conditional on RepositoryRoot? I would have done so before, but I thought it was always set and wouldn't be hit.

ryanbrandenburg:
Updated with the RepositoryRoot fix.

natemcmaster:
:shipit:

PR #433, opened 2017-10-20
Separate error for null MetadataResource

ryanbrandenburg:
We ran into the case from myget and want to know if we're getting a null `MetadataResource` or if items are sometimes missing from the list.


modules/KoreBuild.Tasks/DownloadNuGetPackages.cs
```diff
@@ -167,14 +167,19 @@ public override bool Execute()
             CancellationToken cancellationToken)
         {
             var remoteLibraryProvider = new SourceRepositoryDependencyProvider(repo, logger, cacheContext, ignoreFailedSources: false, ignoreWarning: false);
-            var downloads = new List<Task<bool>>();
             var metadataResource = await repo.GetResourceAsync<MetadataResource>();
 
+            if (metadataResource == null)
+            {
+                logger.LogError($"MetadataResource for '{repo}' was null.");
```
natemcmaster: add `return false` to exit early and avoid NRE

Issue #436, opened 2017-10-23
Add KoreBuild.Console commands that shim into korebuild MSBuild targets


natemcmaster:
Just like dotnet-CLI shims dotnet-build, dotnet-restore, and others into MSBuild targets, we should do this for some of our korebuild targets to make it easier to use korebuild targets.

Candidates:
```
run.ps1 generate resx  => /t:Resx
run.ps1 generate deps  => /t:GenerateDependenciesFile

run.ps1 upgrade => updates everything, including KoreBuild
run.ps1 upgrade deps => /t:UpgradeDependencies
```

PR #438, opened 2017-10-23
Implement upgrade-deps and generate-deps CLI commands

natemcmaster:
Fixes https://github.com/aspnet/BuildTools/issues/436

I've also updated https://github.com/aspnet/BuildTools/wiki/PackageReference-management

### DIFF
--- a/tools/KoreBuild.Console/Commands/DependenciesGenerateCommand.cs
+++ b/tools/KoreBuild.Console/Commands/DependenciesGenerateCommand.cs
@@ -1,0 +1,72 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace KoreBuild.Console.Commands
+{
+    internal class DependenciesGenerateCommand : SubCommandBase
+    {
+        private CommandOption _configOpt;
+        private CommandOption _fileOpt;
+
+        public override void Configure(CommandLineApplication application)
+        {
+            application.Description = "Generates a build/dependencies.props file and updates csproj files to use variables";
+            application.ExtendedHelpText = @"
+MORE INFO:
+
+    This command will generate a dependencies.props file and adjust all PackageReference's in csproj files
+    to use the MSBuild variables it generates.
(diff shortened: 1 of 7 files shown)


tools/KoreBuild.Console/Commands/DependenciesGenerateCommand.cs
```diff
@@ -0,0 +1,72 @@
...
+    Example output:
+
+    <Project>
+      <PropertyGroup Label=""Package Versions"">
+        <MyPackageVersion>1.0.0</MyPackageVersion>
+      </PropertyGroup>
+    </Project>
+";
+
+            _configOpt = application.Option("-c|--configuration <CONFIG>", "The MSBuild configuration. Defaults to 'Debug'.", CommandOptionType.SingleValue);
```
ryanbrandenburg: I was under the impression that the first argument of Option was the descriptor, not the help string (ie it should be `"-c|--configuration"`).

tools/KoreBuild.Console/Commands/DependenciesGenerateCommand.cs
```diff
@@ -0,0 +1,72 @@
...
+            base.Configure(application);
+        }
+
+        protected override int Execute()
+        {
+            var args = new List<string>
+            {
+                "msbuild",
+                Path.Combine(KoreBuildDir, "KoreBuild.proj"),
+                "-t:GenerateDependenciesPropsFile",
```
ryanbrandenburg: I didn't know you could use -t instead of /t, that's useful!

tools/KoreBuild.Console/Commands/DependenciesGenerateCommand.cs
natemcmaster: It is.

https://github.com/aspnet/BuildTools/blob/448eca8e5df4c30f443af737cbd4c6e445cd76d5/shared/Microsoft.Extensions.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs#L74-L75

The expected format of `template` is `-$(ShortName)|--$(LongName) <$(ValueName)>`

https://github.com/aspnet/BuildTools/blob/448eca8e5df4c30f443af737cbd4c6e445cd76d5/shared/Microsoft.Extensions.CommandLineUtils.Sources/CommandLine/CommandOption.cs#L18-L46

Issue #428, opened 2017-10-16
Support using full MSBuild


natemcmaster:
Projects that need full MSBuild:

 - aspnet/Razor - has a VSIX project
 - aspnet/DotNettools - will have VSIX project soon
 - aspnet/AspNetCoreModule - for native cpp projects cc @jkotalik 

Ideas:

 - We already support a korebuild.json file. We could add more options there.

```json
{
   "toolsets": {
      "visualstudio": { 
          "required": true,
          "minVersion": "15.3.0",
          "installFile": "./visualstudio.json"
       }
    }
}
```
 - Visual Studio supports defining installation parameters as a file. https://docs.microsoft.com/en-us/visualstudio/install/automated-installation-with-response-file

Example:
```
vs_enterprise.exe --in visualstudio.json
```
```json
{
  "installChannelUri": ".\\ChannelManifest.json",
  "channelUri": "https://aka.ms/vs/15/release/channel",
  "installCatalogUri": ".\\Catalog.json",
  "channelId": "VisualStudio.15.Release",
  "productId": "Microsoft.VisualStudio.Product.Enterprise",

  "installPath": "C:\\VS2017",
  "quiet": false,
  "passive": false,
  "includeRecommended": true,
  "norestart": false,

  "addProductLang": [
    "en-US",
    "fr-FR"
    ],

    "add": [
        "Microsoft.VisualStudio.Workload.ManagedDesktop",
        "Microsoft.VisualStudio.Workload.Data",
        "Microsoft.VisualStudio.Workload.NativeDesktop",
        "Microsoft.VisualStudio.Workload.NetWeb",
        "Microsoft.VisualStudio.Workload.Office",
        "Microsoft.VisualStudio.Workload.Universal",
        "Component.GitHub.VisualStudio"
    ]
}
```

PR #443, opened 2017-10-27
Add support for requiring Visual Studio's MSBuild from KoreBuild

natemcmaster:
Resolves https://github.com/aspnet/BuildTools/issues/428

Add's support for requiring VS's MSBuild in order to build a repo. When VS is installed, the MSBuild properties `VisualStudioMSBuildx86Path` and `VisualStudioMSBuildx64Path` will be set to the path of MSBuild.exe. You can use this instead of globbing over C:\Program Files (x86)\Microsoft Visual Studio\ to find 'full' MSBuild.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -12,6 +12,7 @@
     <SystemReflectionMetadataVersion>1.5.0</SystemReflectionMetadataVersion>
     <TestSdkVersion>15.3.0</TestSdkVersion>
     <XunitVersion>2.3.0-rc3-build3818</XunitVersion>
+    <VSWherePackageVersion>2.2.7</VSWherePackageVersion>
   </PropertyGroup>
 
   <!--
--- a/modules/KoreBuild.Tasks/GetToolsets.cs
+++ b/modules/KoreBuild.Tasks/GetToolsets.cs
@@ -1,0 +1,97 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Runtime.InteropServices;
+using KoreBuild.Tasks.Utilities;
+using Microsoft.Build.Framework;
+
+namespace KoreBuild.Tasks
+{
+    /// <summary>
+    /// Finds toolset information as listed in korebuild.json
+    /// </summary>
+    public class GetToolsets : Microsoft.Build.Utilities.Task
+    {
+        /// <summary>
+        /// The path to the korebuild.json file.
+        /// </summary>
+        [Required]
+        public string ConfigFile { get; set; }
+
+        /// <summary>
+        /// The path to MSBuild.exe (x86), if the 'visualstudio' toolset was specified. It will be empty if this file does not exist.
+        /// </summary>
+        [Output]
+        public string VisualStudioMSBuildx86Path { get; set; }
+
+        /// <summary>
+        /// The path to MSBuild.exe (x64), if the 'visualstudio' toolset was specified. It will be empty if this file does not exist.
+        /// </summary>
+        [Output]
+        public string VisualStudioMSBuildx64Path { get; set; }
+
+        public override bool Execute()
+        {
+            if (!File.Exists(ConfigFile))
+            {
+                Log.LogError($"Could not load the korebuild config file from '{ConfigFile}'");
+                return false;
+            }
+
+            var settings = KoreBuildSettings.Load(ConfigFile);
+
+            if (settings?.Toolsets == null)
+            {
+                Log.LogMessage(MessageImportance.Normal, "No recognized toolsets specified.");
+                return true;
+            }
+
+            foreach (var toolset in settings.Toolsets)
+            {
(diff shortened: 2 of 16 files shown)


build/dependencies.props
```diff
@@ -12,6 +12,7 @@
     <SystemReflectionMetadataVersion>1.5.0</SystemReflectionMetadataVersion>
     <TestSdkVersion>15.3.0</TestSdkVersion>
     <XunitVersion>2.3.0-rc3-build3818</XunitVersion>
+    <VSWherePackageVersion>2.2.7</VSWherePackageVersion>
```
jkotalik: Will VS where become a dependency in Universe?

modules/KoreBuild.Tasks/GetToolsets.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            return !Log.HasLoggedErrors;
+        }
+
+        private void GetVisualStudio(KoreBuildSettings.VisualStudioToolset vsToolset)
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                Log.LogMessage(MessageImportance.Low, "Skipping Visual Studio verification on non-Windows platforms.");
```
jkotalik: Wouldn't we want this to be an error? If a repo requires to be built on windows, we shouldn't skip that check.

modules/KoreBuild.Tasks/Utilities/VsWhere.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            if (process.ExitCode != 0)
+            {
+                log.LogMessage(MessageImportance.Low, "vswhere output = " + output);
+                log.LogError("vswhere failed.");
+                return Array.Empty<VsInstallation>();
+            }
+
+            return JsonConvert.DeserializeObject<VsInstallation[]>(output);
+
```
jkotalik: nit new line.

modules/KoreBuild.Tasks/Utilities/VsWhere.cs
```diff
@@ -0,0 +1,117 @@
...
+            };
+            if (includePrerelease)
+            {
+                args.Add("-prerelease");
+            }
+
+            return GetInstallations(args, log).FirstOrDefault();
+        }
+
+        public static VsInstallation FindMostCompatibleInstallation(KoreBuildSettings.VisualStudioToolset toolset, TaskLoggingHelper log)
```
jkotalik: The method seems to be finding the first compatible Installation. Rename to "FindFirstCompatibleInstallation"?

test/KoreBuild.Tasks.Tests/VSWhereTests.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (!result)
+                {
+                    // Don't fail the test. aspnet/BuildTools might be building without a version of VS installed.
+                    // We mostly want to make sure this task doesn't throw exceptions and fails gracefully
+                    Assert.NotEmpty(engine.Errors);
+                }
+                else
+                {
+                    Assert.NotEmpty(task.InstallationBasePath);
+                    Assert.True(Directory.Exists(task.InstallationBasePath), "VS installation was not found");
```
jkotalik: Is it guaranteed that the that the installation base path will not be null? I'd assume so because we are requiring the nuget package.

tools/KoreBuildSettings.cs
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
jkotalik: Was korebuild.json not being read before?

jkotalik:
Also seems to be some exceptions in the Travis agent due to OS checks 😄

tools/KoreBuildSettings.cs
natemcmaster: It was read by the bootstrapper only. This is the first use case for reading korebuild.json from KoreBuild. I expect there will be more.

test/KoreBuild.Tasks.Tests/VSWhereTests.cs
natemcmaster: The installation base path should be non-null if the task succeeds. The task can fail if VS is not installed. vswhere is a standaone console tool that can be installed without installing vs.

build/dependencies.props
natemcmaster: Nope. It's only a build-tool dependency, not a framework dependency.

modules/KoreBuild.Tasks/Utilities/VsWhere.cs
natemcmaster: 'FindLatestCompatibleInstallation' would be better. It finds the newest VS that meets the given requirements. "first" is a little ambiguous and implies there is some kind of sorting on name or version.

modules/KoreBuild.Tasks/GetToolsets.cs
natemcmaster: Yup, good catch.

Issue #398, opened 2017-09-12
Add a warning for floating PackageReference versions


natemcmaster:
PackageReference floating versions cause unrepeatable builds and should be avoided as much as possible. Let's add a warning so we can find and remove them.

```
warning KRBXYZ: Floating package reference versions found in the following projects. This will cause non-deterministic build results.
   A.csproj
      - Pkg1/2.0.0-*
   B.csproj
       - Pkg1/1.0.0-*
       - Pkg2/1.0.0-*
```

PR #455, opened 2017-11-03
Remove PackageLineup

### LINKED ISSUE #384: Store package lineup data in a custom format
The initial implementation of PackageLineup uses NuGet's `<dependencies>` metadata to express versions. This design may limit us in some ways from expression PackageLineup rules.

For example:
 - https://github.com/aspnet/BuildTools/pull/382#discussion_r135300120. Dependencies allow `<group targetFramework="xyz">`, but you can only specify one framework in the group. What this means can be open to interpretation. i.e. only projects that exactly match? all compatible frameworks? only frameworks within a given range? Nuspec's metadata does not allow us to add more info to the lineup package to determine this behavior.

cc @pranavkm

### LINKED ISSUE #392: KRB4002 can't be suppressed if a <PackageReference> needs more than one NoWarn
This line of code: https://github.com/aspnet/BuildTools/blob/07229f1322a50424c2830032e0f9c1786cb0e865/modules/KoreBuild.Tasks/Policy.VersionRestrictions.targets#L14

... is checking whether the `NoWarn` value exactly equals `KRB4002`. This is OK if that's the only warning suppression you want to use. But if you're trying to suppress `KRB4002` *and* something else, it doesn't appear there's any way to do that.

Could the `WithMetadataValue` check be replaced by a substring search, or a search through the value split on some separator (space, comma, or semicolon - whatever matches convention best)?

### LINKED ISSUE #414: Support F# projects in new lineups system
Currently, if you have an `fsproj` referenced via `<ProjectReference>` from some `csproj`, then running `build /t:Restore` fails with:

```
error KRB1001: Package references have not been pinned. Run './build.cmd /t:Pin'. [some\path\FSharpWebSite.fsproj]
```

At a guess, maybe some of the code in buildtools works by looking for `csproj` files explicitly rather than allowing arbitrary project types ([example](https://github.com/aspnet/BuildTools/blob/2fdcbd3f8c1461caed668ebeaf42c9545d30fcbb/files/KoreBuild/modules/solutionbuild/module.targets#L9)).

It would be great if we were able to support `.fsproj` because in the past we had functional tests covering ASP.NET apps written in F#. These had to be disabled because of build system issues, but it would be ideal to bring them back.

natemcmaster:
This feature was experimental. Although it was good, it had some limitations. Ultimately, we decided using MSBuild variables and the 'dependencies.props' pattern was simpler, more flexible, and easier for our developers to understand.

A few other minor changes:
 - FolderLineup was only used internally. It's been replaced by MSBuild properties for 'BundledPackages'
 - Added a global property KoreBuildVersion. Was required to make tests pass, and could be generally useful in projects
 - A few other cleanups
 - Begin to enforce PackageReference + dependencies.props pattern by default. This can be disabled.

As a result, these bugs don't apply anymore:
Closes #392 
Closes #414 
Closes #384 

Incidentally, this fixes https://github.com/aspnet/BuildTools/issues/398


files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -68,7 +68,8 @@ function Invoke-RepositoryBuild(
         $sdkVersion = __get_dotnet_sdk_version
         if ($sdkVersion -ne 'latest') {
             "{ `"sdk`": { `"version`": `"$sdkVersion`" } }" | Out-File (Join-Path $Path 'global.json') -Encoding ascii
-        } else {
+        }
```
ryanbrandenburg: Do you have an extention for VSCode that formats PS files for you or do you just notice these yourself?

modules/KoreBuild.Tasks/CheckPackageReferences.cs
```diff
@@ -35,14 +35,20 @@ public class CheckPackageReferences : Microsoft.Build.Utilities.Task
 
         public override bool Execute()
         {
+            if (Projects == null || Projects.Length == 0)
```
ryanbrandenburg: Should we maybe do a `Log.LogVerbose` or something here? Just something to say that it succeeded because it didn't have any projects?

modules/KoreBuild.Tasks/Internal/ProjectModel/ProjectInfo.cs
```diff
@@ -44,7 +41,5 @@ internal class ProjectInfo
 
         public IReadOnlyList<ProjectFrameworkInfo> Frameworks { get; }
         public IReadOnlyList<DotNetCliReferenceInfo> Tools { get; }
-
-        public MSBuildProjectExtension TargetsExtension { get; }
```
ryanbrandenburg: I guess we just didn't use this anywhere?

modules/KoreBuild.Tasks/Utilities/DependencyVersionsFile.cs
```diff
@@ -75,16 +73,15 @@ public static DependencyVersionsFile Create(bool addOverrideImport)
...
-        public static bool TryLoad(string sourceFile, TaskLoggingHelper Log, out DependencyVersionsFile file)
+        public static bool TryLoad(string sourceFile, out DependencyVersionsFile file)
         {
             try
             {
                 file = Load(sourceFile);
                 return true;
             }
-            catch (InvalidProjectFileException ex)
+            catch
```
ryanbrandenburg: Why remove the Log here? Seems worrying to catch everything, not log, then continue.

natemcmaster: The caller should take care of logging when the project fails to load.

modules/KoreBuild.Tasks/Internal/ProjectModel/ProjectInfo.cs
natemcmaster: Correct. It was only used by lineups.

files/KoreBuild/scripts/KoreBuild.psm1
natemcmaster: VS Code formats this for me. https://marketplace.visualstudio.com/items?itemName=ms-vscode.PowerShell

modules/KoreBuild.Tasks/CheckPackageReferences.cs
natemcmaster: updated

PR #464, opened 2017-11-10
Always generate a global.json

### LINKED ISSUE #462: run.ps1 should always generate global.json
global.json may be required for some of the commands KoreBuild.Console executes. Currently only `default-build` generates the global.json file

cc @ryanbrandenburg

ryanbrandenburg:
Fixes #462.


ryanbrandenburg:
Coherence final is the last non-obsolete repo that uses `Invoke-RepositoryBuild` directly. Are we keeping Coherence-Final? What does it do these days?

natemcmaster:
coherence-final is the 1.x package cache stuff.

see #465

ryanbrandenburg:
Cool, I'll wait to merge this one until after #465.

PR #477, opened 2017-11-21
Support publishing assets and skipping tests for orchestrated build

natemcmaster:
- Adds support for `/p:SkipTests=true` to skip tests
- Adds support for `/t:Publish /p:PublishBlobFeedUrl=(someurl) /p:PublishBlobFeedKey=(key)`

### DIFF
--- a/modules/Publish/module.targets
+++ b/modules/Publish/module.targets
@@ -1,0 +1,36 @@
+<Project>
+  <PropertyGroup>
+    <PublishDependsOn>$(PublishDependsOn);PublishToAzure</PublishDependsOn>
+
+    <PushToBlobFeed_Overwrite Condition="'$(PushToBlobFeed_Overwrite)' == ''">false</PushToBlobFeed_Overwrite>
+    <PushToBlobFeed_MaxClients Condition="'$(PushToBlobFeed_MaxClients)' == ''">8</PushToBlobFeed_MaxClients>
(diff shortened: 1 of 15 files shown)


modules/Publish/module.targets
```diff
@@ -0,0 +1,36 @@
...
+
+  <Target Name="PublishToAzure" DependsOnTargets="GetArtifactInfo" Condition="'$(PublishBlobFeedUrl)' != ''">
+    <ItemGroup>
+      <PackagesToPublish Include="%(ArtifactInfo.Identity)" Condition="'%(ArtifactInfo.ArtifactType)' == 'NuGetPackage'" />
+      <FilesToPublish Include="%(ArtifactInfo.Identity)" Condition="'%(ArtifactInfo.AzureAssetsPath)' != ''">
+        <RelativeBlobPath>$(FileRelativePathBase)%(ArtifactInfo.AzureAssetsPath)</RelativeBlobPath>
+      </FilesToPublish>
+    </ItemGroup>
+
+    <PushToBlobFeed ExpectedFeedUrl="$(PublishBlobFeedUrl)"
```
ryanbrandenburg: Could we get a little comment in here saying what these two Push tasks are trying to accomplish? It's not super clear to me just from the parameters used.

natemcmaster: Yup, I added a comments pointing to https://github.com/dotnet/buildtools/tree/master/src/Microsoft.DotNet.Build.Tasks.Feed for more details.

PR #480, opened 2017-11-21
Update external dependencies

ryanbrandenburg:
This task would basically only be used by Universe so far, but by putting it here we get to take advantage of a lot of existing code to handle project-type files.

### DIFF
--- a/modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
+++ b/modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
@@ -1,0 +1,57 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using KoreBuild.Tasks.Utilities;
+using Microsoft.Build.Framework;
+using System.Collections.Generic;
+
+namespace KoreBuild.Tasks
+{
+    public class UpgradeExternalDependencies : Microsoft.Build.Utilities.Task
+    {
+
+        [Required]
+        public string UpdateSource { get; set; }
+
+        /// <summary>
+        /// The dependencies.props file to update
+        /// </summary>
+        [Required]
+        public string DependenciesFile { get; set; }
+
+        public override bool Execute()
+        {
+            var latest = GetLatestDeps();
+
+            return UpdateDependencyFile(latest);
+        }
+
+        private IReadOnlyDictionary<string, string> GetLatestDeps()
+        {
+            return DependencyVersionsFile.Load(UpdateSource).VersionVariables;
+        }
+
+        private bool UpdateDependencyFile(IReadOnlyDictionary<string, string> latest)
+        {
+            if (!DependencyVersionsFile.TryLoad(DependenciesFile, out var localVersionsFile))
+            {
+                Log.LogError($"Could not load file from {DependenciesFile}");
+                return false;
+            }
+
+            foreach (var sourceVariable in latest)
+            {
+                if (!localVersionsFile.VersionVariables.ContainsKey(sourceVariable.Key))
+                {
+                    Log.LogWarning($"Creating variable {sourceVariable.Key}");
+                }
+                Log.LogWarning($"Setting '{sourceVariable.Key}' to '{sourceVariable.Value}'",);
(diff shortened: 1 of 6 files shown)


modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using KoreBuild.Tasks.Utilities;
+using Microsoft.Build.Construction;
+using Microsoft.Build.Framework;
+using System.Collections.Generic;
```
natemcmaster: Nit:  we typically [sort System usings](http://dailydotnettips.com/2015/01/06/automatically-place-the-system-directives-first-when-sorting-usings-option-in-visual-studio/) before others.

modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Linq;
+
+namespace KoreBuild.Tasks
+{
+    public class UpgradeExternalDependencies : Microsoft.Build.Utilities.Task
```
natemcmaster: Add a description about what this task does

modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        [Required]
```
natemcmaster: Nit: docs

modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
...
+            var updateCount = 0;
+
+            foreach (var property in latest)
+            {
+                if (!localVersionsFile.VersionVariables.ContainsKey(property.Name))
+                {
+                    Log.LogWarning($"Creating variable {property.Name}");
+                }
+
+                if(localVersionsFile.VersionVariables[property.Name] != property.Value)
```
natemcmaster: This will throw KeyNotFoundException

modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    Log.LogWarning($"Setting '{property.Name}' to '{property.Value}'");
```
natemcmaster: Why warning? Isn't this what we want?

modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
...
+                return false;
+            }
+
+            var updateCount = 0;
+
+            foreach (var property in latest)
+            {
+                if (!localVersionsFile.VersionVariables.ContainsKey(property.Name))
+                {
+                    Log.LogWarning($"Creating variable {property.Name}");
```
natemcmaster: Why are we creating new variables? Should we just change existing ones?

modules/KoreBuild.Tasks/UpgradeExternalDependencies.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.Build.Framework;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace KoreBuild.Tasks
+{
+    public class UpgradeExternalDependencies : Microsoft.Build.Utilities.Task
+    {
+        [Required]
+        public string UpdateSource { get; set; }
```
natemcmaster: Naming: PackageVersionsFilePath might be better.

natemcmaster:
It seems like this has a long of redundance with the UpgradeDependencies task. You could probably abstract it.

Also, we need to figure out first what our inputs are to upgrading our external repos. Are we pulling from dotnet/versions? Myget? a package? etc. I'm not currently sure about how we can automate this in a reliable way.

ryanbrandenburg:
The manifest this will target is not a well known quantity yet, so I'm closing this PR for now.

PR #482, opened 2017-11-22
Add support for overriding package versions with DotNetPackageVersionPropsPath

natemcmaster:
Part of orchestrated build. Adds an import to DotNetPackageVersionPropsPath to allow the versions of packages used in build tools to be specified via orchestrated build.


natemcmaster:
On TC: I'll add a pre-build step that downloads PB_PackageVersionPropsUrl and builds the repo with it.

PR #483, opened 2017-11-22
Allow BlobStorage in bootstrappers

ryanbrandenburg:
The addition of a ToolsSourceSuffix lets us append a query string, allowing auth and access to blob storage.

### DIFF
--- a/scripts/bootstrapper/run.sh
+++ b/scripts/bootstrapper/run.sh
@@ -17,6 +17,7 @@ update=false
 repo_path="$DIR"
 channel=''
 tools_source=''
+tools_source_suffix=''
 
 #
 # Functions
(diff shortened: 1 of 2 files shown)


natemcmaster:
Syntax error:
```
 /tmp/korebuild/07c1e2a4-c723-4590-ac57-cf03bf03cfc2/run.sh: line 172: syntax error near unexpected token `)'
```

scripts/bootstrapper/run.sh
```diff
@@ -106,14 +107,14 @@ __get_remote_file() {
 
     local failed=false
     if __machine_has wget; then
-        wget --tries 10 --quiet -O "$local_path" "$remote_path" || failed=true
+        wget --tries 10 --quiet -O "$local_path" "$remote_path"+"$remote_path_suffix" || failed=true
```
natemcmaster: Not valid bash. Do this instead.
```sh
"${remote_path}${remote_path_suffix}"
```

scripts/bootstrapper/run.sh
```diff
@@ -164,6 +165,10 @@ while [[ $# -gt 0 ]]; do
             tools_source="${1:-}"
             [ -z "$tools_source" ] && __usage
             ;;
+        --tools-source-suffix|-ToolsSourceSuffix)
+            shift
+            tools_source_suffix="${1:-}"
```
natemcmaster: Set a default value for this variable at the top of the file

scripts/bootstrapper/run.sh
```diff
@@ -164,6 +165,10 @@ while [[ $# -gt 0 ]]; do
             tools_source="${1:-}"
             [ -z "$tools_source" ] && __usage
             ;;
+        --tools-source-suffix|-ToolsSourceSuffix)
```
natemcmaster: Update the __usage as well

scripts/bootstrapper/run.sh
```diff
@@ -164,6 +167,10 @@ while [[ $# -gt 0 ]]; do
...
+            shift
+            tools_source_suffix="${1:-}"
+            [ -z "$tools_source_suffix" ] && __usage
```
natemcmaster: Add a line with `;;` after this to fix the syntax error

> /tmp/korebuild/98b6cb1f-b9a1-4a37-a2c9-25b6f5cb7ffb/run.sh: line 174: syntax error near unexpected token `)'

ryanbrandenburg:
:up::date: although it looks like @natemcmaster  folded this into #489?

natemcmaster:
Not exactly. #489 is just for the build.ps1 to build aspnet/BuildTools...not the generic bootstrapper

PR #486, opened 2017-11-22
Add support for excluding rules

pranavkm:
Usage: https://github.com/aspnet/Coherence-Signed/pull/705


modules/NuGetPackageVerifier/console/CompositeRules/NonAdxVerificationCompositeRule.cs
```diff
@@ -1,25 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Collections.Generic;
-
-namespace NuGetPackageVerifier.Rules
-{
-    public class NonAdxVerificationCompositeRule : IPackageVerifierRule
```
pranavkm: We don't use this rule anymore.

PR #490, opened 2017-11-28
Remove the third-party package version rule

### LINKED ISSUE #265: NPV: Specifying exclusions with wildcard doesn't work for ThirdPartyDependencyRule
For instance,
This works,
```JSON
"Microsoft.AspNetCore.All": {
  "Exclusions": {
    "THIRDPARTY_DEPENDENCY_NOT_REGISTERED": {
      "Microsoft.AspNetCore.All; .NETCoreApp,Version=v2.0": "Reason"
    }
  }
}
```

This doesn't work,
```JSON
"Microsoft.AspNetCore.All": {
  "Exclusions": {
    "THIRDPARTY_DEPENDENCY_NOT_REGISTERED": {
      "*": "Reason"
    }
  }
}
```

Specifying wildcard works in other cases like here https://github.com/aspnet/Coherence-Signed/blob/dev/tools/NugetPackageVerifier.json#L27

### LINKED ISSUE #266: ThirdParty.json shouldn't be tied to NuGetPackageVerifier

natemcmaster:
This check is redundant with checks that exist in Universe. Third-party package validation is now checked in aspnet/Universe by checking this list: https://github.com/aspnet/Universe/blob/dev/build/dependencies.props and this task: https://github.com/aspnet/Universe/blob/dev/build/tasks/VerifyCoherentVersions.cs#L100-L112

Resolves #266.
Resolves #265.


ajaybhargavb:
👏

PR #497, opened 2017-11-29
Add support for installing multiple runtimes

pranavkm:


files/KoreBuild/scripts/get-dotnet.sh
```diff
@@ -65,7 +65,7 @@ fi
 channel='preview'
 version=$(__get_dotnet_sdk_version)
 runtime_channel='master'
-runtime_version=$(< "$__script_dir/../config/runtime.version" head -1 | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+readarray runtime_versions < "$__script_dir/../config/runtime.versions"
```
pranavkm: `readarray` requires bash v4 or higher. Just checking in case this will be an issue.

files/KoreBuild/scripts/get-dotnet.sh
```diff
@@ -79,16 +79,18 @@ if [ ! -z "${KOREBUILD_DOTNET_SHARED_RUNTIME_CHANNEL:-}" ]; then
 fi
 
 if [ ! -z "${KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION:-}" ]; then
```
pranavkm: Do we still use these switches anywhere?

files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -197,12 +197,14 @@ function Install-Tools(
         Write-Warning "dotnet shared runtime channel overridden by KOREBUILD_DOTNET_SHARED_RUNTIME_CHANNEL"
     }
     if ($env:KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION) {
-        $runtimeVersion = $env:KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION
+        $runtimeVersions = $env:KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION
```
natemcmaster: Environment variables are always a single string. We should probably split on some separator, such as
`$env:KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION -split ';'`

pranavkm: Forgot to click submit on the my comments earlier. Do we still use these switches? If not, might be nicer to just get rid of these entirely. I'd certainly not want to support adding more features (multiple versions for instance) to it though.

files/KoreBuild/scripts/get-dotnet.sh
natemcmaster: The default version of bash on macOS is 3.2 :-/

files/KoreBuild/scripts/KoreBuild.psm1
natemcmaster: I'd love to get rid of these altogether. 👍 We do have a few edge cases where we need them, but otherwise I'm in favor of dropping these.

pranavkm:
Closing in favor of alternative approach.

PR #498, opened 2017-11-29
Skip installing unversioned files when installing shared runtime.

pranavkm:


files/KoreBuild/scripts/dotnet-install.sh
```diff
@@ -790,7 +790,6 @@ do
             runtime_id="$1"
             ;;
         --skip-non-versioned-files|-[Ss]kip[Nn]on[Vv]ersioned[Ff]iles)
-            shift
```
natemcmaster: 😬 oops. Looks like dotnet/cli has this bug, too.

PR #499, opened 2017-11-29
Add DownloadFile task

natemcmaster:
Required so we can download orchestrated build artifacts from the blob feed in universe and other places.

### DIFF
--- a/test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
+++ b/test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
@@ -1,0 +1,81 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using BuildTools.Tasks.Tests;
+using Xunit;
+using Xunit.Abstractions;
+using Task = System.Threading.Tasks.Task;
+
+namespace KoreBuild.Tasks.Tests
+{
+    public class DownloadFileTests
+    {
+        private readonly ITestOutputHelper _output;
+
+        public DownloadFileTests(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [Fact]
+        public async Task ItDownloadAFile()
+        {
+            var expectedPath = Path.Combine(AppContext.BaseDirectory, "microsoft.com.2.html");
+            var task = new DownloadFile
+            {
+                Uri = "http://example.org/index.html",
+                DestinationPath = expectedPath,
+                BuildEngine = new MockEngine(_output),
+            };
+
+            if (File.Exists(expectedPath))
+            {
+                File.Delete(expectedPath);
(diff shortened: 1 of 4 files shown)


test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                DestinationPath = expectedPath,
+                BuildEngine = new MockEngine(_output),
+            };
+
+            if (File.Exists(expectedPath))
+            {
+                File.Delete(expectedPath);
+            }
+
+            Assert.False(File.Exists(expectedPath), "The file should not exist yet");
```
pranavkm: Shouldn't need this - `File.Delete` would have thrown

test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            Assert.True(File.Exists(expectedPath), "The file should exist");
+        }
+
+        [Fact]
+        public async Task ItFailsForFilesThatDoNotExist()
+        {
+            var engine = new MockEngine(_output) { ContinueOnError = true };
+            var task = new DownloadFile
+            {
+                Uri = "http://localhost/this/file/does/not/exist",
```
pranavkm: Could we add a timeout so this doesn't wait forever if there's a server responding locally?

test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var packages = new[]
+            {
+                new TaskItem("Newtonsoft.Json", new Hashtable { ["Version"] = "9.0.1", ["Source"] = "https://api.nuget.org/v3/index.json"} ),
+            };
+
+            var expectedPath = Path.Combine(AppContext.BaseDirectory, "microsoft.com.html");
+            var task = new DownloadFile
+            {
+                Uri = "https://microsoft.com/",
```
pranavkm: `http://example.org/index.html` (I'd rather we not use TLS here to avoid flakiness due to server cert expiration)

test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public DownloadTests(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [Fact]
+        public async Task ItDownloadAFile()
+        {
+            var packages = new[]
```
pranavkm: Remove?

test/KoreBuild.Tasks.Tests/DownloadFileTests.cs
natemcmaster: Added a timeout that defaults to 2 minutes to start the response, 15 minutes to finish the response.

Issue #485, opened 2017-11-22
API Check does not understand `TypeForwardedTo`


dougbu:
This issue not only mandates redundant breaking change descriptions but also means we have no way to check whether the type forwarding destination remains compatible with the original type.

For example, the MVC ApiExplorer assembly type forwards eight types to the Abstractions assembly in the project's [AssemblyInfo.cs file](https://github.com/aspnet/Mvc/blob/008a562bf5b1ab89e4b83c60faaabd1d12ffcb82/src/Microsoft.AspNetCore.Mvc.ApiExplorer/Properties/AssemblyInfo.cs#L6-L13).

The commit that did this (https://github.com/aspnet/Mvc/commit/07c22f2b29a684cd3371d792e77cbb55523f0f1f) had to **also** include breaking change descriptions for the same types. The updated descriptions were last found in the project's [breakingchanges.netcore.json file](https://github.com/aspnet/Mvc/blob/38712609bb2291202bfa760ee62348691de0ddae/src/Microsoft.AspNetCore.Mvc.ApiExplorer/breakingchanges.netcore.json#L2-L33). (The breaking changes went away in https://github.com/aspnet/Mvc/commit/a8eda83222711a4ee63307d1c71d184a92359351#diff-fbf88c3372073b371a0299db7fd4c676, when the API Check files were reset for the 2.0.0 release.)

dougbu:
Self-assigning because I'll soon need to create additional type forwards.

dougbu:
Information needed to fix this bug is not available on .NET Framework. Options include:
1. Fix bug only on .NET Core App 2.1, using the `Assembly.GetForwardedTypes()` method (added in 2.1). This could result in odd inconsistencies between baseline and breaking change listings for cross-compiled assemblies. But, that downside may be irrelevant because Asp.NET Core contains very few `[TypeForwardedTo]` attributes, they are all in one assembly (Microsoft.AspNetCore.Mvc.ApiExplorer.dll), and that assembly targets only .NET Standard 2.0. Of course, inconsistencies could come up in the future&hellip;
2. Use System.Reflection.Metadata to find forwarded types on all platforms. The necessary APIs appear to exist there. (See [this question](https://github.com/dotnet/corefx/issues/19713#issuecomment-301137320) and [answer](https://github.com/dotnet/corefx/issues/19713#issuecomment-301137784) in dotnet/corefx#19713.)
[...]

PR #500, opened 2017-11-29
Update APICheck to understand `[TypeForwardedTo]`

dougbu:
- #485


dougbu:
@Eilon the types we're currently forwarding are listed in the src/Microsoft.AspNetCore.Mvc.Abstractions/baseline.netcore.json file. Updating src/Microsoft.AspNetCore.Mvc.ApiExplorer/baseline.netcore.json after this goes in would catch someone removing the `[TypeForwardedTo]` attributes. But, that's probably the only value of such a follow-on. Worth doing?

dougbu:
ef83157a1e

PR #501, opened 2017-11-29
StrongNames Targets

ryanbrandenburg:
Adding targets so that we can enable and disable StrongName checking if we have a problem with it in the future.

### DIFF
--- a/modules/KoreBuild.Tasks/DisableSkipStrongName.cs
+++ b/modules/KoreBuild.Tasks/DisableSkipStrongName.cs
@@ -1,0 +1,70 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using KoreBuild.Tasks.SkipStrongNames;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.Win32;
+
+namespace KoreBuild.Tasks
+{
+    public class DisableSkipStrongName : Task
+    {
+        [Required]
+        public string XmlFile { get; set; }
+
+        public override bool Execute()
+        {
+            var configuration = new StrongNameConfiguration(AssembliesFile.Read(XmlFile));
+
+            return SkipDisable(configuration);
+        }
+
+        private bool SkipDisable(StrongNameConfiguration configuration)
+        {
+            bool printedHeader = false;
+
+            foreach (RegistrySection section in WindowsRegistry.Sections)
+            {
+                using (RegistryKey registryKey = WindowsRegistry.CreateWritableVerificationRegistryKey(section))
+                {
+                    if (registryKey == null)
+                    {
+                        Log.LogError($"Unable to open writable verification registry key for {section}.");
+                        return false;
+                    }
+
+                    foreach (var assembly in configuration.FilteredAssemblySpecifications[section])
+                    {
+                        RegistryKey subKey = registryKey.OpenSubKey(assembly);
+
+                        if (subKey != null)
+                        {
+                            if (!printedHeader)
+                            {
+                                printedHeader = true;
+                                Log.LogMessage("Deleting registry entries:");
+                            }
+
+                            Log.LogMessage($"  {registryKey}\\{assembly}");
+                            subKey.Dispose();
+                            registryKey.DeleteSubKeyTree(assembly);
+                        }
+                    }
+                }
+            }
+
+            if (!printedHeader)
+            {
+                Log.LogMessage("Skip Strong Names is already disabled.");
+            }
+            else
+            {
+                Log.LogMessage("");
+                Log.LogMessage("Skip Strong Names was successfully disabled.");
+            }
+
+            return true;
+        }
+    }
+}
--- a/modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
+++ b/modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
@@ -12,6 +12,7 @@
     <Compile Include="..\..\shared\Utilities\MSBuildListSplitter.cs" />
     <Compile Include="..\..\tools\KoreBuildSettings.cs" />
     <Content Include="$(VSWhereDir)vswhere.exe" CopyToOutputDirectory="PreserveNewest" CopyToPublishDirectory="PreserveNewest" />
+    <Content Include="$(MSBuildThisFileDirectory)SkipStrongNames.xml" CopyToOutputDirectory="PreserveNewest" CopyToPublishDirectory="PreserveNewest" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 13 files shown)


build/sources.props
```diff
@@ -5,6 +5,7 @@
     <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
     <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
       $(RestoreSources);
+      https://www.myget.org/F/30de4ee06dd54956a82013fa17a3accb;
```
pranavkm: Regular MyGet feeds aren't super reliable. Also, we should be using the v3 protocol if we do reference a feed.

modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
```diff
@@ -5,13 +5,19 @@
     <AssemblyName>Internal.AspNetCore.KoreBuild.Tasks</AssemblyName>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <SkipStrongNamesDir>$(NuGetPackageRoot)\microsoft.web.skipstrongnames\$(MicrosoftWebSkipStrongNamesPackageVersion)\tools</SkipStrongNamesDir>
```
pranavkm: Trailing slash.

modules/KoreBuild.Tasks/module.props
```diff
@@ -1,4 +1,8 @@
 <Project>
+  <PropertyGroup>
+    <SkipStrongNamesExe>$(MSBuildThisFileDirectory)SkipStrongNames.exe</SkipStrongNamesExe>
```
pranavkm: Put this with the other properties (`DefaultDotnetAssetArch`)

modules/KoreBuild.Tasks/module.targets
```diff
@@ -9,6 +8,18 @@
     <DependencyVersionsFile Condition="'$(DependencyVersionsFile)' == ''">$(RepositoryRoot)build\dependencies.props</DependencyVersionsFile>
   </PropertyGroup>
 
+  <Target Name="EnableSkipStrongNames" Inputs="$(SkipStrongNamesXml)">
+    <Error Text="The parameter 'SkipStrongNamesXml' was not set." Condition="'$(SkipStrongNamesXml)' == ''" />
```
pranavkm: We'd want to include a file by default (See https://github.com/aspnet/AspNetWebStack/blob/master/tools/SkipStrongNames.xml). You could leave it empty now but have the right structure. The ideal workflow would be for us to do a regular build and this target kicks in without any additional configuration.

build/sources.props
pranavkm: @natemcmaster is dotnet.myget.org/gallery/aspnetcore-tools a viable feed to mirror this package to?

natemcmaster: Depends. Where is the source code for this package?

ryanbrandenburg: @Eilon says it's probably in TFS, @dougbu might know more.

natemcmaster: Can we copy this code into aspnet/BuildTools? I'd like to avoid adding a dependency to a thing that we don't know how to recreate. Also, build-from-source....

dougbu: What is the package of interest here?

natemcmaster: microsoft.web.skipstrongnames

dougbu: Oddly the package says that comes from the old ASP.NET Web Stack site on CodePlex. But, I can't see anything in TFS for it. I'll look around a few other places.

pranavkm: All this does is set \ remove a registry key under a specific location. If it's important to have sources for this rather than consume it as a binary, it'll be easier to just rewrite it as a task.

dougbu: Gallery shows v3 source is https://www.myget.org/F/30de4ee06dd54956a82013fa17a3accb/api/v3/index.json

Eilon: Yeah I talked to @ryanbrandenburg yesterday. If we can't find the source, let's just decompile w/ ILSpy because I figure the whole app is like a 50-line `Program.Main()` and nothing more (it just reads an XML file and does a `for()` loop to 'register' each assembly/public key token). Whether we do it as a task or a tool is up to whoever wants to write it.

modules/KoreBuild.Tasks/module.targets
```diff
@@ -1,19 +1,21 @@
...
 
     <DisablePackageReferenceRestrictions Condition=" '$(DisablePackageReferenceRestrictions)' == '' ">false</DisablePackageReferenceRestrictions>
     <KoreBuildConfigFile Condition="'$(KoreBuildConfigFile)' == ''">$(RepositoryRoot)korebuild.json</KoreBuildConfigFile>
     <DependencyVersionsFile Condition="'$(DependencyVersionsFile)' == ''">$(RepositoryRoot)build\dependencies.props</DependencyVersionsFile>
+
   </PropertyGroup>
 
   <Target Name="EnableSkipStrongNames">
-    <Exec Command='"$(SkipStrongNamesExe)" -e "$(SkipStrongNamesXml)"' />
+    <Exec Command='dotnet "$(KoreBuildConsoleDll)" skip-strong-name enable --xml-file "$(SkipStrongNamesXml)"' />
```
natemcmaster: This will most likely work find, but FYI you can use the [RunDotNet task](https://github.com/aspnet/BuildTools/blob/dev/modules/BuildTools.Tasks/RunDotNet.cs) to avoid potential issues in PATH.
```
<RunDotNet Command='"$(KoreBuildConsoleDll)" skip-strong-name enable --xml-file "$(SkipStrongNamesXml)"' />
```

modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
```diff
@@ -5,13 +5,18 @@
     <AssemblyName>Internal.AspNetCore.KoreBuild.Tasks</AssemblyName>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <SkipStrongNamesExe>$(NuGetPackageRoot)\microsoft.web.skipstrongnames\$(MicrosoftWebSkipStrongNamesPackageVersion)\tools\SkipStrongNames.exe</SkipStrongNamesExe>
```
natemcmaster: Not needed?

tools/KoreBuild.Console/Commands/SkipStrongName/BaseSkipStrongNameCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public override void Configure(CommandLineApplication application)
+        {
+            _xmlFileOption = application.Option("-x|--xml-file", "The file to use in skipping commands", CommandOptionType.SingleValue);
+        }
+
+        protected override bool IsValid()
+        {
+            if (string.IsNullOrEmpty(XmlFile))
```
natemcmaster: Add also a `File.Exists` check.

tools/KoreBuild.Console/Commands/DisableSkipStrongNameCommand.cs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Win32;
+using SkipStrongNames;
+
+namespace KoreBuild.Console.Commands
+{
+    internal class DisableSkipStrongNameCommand : BaseSkipStrongNameCommand
```
pranavkm: Can we make this a task? If it works, it's much nicer than spawning a process and reading xml files.

modules/KoreBuild.Tasks/DisableSkipStrongName.cs
```diff
@@ -1,25 +1,26 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
 using Microsoft.Win32;
-using SkipStrongNames;
 
-namespace KoreBuild.Console.Commands
+namespace KoreBuild.Tasks
 {
-    internal class DisableSkipStrongNameCommand : BaseSkipStrongNameCommand
+    public class DisableSkipStrongName : Task
```
natemcmaster: 👏

modules/KoreBuild.Tasks/DisableSkipStrongName.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Microsoft.Win32;
+
+namespace KoreBuild.Tasks
+{
+    public class DisableSkipStrongName : Task
+    {
+        [Required]
+        public string XmlFile { get; set; }
```
pranavkm: I'll leave it to you - you could pass this in as an ItemGroup. Not sure it's worth the effort since the current code works.

PR #504, opened 2017-12-04
Add tasks to generate bill of materials and compute/verify file checksums

natemcmaster:
Per discussion on https://github.com/aspnet/Universe/issues/679

### DIFF
--- a/modules/KoreBuild.Tasks/ComputeManyChecksums.cs
+++ b/modules/KoreBuild.Tasks/ComputeManyChecksums.cs
@@ -1,0 +1,40 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Build.Framework;
+
+namespace KoreBuild.Tasks
+{
+    /// <summary>
+    /// Computes the hash value for many files and updates the metadata of an item with its value.
+    /// </summary>
+    public class ComputeManyChecksum : Microsoft.Build.Utilities.Task
+    {
+        /// <summary>
+        /// A group of files.
+        /// </summary>
+        [Required]
+        [Output]
+        public ITaskItem[] Items { get; set; }
+
+        /// <summary>
+        /// The algorithm. Allowed values: SHA256, SHA384, SHA512.
+        /// </summary>
+        public string Algorithm { get; set; } = "SHA256";
+
+        /// <summary>
+        /// The metadata name to set.
+        /// </summary>
+        public string MetadataName { get; set; } = "FileHash";
+
+        public override bool Execute()
+        {
+            foreach (var file in Items)
+            {
+                file.SetMetadata(MetadataName, HashHelper.GetFileHash(Algorithm, file.ItemSpec));
+            }
+
+            return true;
+        }
+    }
+}
--- a/modules/KoreBuild.Tasks/GenerateBillOfMaterials.cs
+++ b/modules/KoreBuild.Tasks/GenerateBillOfMaterials.cs
@@ -1,0 +1,133 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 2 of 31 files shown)


modules/KoreBuild.Tasks/ComputeManyChecksums.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// Computes the hash value for many files and updates the metadata of an item with its value.
+    /// </summary>
+    public class ComputeManyChecksum : Microsoft.Build.Utilities.Task
+    {
+        /// <summary>
+        /// A group of files.
+        /// </summary>
+        [Required]
+        [Output]
+        public ITaskItem[] Items { get; set; }
```
ryanbrandenburg: So this parameter is both the output and the required input? Do we really have a true output here other than editing the files?

modules/KoreBuild.Tasks/GenerateBillOfMaterials.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace KoreBuild.Tasks
+{
+    public class GenerateBillOfMaterials : Microsoft.Build.Utilities.Task
+    {
+        [Required]
+        public ITaskItem[] Artifacts { get; set; }
+
+        [Required]
+        [Output]
+        public string OutputPath { get; set; }
```
ryanbrandenburg: Same question as above.

modules/KoreBuild.Tasks/VerifyChecksum.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// The file path.
+        /// </summary>
+        [Required]
+        public string File { get; set; }
+
+        /// <summary>
+        /// The algorithm. Allowed values: SHA256, SHA384, SHA512.
+        /// </summary>
+        public string Algorithm { get; set; } = "SHA256";
```
ryanbrandenburg: Should this be an Enum since we have a defined (and small) set of allowed values, or do tasks not play well with enums? Kind syntactic sugar either way.

scripts/bootstrapper/build.ps1
```diff
@@ -84,6 +88,10 @@ function Get-KoreBuild {
     $version = $version.TrimStart('version:').Trim()
     $korebuildPath = Join-Paths $DotNetHome ('buildtools', 'korebuild', $version)
 
+    if ($Reinstall -and (Test-Path $korebuildPath)) {
```
ryanbrandenburg: Why do we need `-Reinstall` now?

modules/KoreBuild.Tasks/ComputeManyChecksums.cs
natemcmaster: This is a MSBuild-ism required to update items in-place with new metadata. The expected usage is:

```xml
    <ComputeManyChecksum Items="@(ArtifactInfo)">
      <Output TaskParameter="Items" ItemName="_ArtifactInfoWithFileHash" />
    </ComputeManyChecksum>
```

scripts/bootstrapper/build.ps1
natemcmaster: So the functional tests can force a re-install of a new build. I removed the setting that used a random, time-based build number for each build, so each build of build tools will be `2.0.0-rtm-t000`.

modules/KoreBuild.Tasks/VerifyChecksum.cs
natemcmaster: I wish. MSBuild doesn't support enum parameters on tasks.

scripts/bootstrapper/build.ps1
ryanbrandenburg: Understood, so we don't need to propagate this change out to all the repos since it's test only, but when it does eventually copy that's fine.

natemcmaster: Yup. It's a non-essential change for other repos.

natemcmaster:
Close/re-open to trigger travis.

ryanbrandenburg:
Didn't know closing and re-opening triggered, good to know.

natemcmaster:
It sometimes works, but didn't this time. :-(

PR #505, opened 2017-12-07
Add Internal.AspNetCore.SiteExtension.Sdk package

pakrym:
https://github.com/aspnet/Universe/pull/697


src/Internal.AspNetCore.HostingStartup.Sdk/TrimDeps.cs
```diff
@@ -0,0 +1,60 @@
...
+    public class TrimDeps : Task
+    {
+        [Required]
+        public ITaskItem[] DepsFiles { get; set; }
+
+        public override bool Execute()
+        {
+            foreach (var depsFile in DepsFiles)
+            {
+                ChangeEntryPointLibraryName(depsFile.GetMetadata("Identity"));
```
natemcmaster: FYI - this is identical to `depsFile.ItemSpec`. I have a slight preference for this instead of .GetMetadata, but its functionally the same.

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.props
```diff
@@ -0,0 +1,17 @@
...
+-->
+<Project TreatAsLocalProperty="TaskFolder;TaskAssembly">
+  <PropertyGroup>
+    <TaskFolder Condition=" '$(MSBuildRuntimeType)' == 'Core' ">netstandard1.6</TaskFolder>
+    <TaskFolder Condition=" '$(MSBuildRuntimeType)' != 'Core' ">net46</TaskFolder>
+    <TaskAssembly>$(MSBuildThisFileDirectory)..\tasks\$(TaskFolder)\Internal.AspNetCore.HostingStartup.Sdk.dll</TaskAssembly>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Remove="obj\depswork\*.*" />
```
natemcmaster: BaseIntermediateOutputPath is excluded by default. Why was this needed?

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.targets
```diff
@@ -0,0 +1,50 @@
...
+  This file is automatically imported by NuGet into a user's project
+  when it targets a single framework, or in classic (pre 2017) csproj projects.
+-->
+
+<Project>
+
+  <Target Name="GenerateHostingStartupDeps" BeforeTargets="Build" Condition="'@(HostingStartupPackageReference->Count())' != '0'">
+    <PropertyGroup>
+      <_TemplatesDirectory>$(MSBuildThisFileDirectory)..\content\</_TemplatesDirectory>
+      <_DepsOutputDirectory>$(MSBuildProjectDirectory)\$(BaseIntermediateOutputPath)</_DepsOutputDirectory>
```
natemcmaster: Use `$(IntermediateOutputPath)` instead of BaseIntermediateOutputPath. This will already have been normalized for config + target framework, and have been turned into a full path.

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.targets
```diff
@@ -0,0 +1,50 @@
...
+      <_WorkingDirectory>$(_DepsOutputDirectory)\depswork</_WorkingDirectory>
+      <_BasePackagePath>content\additionaldeps\</_BasePackagePath>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <_TemplateFiles Include="$(MSBuildThisFileDirectory)..\content\HostingStartup\*.*" />
```
natemcmaster: Should this be recursive? This will only capture files in the top directory.

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.targets
```diff
@@ -0,0 +1,50 @@
...
+    </ItemGroup>
+
+    <RemoveDir Directories="%(_HostingStartupPackageReference.WorkingDirectory)" />
+
+    <Copy SourceFiles="@(_TemplateFiles)" DestinationFolder="%(_HostingStartupPackageReference.WorkingDirectory)" />
+
+    <MSBuild Projects="%(_HostingStartupPackageReference.Project)" Targets="Restore" />
+
+    <MSBuild Projects="%(_HostingStartupPackageReference.Project)"
+             Targets="Restore;Clean;Publish"
```
natemcmaster: Remove duplicate call to the Restore target

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.targets
```diff
@@ -0,0 +1,50 @@
...
+        TrimmedDepsFile="$(_DepsOutputDirectory)%(HostingStartupPackageReference.Identity)\%(HostingStartupPackageReference.Identity).deps.json"
+        PackagePath="%(HostingStartupPackageReference.Identity)\shared\Microsoft.NETCore.App\$(HostingStartupRuntimeFrameworkVersion)\"
+         />
+    </ItemGroup>
+
+    <RemoveDir Directories="%(_HostingStartupPackageReference.WorkingDirectory)" />
+
+    <Copy SourceFiles="@(_TemplateFiles)" DestinationFolder="%(_HostingStartupPackageReference.WorkingDirectory)" />
+
+    <MSBuild Projects="%(_HostingStartupPackageReference.Project)" Targets="Restore" />
```
natemcmaster: Restore shouldn't succeed unless you have the properties set `HostingStartupPackageName=%(_HostingStartupPackageReference.Identity);HostingStartupPackageVersion=%(_HostingStartupPackageReference.Version)`, right?

src/Internal.AspNetCore.HostingStartup.Sdk/buildMultiTargeting/Internal.AspNetCore.HostingStartup.Sdk.props
```diff
@@ -0,0 +1,7 @@
+<!--
+  This file is automatically imported by NuGet into a user's project
```
natemcmaster: What are the scenarios for mutli-targeting hosting startup projects? Adding these imports is unlikely to work in a multi-target scenario since these projects invoke a new project build context per TFM.

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.targets
```diff
@@ -0,0 +1,50 @@
...
+    <MSBuild Projects="%(_HostingStartupPackageReference.Project)"
+             Targets="Restore;Clean;Publish"
+             Properties="PublishDir=%(_HostingStartupPackageReference.WorkingDirectory);HostingStartupPackageName=%(_HostingStartupPackageReference.Identity);HostingStartupPackageVersion=%(_HostingStartupPackageReference.Version)" />
+
+    <Copy SourceFiles="%(_HostingStartupPackageReference.DepsFile)" DestinationFiles="%(_HostingStartupPackageReference.TrimmedDepsFile)" />
+
+    <RepoTasks.TrimDeps DepsFiles="%(_HostingStartupPackageReference.TrimmedDepsFile)" />
+
+    <ItemGroup>
+      <Content Include="%(_HostingStartupPackageReference.TrimmedDepsFile)" PackagePath="%(_HostingStartupPackageReference.PackagePath)" />
```
natemcmaster: May not work when calling `dotnet pack --no-build` and in multi-targeting scenarios. Targets that add custom content to packages should add items to the `BuildOutputInPackage` item group and chain into TargetsForTfmSpecificBuildOutput.

cref https://github.com/NuGet/NuGet.Client/pull/1255

Example:
```xml
<PropertyGroup>
  <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);MyGenTarget</TargetsForTfmSpecificBuildOutput>
</PropertyGroup>

<Target Name="MyGenTarget">
 <!-- do stuff-->
  <ItemGroup>
     <BuildOutputInPackage Include="myfile.txt" />
   </ItemGroup>
</Target>
```

src/Internal.AspNetCore.HostingStartup.Sdk/buildMultiTargeting/Internal.AspNetCore.HostingStartup.Sdk.props
pakrym: I just copied this from the sample. Would it work fine if I remove them?

natemcmaster: Yup, it should work ok

src/Internal.AspNetCore.HostingStartup.Sdk/build/Internal.AspNetCore.HostingStartup.Sdk.targets
pakrym: Problem with `BuildOutputInPackage` is that it put's files into `lib` subdirectory even if `TargetPath` is set. So I have to set it to `TargetPath="..\..\$(_BasePackagePath)"`

pakrym: I think I'll stick with `Content` for now and see if we have any issues.

natemcmaster: Hmm, what about <TfmSpecificPackageFile> + TargetsForTfmSpecificContentInPackage?

https://github.com/NuGet/NuGet.Client/blob/4.3.0.4339/src/NuGet.Core/NuGet.Build.Tasks.Pack/Pack.targets#L275-L278

Also, you might consider simplifying this by just using a nuspec file. I've found this is the only way to do anything custom.

cc @rohit21agrawal

src/Internal.AspNetCore.SiteExtension.Sdk/Internal.AspNetCore.SiteExtension.Sdk.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <Description>Build targets and extensions to Microsoft.NET.Sdk. This package is intended for Microsoft use only.</Description>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
+    <NuspecFile>$(MSBuildThisFileDirectory)$(MSBuildProjectName).nuspec</NuspecFile>
+    <IncludeSymbols>false</IncludeSymbols>
+    <IncludeSource>false</IncludeSource>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="$(Tooling_NewtonsoftJsonPackageVersion)" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="15.1.1012" />
```
JunTaoLuo: Are we sure we want these as hardcoded version numbers? @natemcmaster

src/Internal.AspNetCore.SiteExtension.Sdk/Internal.AspNetCore.SiteExtension.Sdk.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageId>Internal.AspNetCore.SiteExtension.Sdk</PackageId>
+    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
```
JunTaoLuo: Just double checking to see we want to target netcoreapp2.0 and not 2.1?

JunTaoLuo: Actually, this is an internal tool so this makes sense.

src/Internal.AspNetCore.SiteExtension.Sdk/Internal.AspNetCore.SiteExtension.Sdk.csproj
natemcmaster: True. There are package versions available for this. Would be better to use those.

src/Internal.AspNetCore.SiteExtension.Sdk/Internal.AspNetCore.SiteExtension.Sdk.csproj
natemcmaster: Right. This is for a task assembly, not a runtime one.

src/Internal.AspNetCore.SiteExtension.Sdk/Internal.AspNetCore.SiteExtension.Sdk.csproj
pakrym: fixed

src/Internal.AspNetCore.SiteExtension.Sdk/Internal.AspNetCore.SiteExtension.Sdk.csproj
pakrym: Same as in `Internal.AspNetCore.Sdk`

pakrym:
@JunTaoLuo 
Site extension project would look the following way:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <HostingStartupRuntimeFrameworkVersion>2.0.0</HostingStartupRuntimeFrameworkVersion>
  </PropertyGroup>

  <ItemGroup>
    <HostingStartupPackageReference Include="microsoft.aspnetcore.azureappservices.hostingstartup" Version="2.0.1" />
    <HostingStartupPackageReference Include="microsoft.aspnetcore.applicationinsights.hostingstartup" Version="2.0.1" />

    <PackageReference Include="Internal.AspNetCore.SiteExtension.Sdk" Version="2.1.0-*" />
  </ItemGroup>

</Project>

```

And when you dotnet pack it magic things would happen and deps and dlls would get embedded into the package.

PR #506, opened 2017-12-07
Make SourceLink use the correct slash

ryanbrandenburg:
In local testing I found that we needed an escaped `\` instead of a `/` to make VS use SourceLink successfully. This jives with dotnet's documentation [here](https://github.com/dotnet/core/blob/3cc74f3cd7ff6ca2d94e4e1b882e8551b5c32a50/Documentation/diagnostics/source_link.md#examples).

It looked like this slipped through as code cleanup in the PR and likely in testing the old package was accidentally used. I don't know of a way to test this "for real" outside of by-hand in VS, so unit tests and re-verifying once this goes through Universe will have to do.

### DIFF
--- a/modules/BuildTools.Tasks/CreateSourceLink.cs
+++ b/modules/BuildTools.Tasks/CreateSourceLink.cs
@@ -35,16 +35,20 @@ namespace Microsoft.AspNetCore.BuildTools
         {
             var rootDirectory = Path.Combine(Path.GetFullPath(RootDirectory), "*");
 
-            rootDirectory = rootDirectory.Replace("/", "\\");
-            rootDirectory = rootDirectory.Replace("\\", "\\\\");
+            var escapedPath = JsonEscapePath(rootDirectory);
 
             var codeSource = ConvertUrl();
 
-            File.WriteAllText(DestinationFile, $"{{\"documents\":{{\"{rootDirectory}\":\"{codeSource}\"}}}}");
+            File.WriteAllText(DestinationFile, $"{{\"documents\":{{\"{escapedPath}\":\"{codeSource}\"}}}}");
 
             SourceLinkFile = DestinationFile;
 
             return true;
+        }
+
+        private string JsonEscapePath(string path)
+        {
+            return path.Replace("\\", "\\\\");
         }
 
         private string ConvertUrl()
--- a/modules/BuildTools.Tasks/CreateSourceLink.cs
+++ b/modules/BuildTools.Tasks/CreateSourceLink.cs
@@ -34,7 +34,7 @@ namespace Microsoft.AspNetCore.BuildTools
         public override bool Execute()
         {
             var rootDirectory = Path.Combine(Path.GetFullPath(RootDirectory), "*");
-            rootDirectory = rootDirectory.Replace("\\", "/");
+            rootDirectory = rootDirectory.Replace("\\", "\\\\");
 
             var codeSource = ConvertUrl();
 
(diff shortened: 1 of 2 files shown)


modules/BuildTools.Tasks/CreateSourceLink.cs
```diff
@@ -34,7 +34,9 @@ public class CreateSourceLink : Task
         public override bool Execute()
         {
             var rootDirectory = Path.Combine(Path.GetFullPath(RootDirectory), "*");
-            rootDirectory = rootDirectory.Replace("\\", "/");
+
+            rootDirectory = rootDirectory.Replace("/", "\\");
```
natemcmaster: If I understand [this doc](https://github.com/dotnet/core/blob/3cc74f3cd7ff6ca2d94e4e1b882e8551b5c32a50/Documentation/diagnostics/source_link.md#examples) correctly, we should normalize `/` and `\` to `System.IO.Path.DirectorySeparatorChar` so LInux builds also work.

tmat: The paths should be the same as passed to the csc. The only escaping that should be done here is JSON escaping (```\``` -> ```\\```).

natemcmaster: Didn't know that existed. Thanks for the pointer.

Do you have any plans to make this a feature of the SDK? I'm sure other devs would like this too, but it took a lot of research to figure this out. Even then, we still don't have it 100% working.

tmat: Yes, we do. Working on it.

tmat: BTW, do you build with Deterministic=true?

natemcmaster: 👍 cool.

We haven't changed the default AFAIK. Deterministic is true by default, right?

tmat: Yes. However, to get fully deterministic build you'll also need to set PathMap. This will make paths embedded in the DLL and PDBs independent of the directory the repo was cloned to.
See https://github.com/dotnet/roslyn/blob/master/build/Targets/Settings.props#L136 and https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/tools/Workarounds.targets#L103

tmat: Another thing to consider is submodules. They complicate things.

ryanbrandenburg: Submodules complicate things for your SourceLink targets or for deterministic builds? How do they complicate things? Our packages are produced by Universe, which does use submodules.

ryanbrandenburg:
:up::date: @tmat it seems like what we're doing now is equivalent to your target, and is working now, so I think we'll stick with the task for the time being rather than take a dependency on you in this repo. Super excited to hear that it's making its way into the SDK though, I think lots of devs will find that useful, and at that point to can evaluate using the SDK to do this instead.

modules/BuildTools.Tasks/CreateSourceLink.cs
tmat: For SourceLink, because the simple logic implemented in [SourceLink.targets](https://github.com/dotnet/roslyn-tools/blob/master/sdks/RepoToolset/tools/SourceLink.targets) doesn't automatically include adding submodule entries into source link file. Once we have source link integrated in dotnet SDK we will handle them.

natemcmaster: Submodules isn't really a problem for us at the moment. We do not have any projects that produce DLLs from source files consumed from a submodule. We use submodules, but only for coordinating a build of [multiple repos together](https://github.com/aspnet/Universe/tree/dev/modules), not for sharing sources between them.

tmat: Good :)

natemcmaster:
We should also implement the deterministic path map value, but I can send that in a follow-up PR.

vancem:
@vancem FYI

PR #511, opened 2017-12-09
Update PackageOwnershipRule to take advantage of reserved prefixes

### LINKED ISSUE #484: Update PackageOwnershipRule to take advantage of reserved prefixes
NuGet.org added a feature where owners can reserve a prefix so that a family of packages that share this can only be uploaded by the specific owner. As part of this change, the AspNetCore team already owns the following prefixes:

* Microsoft.AspNetCore.*
* Microsoft.AspNet.*
* Microsoft.CodeAnalysis.Razor
* Microsoft.Data.Sqlite.*
* Microsof.Dotnet.*
* Microsoft.Extensions.*
* Microsoft.EntityFrameworkCore.*
* Microsoft.Net.Http.*
* Microsoft.Owin.*
* Microsoft.VisualStudio.*

The check in the tool needs to be updated to exclude any package that starts with this prefix (since we do not need to explicitly squat on those ids). The workflow for package reservation needs to also be updated to account for this new prefix ownership requirement.

pranavkm:
Additionally fix an issue where running the rule on assemblies with
NotNullAttribute was throwing.

Fixes #484


modules/NuGetPackageVerifier/console/Rules/AssemblyIsBuiltInReleaseConfigurationRule.cs
```diff
@@ -20,47 +20,70 @@ public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context
...
-                                    var type = reader.GetTypeReference((TypeReferenceHandle)ctor.Parent);
-                                    var typeName = reader.GetString(type.Name);
-                                    if (string.Equals(typeof(DebuggableAttribute).Name, typeName, StringComparison.Ordinal))
-                                    {
-                                        var attribute = customAttribute.DecodeValue(NullProvider.Instance);
+                                var ctor = reader.GetMemberReference((MemberReferenceHandle)customAttribute.Constructor);
+                                var type = reader.GetTypeReference((TypeReferenceHandle)ctor.Parent);
+                                typeName = reader.GetString(type.Name);
+                            }
+                            else if (customAttribute.Constructor.Kind == HandleKind.MethodDefinition)
```
pranavkm: Need to figure out why this doesn't happen on the CI. Locally running this on the signed package list was failing for the test specification package when it encountered `NotNullAttribute`

pranavkm:
Discussed with @Eilon offline. The plan is to keep a list of 3rd party packages that have one of the reserved prefixes and use fail the task if the package id matches. This avoids the need to do requests to the gallery as part of running the tool and maintaining a list of owned packages.

modules/NuGetPackageVerifier/console/Rules/PackageOwnershipRule.cs
```diff
@@ -3,127 +3,199 @@
...
+            "Microsoft.AspNetCore.Authentication.QQ",
+            "Microsoft.AspNetCore.Authentication.QQConnect",
+            "Microsoft.AspNetCore.HealthChecks",
+            "Microsoft.AspNetCore.Identity.MongoDB",
+            "Microsoft.AspNetCore.JsonPatch.Net40",
+            "Microsoft.AspNetCore.Mono",
+            "Microsoft.AspNetCore.Mvc.Formatters.Xml.Extensions",
+            "Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer",
+            "Microsoft.AspNetCore.Mvc.Versioning",
+            "Microsoft.AspNetCore.OData",
```
pranavkm: Are they aware of this?

Eilon: Dunno. @danroth27 can you get in touch with the OData folks to let them know that some 3rd party has taken up some names that they might otherwise want to use? (Just an FYI, it's up to them if they want to do anything about it.)

modules/NuGetPackageVerifier/console/Rules/PackageOwnershipRule.cs
```diff
@@ -30,8 +30,163 @@ public class PackageOwnershipRule : IPackageVerifierRule
...
+            "Microsoft.AspNet.WebApi.WebHost.zh-Hans1",
+            "Microsoft.AspNet.WebPages-vHalfNext",
+            "Microsoft.AspNetCore.ApiHelp.Core",
+            "Microsoft.AspNetCore.ApiHelp",
+            "Microsoft.AspNetCore.ApiWidgets",
+            "Microsoft.AspNetCore.Authentication.ActiveDirectory",
+            "Microsoft.AspNetCore.Authentication.LinkedIn",
+            "Microsoft.AspNetCore.Authentication.QQ",
+            "Microsoft.AspNetCore.Authentication.QQConnect",
+            "Microsoft.AspNetCore.HealthChecks",
```
natemcmaster: I thought this was us: https://github.com/dotnet-architecture/HealthChecks/

natemcmaster: Nvm, forgot we backed it out of the 2.1 release for now

PR #512, opened 2017-12-11
Fix the already-installed check in dotnet-install when only shared runtime is being installed

natemcmaster:
When `--shared-runtime` is specified, the scripts were checking the wrong folder in the "is installed?" check.


natemcmaster:
I've also opened to make this fix in dotnet/cli. https://github.com/dotnet/cli/pull/8210

PR #515, opened 2017-12-11
Add additional source support to siteextension sdk

pakrym:


src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
```diff
@@ -17,6 +17,7 @@
       <_TemplateFiles Include="$(MSBuildThisFileDirectory)..\content\HostingStartup\**\*" />
       <_HostingStartupPackageReference
         Include="%(HostingStartupPackageReference.Identity)"
+        Source="%(HostingStartupPackageReference.Source)"
```
JunTaoLuo: Would the source be different between each hosting startup?

pakrym: Not in our case, wanted to keep things flexible.

PR #516, opened 2017-12-11
Update the dotnet SDK to 2.2.0-preview1-007796

ryanbrandenburg:


ryanbrandenburg:
Tested locally against Universe.

natemcmaster:
Does it work in Visual Studio too?

JunTaoLuo:
Can we update the runtime as well. It looks like it's set to 2.0.3 right now but we should ingest something newer. The SDK uses 2.1.0-preview1-26001-02 for the runtime for example.

ryanbrandenburg:
Universe doesn't have a good test case, but Logging worked.

ryanbrandenburg:
I'll update the Runtime as a separate thing so we don't delay getting people the SDK.

natemcmaster:
@JunTaoLuo the runtime is determined by repos using dependencies.props and the `<DotNetRuntime>` item group. Example: https://github.com/aspnet/Mvc/blob/dev/build/repo.props#L15-L18

JunTaoLuo:
@natemcmaster Ah right. So what's the purpose of this file: https://github.com/aspnet/BuildTools/blob/dev/files/KoreBuild/config/runtime.version?

natemcmaster:
It's the version that the console tools in KoreBuild need. e.g. NuGetPackageVerifier, KoreBuild.Console, ApiCheck

PR #517, opened 2017-12-12
Update runtime version

ryanbrandenburg:
Tested this locally against Universe command line and logging in VS.


ryanbrandenburg:
It appears that this won't work on a clean machine because it can't find ['Microsoft.NETCore.App', version '2.0.0'](https://travis-ci.org/aspnet/BuildTools/builds/315103143#L714).

files/KoreBuild/config/runtime.version
```diff
@@ -1 +1 @@
-2.0.3
```
natemcmaster: You should only change this if you also re-target all build tools to 2.1.

The repos themselves take care of installing the runtime via the `<DotNetRuntime>` item group. example:  https://github.com/aspnet/Mvc/blob/dev/build/repo.props#L15-L18

natemcmaster:
IMO, it would be better not to take this yet until .NET Core 2.1 is stable. Historically, we've run in roll-forward issues when our build tools depend on pre-release runtimes.

JunTaoLuo:
Spoke to @natemcmaster, yea it looks like we don't want to make the changes here. The version of the shared framework used by the universe build is actually controlled via https://github.com/aspnet/Universe/blob/dev/build/dependencies.props#L46

ryanbrandenburg:
@JunTaoLuo is the value in Universe high enough for your purposes?

Also, closing.

JunTaoLuo:
@ryanbrandenburg I'll test it and yell really loud if it isn't.

PR #520, opened 2017-12-14
Support spaces in path

mikaelm12:
@BrennanConroy


natemcmaster:
Spaces should have been handled by https://github.com/aspnet/BuildTools/blob/dev/modules/KoreBuild.Tasks/InstallDotNet.cs#L156. Can you run `build.cmd /v:n` and find the logger line that matches this:

https://github.com/aspnet/BuildTools/blob/63671122d77fccdc68c02b417cd4e534541ce587/modules/KoreBuild.Tasks/InstallDotNet.cs#L156-L164

PR #523, opened 2017-12-19
Add DotNetToolPackageRule to NGPV

natemcmaster:
Add support for a new package type - global CLI tools.

Also, fixed NullReferenceException's when running NGPV on the CLI tool packages

### DIFF
--- a/modules/NuGetPackageVerifier/console/PackageIssueFactory.cs
+++ b/modules/NuGetPackageVerifier/console/PackageIssueFactory.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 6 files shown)


natemcmaster:
FYI DotnetGlobalTool is just a placeholder name. The spec is unclear on what the `packageType` value should be for global CLI tools.

modules/NuGetPackageVerifier/console/PackageIssueFactory.cs
```diff
@@ -39,7 +66,7 @@ public static PackageVerifierIssue AssemblyMissingServicingAttribute(string asse
                 PackageIssueLevel.Error);
         }
 
-        public static PackageVerifierIssue AssemblyInformationalVersionDoesNotMatchPackageVersion(string assemblyPath, NuGetVersion assemblyInformationalVersion, NuGetVersion packageVersion, string packageId)
+        public static PackageVerifierIssue AssemblyInformationalVersionDoesNotMatchPackageVersion(string assemblyPath, string assemblyInformationalVersion, NuGetVersion packageVersion, string packageId)
```
ryanbrandenburg: Curious why we went from more to less specific?

modules/NuGetPackageVerifier/console/Rules/DotNetToolPackageRule.cs
```diff
@@ -0,0 +1,72 @@
...
+                if (entryPoint?.Value == null)
+                {
+                    yield return PackageIssueFactory.DotNetToolMalformedManifest("Missing EntryPoint");
+                    continue;
+                }
+
+                var runner = command.Attribute("Runner");
+                if (runner?.Value != "dotnet")
+                {
+                    continue;
```
ryanbrandenburg: Should there be an error here? Should we only be doing the above checks on dotnet tools or on everything?

modules/NuGetPackageVerifier/console/Rules/DotNetToolPackageRule.cs
```diff
@@ -0,0 +1,72 @@
...
+                }
+
+                // In the initial implementation of global dools, EntryPoint is just a filename, not a full path inside the package.
```
ryanbrandenburg: "global dools"

modules/NuGetPackageVerifier/console/PackageIssueFactory.cs
natemcmaster: For when NuGetVersion.TryParse fails.

modules/NuGetPackageVerifier/console/Rules/DotNetToolPackageRule.cs
natemcmaster: The goal was to skip further checks of Runner != "dotnet". Runner is not a required attribute on commands.

PR #529, opened 2017-12-23
Add infrastructure to support generating signing requests

natemcmaster:
Add tasks to generate a signing request from MSBuild projects. This will help us decouple to our authenticode signing infrastructure from the layout of artifacts and assemblies on disk.

Read the  docs/Signing.md for details on the file format and how to configure this in csproj files.

### DIFF
--- a/modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
+++ b/modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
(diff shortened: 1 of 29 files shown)


modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
ryanbrandenburg: What requires this change?

natemcmaster: I wanted to use `Path.GetRelativePath`, which is not in ns2.0, but is in netcoreapp2.0.

PR #532, opened 2017-12-28
Add KoreBuild properties section.

ryanbrandenburg:
Add a section kind of like https://github.com/aspnet/Universe/blob/dev/README.md#useful-properties-and-targets to the KoreBuild ReadMe.

### DIFF
--- a/docs/KoreBuild.md
+++ b/docs/KoreBuild.md
@@ -10,14 +10,24 @@ KoreBuild is a set of MSBuild targets, tasks, and console commands used to defin
 KoreBuild commands allow you to take wider and more granular actions.
 Previously repositories were runable in only one way, by doing `.\build.cmd`. But now, you can run multiple options.
 
-Command      | Purpose                                                          | Example
--------------|------------------------------------------------------------------|----------
-install-tools| Installs dotnet, CLI and Shared runtimes.                        | .\run.ps1 install-tools
-docker-build | Runs the build inside docker.                                    | .\run.ps1 docker-build {jessie\|winservercore} /t:SomeTarget /p:Parameters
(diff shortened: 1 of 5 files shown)


docs/KoreBuild.md
```diff
@@ -10,14 +10,23 @@ KoreBuild is a set of MSBuild targets, tasks, and console commands used to defin
...
+generate deps         | Generate a dependencies.props for this project.                  | .\run.ps1 generate deps
+generate api-baseline | Re-generate baselines for all projects.                          | .\run.ps1 generate api-baseline
+
+### KoreBuild properties
+
+Below is a list of some of the properties that KoreBuild recognizes which you might find useful.
+
+Property          | Purpose
+------------------|--------
+GenerateBaselines | Toggles the (re-)generation of baselines for all projects.
```
javiercn: I forgot to mention this yesterday. Shouldn't this be a task instead of a property? I don't want to run the entire build (test included) but just the minimum ammount of work to generate the baselines, so ideally something like build.cmd /t:GenerateApiCheckBaselines that would restore, compile and generate the baselines

javiercn:
Looks good to me, also backport this change to the 2.0.0 branch so that we can use it to generate baselines for the remaining repos that don't have updated baselines

docs/KoreBuild.md
ryanbrandenburg: When you use the `generate api-baselines` command it includes `SkipTests=true`. Making it a task instead of a property is messy. We want to operate on each project within the repo, which the normal build will already do for us.

ryanbrandenburg:
:up::date: with `generate api-baselines`.

PR #534, opened 2017-12-29
Update PackNuspec to support an explicit output path

natemcmaster:
Add PackNuspec.OutputPath and PackNuspec.Overwrite.

### DIFF
--- a/modules/KoreBuild.Tasks/PackNuSpec.cs
+++ b/modules/KoreBuild.Tasks/PackNuSpec.cs
@@ -14,22 +14,59 @@ using NuGet.Versioning;
 
 namespace KoreBuild.Tasks
 {
+    /// <summary>
+    /// Generates a nupkg from a nuspec file
+    /// </summary>
     public class PackNuSpec : Microsoft.Build.Utilities.Task
     {
+        /// <summary>
+        /// The path the nuspec file.
+        /// </summary>
         [Required]
         public string NuspecPath { get; set; }
 
-        [Required]
+        /// <summary>
+        /// Output nupkg is placed in folder + '$(id).$(version).nupkg'.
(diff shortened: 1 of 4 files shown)


modules/KoreBuild.Tasks/PackNuSpec.cs
```diff
@@ -14,22 +14,59 @@
...
+        /// </summary>
         [Required]
         public string NuspecPath { get; set; }
 
-        [Required]
+        /// <summary>
+        /// Output nupkg is placed in folder + '$(id).$(version).nupkg'.
+        /// Either this or <see cref="OutputPath" /> must be specified.
+        /// </summary>
         public string DestinationFolder { get; set; }
```
ryanbrandenburg: Maybe I'm still waking up, but this seems unclear on the difference between `DestinationFolder` and `OutputPath`. is it DestinationFolder -> `$(DestinationFolder)\$(id).$(version).nupkg\$(nupkgName)` OutputPath -> `$(OutputPath)\$(nupkgName)`? What scenarios would make someone want to pick one over the other?

natemcmaster:
```xml
<PackNuspec DestinationFolder="artifacts\build\" /> 
<!-- Produces: artifacts\build\PackageId.PackageVersion.nupkg -->

<PackNuspec OutputPath="artifacts\build\whateverfilenameiwant.nupkg" />
<!-- Produces: artifacts\build\whateverfilenameiwant.nupkg-->
```

natemcmaster: To further clarify: you would want to use OutputPath when producing a nupkg that shouldn't have the $(Version) info in the file name.

PR #535, opened 2018-01-02
Use DOTNET_HOME in push.ps1 if it exists

ryanbrandenburg:


scripts/PushNuGet.psm1
```diff
@@ -78,14 +85,21 @@ function Push-NuGetPackage {
...
+
                 while ($remaining -ge 0) {
                     $arguments = @()
                     if ($apikey) {
                         $arguments = ('--api-key', $apikey)
                     }
 
                     try {
-                        & dotnet nuget push `
+                        & $dotnetExe nuget push `
```
natemcmaster: Try this: `$using:dotnetExe`

scripts/PushNuGet.psm1
```diff
@@ -78,14 +85,21 @@ function Push-NuGetPackage {
                 $ErrorActionPreference = 'Stop'
                 Set-StrictMode -Version Latest
 
+                if ($env:DOTNET_HOME -ne $null) {
```
natemcmaster: I don't think you need to repeat this if you use the `$using` variable scope. 

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_remote_variables

natemcmaster:
Works as is, so feel free to merge. I just learned over break about the `$using:` scope in PS, and thought you might like to know about it too.

ryanbrandenburg:
Nice! :up::date: and will merge once travis and appveyor pass.

PR #537, opened 2018-01-03
ApiCheck generation 2.0.0

ryanbrandenburg:
Fix baseline generation for 2.0.0 and fix the `test.ps1` to work when passing parameters.

### DIFF
--- a/test.ps1
+++ b/test.ps1
@@ -2,8 +2,6 @@
 #requires -version 4
 [CmdletBinding(PositionalBinding = $true)]
 param(
-    [Parameter]
-    [string]$Command = 'default-build',
     [Parameter(Mandatory = $true)]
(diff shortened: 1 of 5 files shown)


test.ps1
```diff
@@ -2,10 +2,10 @@
 #requires -version 4
 [CmdletBinding(PositionalBinding = $true)]
 param(
-    [Parameter]
-    [string]$Command = 'default-build',
     [Parameter(Mandatory = $true)]
     [string]$RepoPath,
+    [Parameter(Position = 1)]
```
natemcmaster: Position = 0?

natemcmaster: On second thought, this doesn't apply in 2.0.0. This was before the run.ps1 factoring, so we can drop this parameter altogether

PR #541, opened 2018-01-04
Allow vstest diag generation

ryanbrandenburg:
This should make it easier to diagnose problems with vstest.


files/KoreBuild/modules/vstest/module.targets
```diff
@@ -66,6 +66,7 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
   <PropertyGroup>
     <TrxFile>$(ArtifactsDir)logs\$(TestGroupName)-$(TargetFramework)-$(BuildNumber).trx</TrxFile>
     <VSTestLogger Condition=" '$(VSTestLogger)' == '' AND '$(TEAMCITY_VERSION)' != '' ">trx;LogFileName=$(TrxFile)</VSTestLogger>
+    <VSTestDiagFile Condition=" '$(VSTestDiag)' == 'true'">$(ArtifactsDir)logs\$(TestGroupName)-$(TargetFramework)-$(BuildNumber).diag</VSTestDiagFile>
```
natemcmaster: Btw, I added a new korebuild var, LogOutputDir. Looks like I missed this file when adding it. Effectively the same so doesn't matter.

https://github.com/aspnet/BuildTools/blob/c390edb6f73e3ab992402835c004414c1fadfd4a/files/KoreBuild/KoreBuild.Common.props#L40

PR #542, opened 2018-01-05
Update SDK

pakrym:


natemcmaster:
As expected, this lifted nuget to a version that has this bug in it: https://github.com/NuGet/Home/issues/6351. Looks like we need to update https://github.com/aspnet/Scaffolding/tree/dev/src/VS.Web.CG.Tools to workaround it

pakrym:
What impact would disabling package verifier rule until the nuget bug is fixed cause? Would the absence of this file break things?

natemcmaster:
The CLI tool wouldn't work as expected. This is a valid failure that we should fix. Rohit seemed to think this should be as simple as updating the PackagePath setting in our csproj file

rohit21agrawal:
yes, please see the workaround here:
https://github.com/NuGet/Home/issues/6351#issuecomment-353654853

pakrym:
https://github.com/aspnet/Scaffolding/pull/670

natemcmaster:
I think @muratg  is also going to write a script to scan our nupkgs for any other extensionless files. I can't think of any others off the top of my head. I think dotnet-ef uses nuspec directly, so they shouldn't be affected

PR #545, opened 2018-01-08
Normalize backslashes in zip entries from UnzipArchive

natemcmaster:


modules/BuildTools.Tasks/UnzipArchive.cs
```diff
@@ -62,7 +67,18 @@ public override bool Execute()
             {
                 foreach (var entry in zip.Entries)
                 {
-                    var fileDest = Path.Combine(Destination, entry.FullName);
+                    var entryPath = entry.FullName;
+                    if (!DisablePathNormalization)
+                    {
+                        if (entry.FullName.IndexOf('\\') >= 0)
```
ryanbrandenburg: Should these be `Path.DirectorySeperator` or is there some reason that it really always should be '/'?

test/BuildTools.Tasks.Tests/UnzipArchiveTest.cs
```diff
@@ -149,6 +149,30 @@ public void DoesNotOverwrite()
             Assert.Equal(contents, File.ReadAllText(path));
         }
 
+        [Fact]
+        public void ItNormalizesBacklashesInPath()
```
ryanbrandenburg: Did we test this against Ubuntu or some other non-windows OS?

natemcmaster: MacOS. I confirmed this by running on tests before/after on my macbook

modules/BuildTools.Tasks/UnzipArchive.cs
natemcmaster: `\` is considered a valid file name character on \*nix file systems, which is why we end up with unzip layouts like this. 
image

We should never produce ZIPs with backslashes in the entry name. `ZipArchive` normalizes these. I've hard-coded the check this way because I want our builds to issue warnings about bad zips.

PR #546, opened 2018-01-09
New Nuget version

ryanbrandenburg:
Take the new NuGet version which will be used in MSBuild. This version includes some breaking changes.

### DIFF
--- a/tools/KoreBuild.Console/Commands/RootCommand.cs
+++ b/tools/KoreBuild.Console/Commands/RootCommand.cs
@@ -22,7 +22,7 @@ namespace KoreBuild.Console.Commands
             application.Command("upgrade", c =>
             {
                 c.HelpOption("-h|--help");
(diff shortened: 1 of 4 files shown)


tools/KoreBuild.Console/Commands/RootCommand.cs
```diff
@@ -22,7 +22,7 @@ public override void Configure(CommandLineApplication application)
             application.Command("upgrade", c =>
             {
                 c.HelpOption("-h|--help");
-                c.Command("deps", new DependenciesUpgradeCommand(context).Configure);
+                c.Command("deps", new DependenciesUpgradeCommand(context).Configure, throwOnUnexpectedArg: false);
```
ryanbrandenburg: We had to do this to make "upgrade deps" work from our `./test.ps1` script, which passes in some parameters.

natemcmaster: 👍

PR #547, opened 2018-01-11
Update SDK to 2.2.0-preview1-007927

natemcmaster:


natemcmaster:
Ran a build of Universe locally. This appears to pass, so moving forward. Will revert if the gatekeeper fails on CI.

PR #550, opened 2018-01-16
Use TeamCity x-plat logger

### LINKED ISSUE #345: Add a TeamCity logger
We should implement a logger that is better suited for TeamCity output. I've asked JetBrains about this and their recommendation was to use their dotnet CLI plugin instead. We prefer not to use these plugins but instead rely on scripts to work well for us, so we should consider making our own logger.

Featuers:

 - Log TeamCity blocks. Makes it easier to navigate output (we currently spit out ~100 MB of text).
 - Produce cleaner console output showing the KoreBuild lifecycle
 - Use flow ID to keep messages from parallel processes inline.

natemcmaster:
Resolves #345


files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -102,20 +105,31 @@ function Invoke-RepositoryBuild(
 
         $msBuildArguments | Out-File -Encoding ASCII -FilePath $msBuildResponseFile
 
+        if ($env:KOREBUILD_TEAMCITY_LOGGER) {
+            @"
+/noconsolelogger
+/verbosity:normal
+"/logger:TeamCity.MSBuild.Logger.TeamCityMSBuildLogger,${env:KOREBUILD_TEAMCITY_LOGGER};teamcity"
```
pranavkm: Is the plan to manually put the binary on individual agents?

natemcmaster: Put on agents, yes. Manually, no. TC allows uploading arbitrary .zip files that are automatically installed to all agents. I already set this up so all configs have this:

image

natemcmaster:
I'm going to revert this. Unfortunately, the logger doesn't pass through service messages, so .trx files and failures from NGPV aren't going to flow through this logger correctly. We'll have to look into a way to workaround this.

PR #551, opened 2018-01-16
Better logging and more skips for StrongName

ryanbrandenburg:
Bump the importance of messages from these tasks so we actually see them, and include more assemblies to the list of things we'll actually skip.

### DIFF
--- a/modules/KoreBuild.Tasks/SkipStrongNames.xml
+++ b/modules/KoreBuild.Tasks/SkipStrongNames.xml
@@ -1,4 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <assemblies defaultPublicKeyToken="31bf3856ad364e35">
+  <assembly name="Microsoft.Win32.Registry" />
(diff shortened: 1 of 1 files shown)


natemcmaster:
> Bump the importance of messages from these tasks so we actually see them

Did you try `/verbosity:normal`?

ryanbrandenburg:
You can also make it visible by changing the verbosity, but since these messages are the core of what we're doing in this target I felt they warranted high importance so that they're always visible.

natemcmaster:
As a rule of thumb, I'm in favor of minimal output where possible, and I don't think these messages make the cut for "high" importance, especially if we start running the targets on every build....but not worth blocking this PR on that so :shipit:

natemcmaster:
Side note: the High/Medium verbosity issue is less important in the context of #345 where CI logs will default to `/verbosity:normal` and produce blocks that make CI output more readable.

image

ryanbrandenburg:
:up::date:

PR #553, opened 2018-01-19
Rename VSTestDiag to VSTestVerboseOutput

natemcmaster:
Turns out VSTestDiag is already used by dotnet-test, and they expect it to be a file path, not a boolean. This was causing failures on CI release/2.0 configs which still use dotnet-test instead of dotnet-vstest.

To avoid this conflict, let's use VSTestVerboseOutput instead.

cref https://github.com/Microsoft/vstest/blob/123adc3ebfdc05be24ee6f44ace1e9eef79e372f/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets#L33-L48

cc @jbagga @ryanbrandenburg

### DIFF
--- a/files/KoreBuild/modules/vstest/module.targets
+++ b/files/KoreBuild/modules/vstest/module.targets
@@ -64,9 +64,9 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
 
   <!-- The 'inner-build' -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


ryanbrandenburg:
We'll have to update the usage on the CI too.

natemcmaster:
I'll update aspnet/Templating to use this new prop

PR #559, opened 2018-01-25
Update the version of Microsoft.DotNet.Build.Tasks.Feed

pranavkm:


natemcmaster:
As a side note, let's look into only downloading this package in Microsoft.DotNet.Build.Tasks.Feed. This task adds a few MB of bloat to KoreBuild, and its only used in aspnet/Universe.

pranavkm:
> As a side note, let's look into only downloading this package in Microsoft.DotNet.Build.Tasks.Feed. 

You lost me here.

natemcmaster:
We bundle Microsoft.DotNet.Build.Tasks.Feed in korebuild.zip, but we probably don't need to. We could change aspnet/Universe to download Microsoft.DotNet.Build.Tasks.Feed on its own.
 
i.e https://aspnetcore.blob.core.windows.net/buildtools/korebuild/artifacts/2.1.0-preview1-1001/korebuild.2.1.0-preview1-1001.zip is 8.7 MB. and 5.3 of that is Microsoft.DotNet.Build.Tasks.Feed.

pranavkm:
👍  I'll move this in to Universe.

natemcmaster:
I'm okay to merge this PR as is and move to Universe later. I was just pointing out that this is something we can do.

PR #563, opened 2018-01-29
Add target for setting the teamcity build number to match the product version

natemcmaster:
For now, I'm not automatically chaining this to anything. There are enough complications in how we use build number and chain build numbers together that this would be a breaking change. To opt-in to this, our CI configs can run `build.cmd /t:SetTeamCityBuildNumberToVersion /t:Build` or chain it into their build via repo.pros/targets


pakrym:
What are the next steps? is there a plan to start using this?

natemcmaster:
I was planning to update aspnet/Universe to use this target

PR #565, opened 2018-01-30
Use TeamCity x-plat logger (take 2)

natemcmaster:
Same implementation as https://github.com/aspnet/BuildTools/pull/550, but with a workaround for https://github.com/JetBrains/TeamCity.MSBuild.Logger/issues/3 - adding a simple task that calls Console.WriteLine instead of Log.LogMessage.

Resolves https://github.com/aspnet/BuildTools/issues/345

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,14 +6,14 @@
 root = true
 
 [*]
-indent_style = space
-charset = utf-8
+indent_style             = space
+charset                  = utf-8
 trim_trailing_whitespace = true
-insert_final_newline = true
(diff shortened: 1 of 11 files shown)


ryanbrandenburg:
Looks good to me.

pakrym:
Would it propagate from inner invocations inside universe?

natemcmaster:
> Would it propagate from inner invocations inside universe?

Yes, but the formatting will look funny until we lift repo builds into the same MSBuild process

PR #568, opened 2018-02-01
Include portable .pdb files in regular nupkgs when Internal.AspNetCore.Sdk is used

natemcmaster:
Allows adding .pdb files in the .nupkg for aspnetcore packages that end up on NuGet.org.

cref https://github.com/aspnet/Universe/issues/131

cc @Eilon @JunTaoLuo @ctaggart

### DIFF
--- a/src/Internal.AspNetCore.Sdk/build/Common.targets
+++ b/src/Internal.AspNetCore.Sdk/build/Common.targets
@@ -21,8 +21,9 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
   </PropertyGroup>
 
   <Target Name="_EnsurePdbsInNupkgArePortable" BeforeTargets="GenerateNuspec" Condition=" '$(IsPackable)' != 'false' ">
-    <Error Text="Symbols included in a nupkg package should be portable. Set DebugType=portable or IncludeSymbols=false."
-      Condition=" '$(DebugType)' != 'portable' AND '$(IncludeSymbols)' == 'true' " />
+    <Error Condition=" '$(DebugType)' != 'portable' AND '$(IncludeSymbols)' == 'true' "
+      Text="Symbols were set to be included but the project was built using DebugType = $(DebugType).
+            Set DebugType = portable to include portable pdbs required for symbol publishing or stop including symbols by setting IncludeSymbols=false." />
   </Target>
 
   <Target Name="_ShowBuildVersion" BeforeTargets="PrepareForBuild">
--- a/src/Internal.AspNetCore.Sdk/build/Common.targets
+++ b/src/Internal.AspNetCore.Sdk/build/Common.targets
@@ -15,6 +15,16 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
     <IncludeSymbols Condition="'$(NuspecFile)'!=''">false</IncludeSymbols>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(IncludeSymbols)' != 'false' ">
+    <!-- Include .pdb files in regular .nupkg files. -->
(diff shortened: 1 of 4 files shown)


src/Internal.AspNetCore.Sdk/build/Common.targets
```diff
@@ -15,6 +15,16 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
     <IncludeSymbols Condition="'$(NuspecFile)'!=''">false</IncludeSymbols>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(IncludeSymbols)' != 'false' ">
+    <!-- Include .pdb files in regular .nupkg files. -->
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+  </PropertyGroup>
+
+  <Target Name="_EnsurePdbsInNupkgArePortable" BeforeTargets="GenerateNuspec" Condition=" '$(IsPackable)' != 'false' ">
+    <Error Text="Symbols included in a nupkg package should be portable. Set DebugType=portable or IncludeSymbols=false."
```
pranavkm: Set DebugType=portable ~~or~~and IncludeSymbols=false

src/Internal.AspNetCore.Sdk/build/Common.targets
```diff
@@ -15,6 +15,16 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
...
+      Condition=" '$(DebugType)' != 'portable' AND '$(IncludeSymbols)' == 'true' " />
```
pranavkm: Btw, the last I looked, symbols for Razor's net46 binaries are `full`. Can you verify this works for the repo?

src/Internal.AspNetCore.Sdk/build/Common.targets
natemcmaster: Maybe I should clarify. What I meant here was `Fix by setting DebugType=portable or disable inclusion of symbols by setting IncludeSymbols=false`

src/Internal.AspNetCore.Sdk/build/Common.targets
natemcmaster: Why are they `full`? Don't we need `portable` for these to work for x-plat?

pranavkm: I'm not entirely sure. @NTaylorMullen might know more.

natemcmaster: Looks like it's due to this, right? https://github.com/aspnet/Razor/issues/1548 If we need `full` for VS, let's still produce `portable` by default. We can convert `portable` symbols to `full` in our VS publishing steps. I believe @muratg knows how to do that.

src/Internal.AspNetCore.Sdk/build/Common.targets
pranavkm: Maybe split this in to two? 

> Symbols were set to be included but the project was built using `DebugType = full`. Set `DebugType = portable` to include portable pdbs required for symbol publishing or stop including symbols by setting `IncludeSymbols=false`

natemcmaster: that's much better. I'll update the error.

src/Internal.AspNetCore.Sdk/build/Common.targets
pranavkm: Ahh, yeah. Completely forgot I made that change 👍

PR #570, opened 2018-02-05
Don't include symbols in the regular .nupkg file but still ensure symbols are portable

natemcmaster:
Undo part of #568. Don't include .pdb files in the regular .nupkg, but still ensure DebugType == portable.

cref https://github.com/aspnet/Universe/issues/131

cc @Eilon

### DIFF
--- a/test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
+++ b/test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
@@ -51,8 +51,7 @@ namespace KoreBuild.FunctionalTests
 
             using (var reader = new PackageArchiveReader(libPackage))
             {
-                Assert.Contains("lib/netstandard2.0/Simple.Lib.pdb", reader.GetFiles());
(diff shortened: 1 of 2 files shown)


test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
```diff
@@ -51,8 +51,7 @@ public SimpleRepoTests(ITestOutputHelper output, RepoTestFixture fixture)
 
             using (var reader = new PackageArchiveReader(libPackage))
             {
-                Assert.Contains("lib/netstandard2.0/Simple.Lib.pdb", reader.GetFiles());
-                Assert.Contains("lib/net461/Simple.Lib.pdb", reader.GetFiles());
+                Assert.Empty(reader.GetFiles().Where(p => Path.GetExtension(p).Equals(".pdb", StringComparison.OrdinalIgnoreCase)));
```
Eilon: My usual comment: always call static `string.Equals` because it properly guards against `null`. Not sure what GetExtension returns (nor do I trust it!).

PR #572, opened 2018-02-07
Ensure Run.ExitCode is non-zero when it fails to launch a process

natemcmaster:
Found this while working on updating publishing in aspnet/Universe. NPM failed to launch, but the ExitCode was still 0, so the MSBuild targets didn't thing there was an error.

### DIFF
--- a/test/BuildTools.Tasks.Tests/RunTaskTests.cs
+++ b/test/BuildTools.Tasks.Tests/RunTaskTests.cs
@@ -1,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.AspNetCore.BuildTools;
+using Xunit;
+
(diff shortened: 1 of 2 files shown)


test/BuildTools.Tasks.Tests/RunTaskTests.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class RunTaskTests
+    {
+        [Fact]
+        public void ExitCodeIsNonZeroIfFailedToStart()
+        {
+            var engine = new MockEngine { ContinueOnError = true };
+            var task = new Run
+            {
+                FileName = "sdfkjskldfsjdflkajsdas",
```
Eilon: Maybe add a check to ensure this task doesn't exist on disk already?
<br />
<br />
<br />
<br />
<br />
<br />
<br />
:trollface:

natemcmaster: haha, I was thinking of writing a console tool named sdfkjskldfsjdflkajsdas, but settled on dsfdflkajsdaskjskldfsj instead.

PR #573, opened 2018-02-07
Fix console output

ryanbrandenburg:


ryanbrandenburg:
@rynowak this should fix the test output problem you noticed.

ryanbrandenburg:
This passed locally, merging quickly to unbreak people.

PR #575, opened 2018-02-12
Fix additional deps trimming bugs

pakrym:


src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
```diff
@@ -14,14 +14,6 @@
       <_BasePackagePath>content\additionaldeps\</_BasePackagePath>
       <_RuntimeStoreManifestFile>$(_DepsOutputDirectory)\rs.csproj</_RuntimeStoreManifestFile>
       <_RuntimeStoreOutput>$(_DepsOutputDirectory)\rs\</_RuntimeStoreOutput>
-      <_AdditionalMsBuildProperties>
```
pakrym: This should not be required because parent MSBuild task invocation with Parameters made these global.

src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
```diff
@@ -75,7 +66,6 @@
       <Content Include="$(_RuntimeStoreOutput)\**\*" PackagePath="content\store\%(RecursiveDir)" />
 
       <Content Include="%(_HostingStartupPackageReference.TrimmedDepsFile)" PackagePath="%(_HostingStartupPackageReference.PackagePath)" />
-      <Content Include="%(_HostingStartupPackageReference.DllPath)" PackagePath="$(_BasePackagePath)" />
```
pakrym: Not required, dll is copied during runtime store build

PR #578, opened 2018-02-13
[release/2.1] Updating SDK to 2.1.300-preview1-008123

### LINKED ISSUE #562: Update SDK version to get "GetNativeManifest" fix
The version of the SDK we currently have is plagued by Microsoft/msbuild#2922. Once there's an SDK which has that fix we should take it so we can start building BuildTools again.

We also need to re-enable the "Update BuildTools" step of Update Universe.

CC @natemcmaster

natemcmaster:
Resolves #562 

cc @rynowak - for real this time, it includes the msbuild fix, so this should unblock your razor work.


rynowak:
Do we need to trigger a build of this spicy boy? Updating korebuild gets me version: 2.1.0-preview1-1015 which uses 2.2.0-preview1-008003

natemcmaster:
Yeah, build was failing cuz myget. Try again. Build tools 2.1.0-preview1-1023 should have the fix

PR #582, opened 2018-02-15
Enable crossgen for siteextension runtime store

pakrym:


pakrym:
We'll try it on Azure when it builds. Builds fine locally.

PR #587, opened 2018-02-20
Use SHA256 in PDBs

natemcmaster:
Use a new feature of the compiler. Cref https://github.com/dotnet/roslyn/pull/24820

cc @tmat


natemcmaster:
Just forwarded you email for context on the change.

tmat:
@ryanbrandenburg Yes. SHA1 needs to be phased out - it's not considered an acceptable hash algorithm (regardless of whether used for security or not).

tmat:
@natemcmaster Thanks!

PR #588, opened 2018-02-20
Update sdk to 2.1.300-preview2-008212

pranavkm:


pranavkm:
There's changes to the Razor Sdk that we want for some test cleanup in Razor. There's a couple more rounds of new Sdk updates before we finalize the Sdk.

PR #591, opened 2018-02-27
Add lzma archive task package to BuildTools

JunTaoLuo:
Addresses https://github.com/aspnet/Universe/issues/595.

Adding tools for LZMA creation to BuildTools using the new task package format. I have tested the functionality with a sample project at https://github.com/JunTaoLuo/TestTaskPackages and it works.


src/Microsoft.AspNetCore.BuildTools.CreateLZMA/sdk/Sdk.Targets
```diff
@@ -0,0 +1,2 @@
+<Project>
```
JunTaoLuo: Without this empty file, the SDK can't be used by the test project.

BuildTools.sln
```diff
@@ -92,6 +92,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Internal.AspNetCore.SiteExt
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NuGetPackageVerifier.Tests", "test\NuGetPackageVerifier.Tests\NuGetPackageVerifier.Tests.csproj", "{439CC7A3-F6E6-46B8-B6A0-05E22E558FC2}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.BuildTools.CreateLZMA", "src\Microsoft.AspNetCore.BuildTools.CreateLZMA\Microsoft.AspNetCore.BuildTools.CreateLZMA.csproj", "{27679B88-FED9-404A-8A41-02BB31CF3019}"
```
Eilon: Can we call this `Internal.AspNetCore.Xyz` to match other stuff? I want to make it as clear as possible this isn't a product feature.

ryanbrandenburg:
Why have this in BuildTools instead of Universe? Will it ever be used outside of Universe?

src/Microsoft.AspNetCore.BuildTools.CreateLZMA/CreateLZMA.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.DotNet.Archive;
+using MSBuildTask = Microsoft.Build.Utilities.Task;
+
+namespace Microsoft.AspNetCore.BuildTools.LZMA
+{
+    public class CreateLZMA : MSBuildTask
+    {
+        [Required]
+        public string OutputArchiveFilePath { get; set; }
```
natemcmaster: Can we make the API for this task align with the patterns we've established for other tasks? We did an API review and came up with this for zips: https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.Build.Tasks.IO#ziparchive. (Snippet copied below)

-----------

### `ZipArchive`

Creates a .zip archive file

#### Common parameters
Task parameter     | Type    | Description
-------------------|---------|---------------------------------------------------------------
OutputPath         | string  | **[Required]**  The path where the zip file should be created. The containing directory will be created if it doesn't already exist.
Overwrite          | boolean | Overwrite output path if it exists

There are two valid usages: a list of files, or an entire directory

#### Parameter set - files

Task parameter     | Type        | Description
-------------------|-------------|----------------------------------------------------------------------------------------------------------------------------------
SourceFiles        | ITaskItem[] | **[Required]** Files to be included in the zip. <br> The `Link` metadata item can be set to explicitly set the zip entry path.
BaseDirectory      | string      | **[Required]** The directory to use as the base directory. The entry path for each item in SourceFiles is relative to this.

```xml
<ItemGroup>
   <JsFiles Include="$(PublishDir)**\*.js" >
   <JsFiles Include="../webpack.js" Link="webpack.js" >
</ItemGroup>

<ZipArchive OutputPath="myapp.js.zip" SourceFiles="@(JsFiles)" BaseDirectory="$(PublishDir)" />
```


#### Parameter set - directory

Task parameter     | Type    | Description
-------------------|---------|-------------
SourceDirectory    | string  | **[Required]** Creates a zip for an entire directory.

Example:
```xml
<ZipArchive OutputPath="myapp.zip" SourceDirectory="$(PublishDir)" />
```

BuildTools.sln
natemcmaster: +1. How about Internal.AspNetCore.Tasks.Lzma for now. I'd avoid naming it CreateLZMA for now because we may want to add more tasks in the future, such as extracting an LZMA.

src/Microsoft.AspNetCore.BuildTools.CreateLZMA/Microsoft.AspnetCore.BuildTools.CreateLZMA.nuspec
```diff
@@ -0,0 +1,19 @@
...
+    <authors>Microsoft</authors>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>$description$</description>
+    <copyright>$copyright$</copyright>
+    <packageTypes>
+      <packageType name="MSBuildSdk" />
+    </packageTypes>
+  </metadata>
+  <files>
+    <file src="bin\$config$\netstandard1.5\$id$.dll" target="build\netstandard1.5\$id$.dll" />
```
natemcmaster: You will also need to copy Microsoft.DotNet.Archive.dll into the package next to the task assembly.

src/Microsoft.AspNetCore.BuildTools.CreateLZMA/sdk/Sdk.Targets
natemcmaster: Add this comment to source code

src/Microsoft.AspNetCore.BuildTools.CreateLZMA/sdk/Sdk.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+  <UsingTask
+    AssemblyFile="$(MSBuildThisFileDirectory)..\build\netstandard1.5\Microsoft.AspNetCore.BuildTools.CreateLZMA.dll"
```
natemcmaster: Need to vary the task assembly loaded based on `$(MSBuildRuntimeType)`.

Example: https://github.com/aspnet/BuildTools/blob/b56c36c46122915b86d56836e089d7e6b57838a8/src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.props#L6-L12

JunTaoLuo:
Will convert this into just a repo task since it's only used in Universe.

PR #595, opened 2018-03-01
Remove bill of materials tasks

natemcmaster:
In the end, this build manifest output didn't prove to be useful enough to keep. The .NET Core engineering team created something similar for  the
Product Construction Plan which was sufficient, so we decided to eliminate this altogether in favor of consolidating build infrastructure.

Resolves https://github.com/aspnet/Universe/issues/679

cc @muratg


pranavkm:
Guessing this is unused in Universe? 🚢 🇮🇹

natemcmaster:
Yeah, Universe set `<GenerateBillOfMaterials>false</GenerateBillOfMaterials>` from the beginning because we didn't finish implementing a bill of materials merge feature to combine manifests from all repo builds.

natemcmaster:
> What are the missing pieces in prodcon manifest that ours used to have?

We never fully finished producing the bill of materials for a universe build....so nothing.

PR #596, opened 2018-03-01
Workaround new hostfxr bug

pakrym:
Can't do it in msbuild DOTNET_HOME is empty most of the time.


pakrym:
https://github.com/aspnet/BuildTools/issues/597

Issue #597, opened 2018-03-01
Remove hostfxr workaround after getting fixed version


pakrym:
https://github.com/aspnet/BuildTools/pull/596

natemcmaster:
Is there a bug filed on core-setup for this? I just ran into more issues with the fxr running benchmarks and tests

pakrym:
@halter73  did too, https://github.com/dotnet/corefx/issues/27571

PR #602, opened 2018-03-08
Revert "Workaround new hostfxr bug"

pakrym:
https://github.com/aspnet/BuildTools/issues/597

Reverts aspnet/BuildTools#596


natemcmaster:
Is there a corresponding SDK update that should happen too?

pakrym:
@natemcmaster  as long as new runtime brings newer version of hostfxr with the fix I don't think we care about SDK.

PR #603, opened 2018-03-08
Fail faster when the artifacts directory isn't clean

natemcmaster:
It's annoying that this check doesn't happen until after the compilation is done. This moves the check before build so you get a reminder faster that you need to clean when switching branches.

### DIFF
--- a/files/KoreBuild/modules/sharedsources/sharedsources.csproj
+++ b/files/KoreBuild/modules/sharedsources/sharedsources.csproj
@@ -42,11 +42,11 @@
   <ItemGroup Condition="'$(NuspecBasePath)'!=''">
     <Compile Include="$(NuspecBasePath)**\*.cs" Exclude="$(DefaultExcludeItems)">
       <Pack>true</Pack>
-      <PackagePath>$(ContentTargetFolders)\cs\netstandard1.0\</PackagePath>
+      <PackagePath>$(ContentTargetFolders)\cs\netstandard1.0\$(PackageId)\%(RecursiveDir)\</PackagePath>
     </Compile>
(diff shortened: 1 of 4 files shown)


files/KoreBuild/modules/sharedsources/sharedsources.csproj
```diff
@@ -42,11 +42,11 @@
   <ItemGroup Condition="'$(NuspecBasePath)'!=''">
     <Compile Include="$(NuspecBasePath)**\*.cs" Exclude="$(DefaultExcludeItems)">
       <Pack>true</Pack>
-      <PackagePath>$(ContentTargetFolders)\cs\netstandard1.0\</PackagePath>
+      <PackagePath>$(ContentTargetFolders)\cs\netstandard1.0\$(PackageId)\%(RecursiveDir)\</PackagePath>
     </Compile>
     <EmbeddedResource Include="$(NuspecBasePath)**\*.resx" Exclude="$(DefaultExcludeItems)">
       <Pack>true</Pack>
-      <PackagePath>$(ContentTargetFolders)\any\any\</PackagePath>
+      <PackagePath>$(ContentTargetFolders)\any\any\$(PackageId)\%(RecursiveDir)\</PackagePath>
```
halter73: What's this about? I don't see where PackagePath is used. Is this now basically PackageContentPath*s*?

natemcmaster: This is just some Solution Explorer sugar. It will make the make share source files appear in a folder that matches the .Sources package ID.

halter73: That's pretty cool.

PR #616, opened 2018-03-21
Fix early termination of PackageVersionMatchesAssemblyVersionRule

natemcmaster:
The rule is missing some errors in ASSEMBLY_FILE_VERSION_MISMATCH and ASSEMBLY_VERSION_MISMATCH.

### DIFF
--- a/modules/NuGetPackageVerifier/console/Rules/PackageVersionMatchesAssemblyVersionRule.cs
+++ b/modules/NuGetPackageVerifier/console/Rules/PackageVersionMatchesAssemblyVersionRule.cs
@@ -23,23 +23,14 @@ namespace NuGetPackageVerifier.Rules
 
                 var infoVersion = assemblyInformationalVersionAttribute?.ConstructorArguments[0].Value?.ToString();
 
-                if (!NuGetVersion.TryParse(infoVersion, out var assemblyInformationalNuGetVersion))
+                if (!NuGetVersion.TryParse(infoVersion, out var assemblyInformationalNuGetVersion) ||
+                    !VersionEquals(context.Metadata.Version, assemblyInformationalNuGetVersion))
                 {
                     yield return PackageIssueFactory.AssemblyInformationalVersionDoesNotMatchPackageVersion(
                        assemblyData.Key,
                        infoVersion,
                        context.Metadata.Version,
                        context.Metadata.Id);
-                    continue;
-                }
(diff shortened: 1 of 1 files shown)


modules/NuGetPackageVerifier/console/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -30,10 +30,8 @@ public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context
                        infoVersion,
                        context.Metadata.Version,
                        context.Metadata.Id);
-                    continue;
                 }
-
-                if (!VersionEquals(context.Metadata.Version, assemblyInformationalNuGetVersion))
+                else if (!VersionEquals(context.Metadata.Version, assemblyInformationalNuGetVersion))
```
kichalla: This should still stay `if`? (like if both this and the above condition fail). Would be good add a test for this (if there are tests already for these).

ryanbrandenburg: Hmm, I was thinking "Oh it will just give two of the same error message", but if there's two different ways it will fail I think I agree with @kichalla that it should fail all of them up-front.

natemcmaster: Updated

modules/NuGetPackageVerifier/console/Rules/PackageVersionMatchesAssemblyVersionRule.cs
```diff
@@ -23,23 +23,14 @@ public IEnumerable<PackageVerifierIssue> Validate(PackageAnalysisContext context
 
                 var infoVersion = assemblyInformationalVersionAttribute?.ConstructorArguments[0].Value?.ToString();
 
-                if (!NuGetVersion.TryParse(infoVersion, out var assemblyInformationalNuGetVersion))
+                if (!NuGetVersion.TryParse(infoVersion, out var assemblyInformationalNuGetVersion)
+                    || !VersionEquals(context.Metadata.Version, assemblyInformationalNuGetVersion))
```
pranavkm: Nit: `||` goes on the previous line.

PR #619, opened 2018-03-22
Unify to RepositoryCommit as the property for storing the commit hash

natemcmaster:
`dotnet pack` on .nuspec files because we have started using $(RepositoryCommit) everwhere, but Internal.AspNetCore.Sdk was only setting the CommitHash property.

workaround for https://github.com/NuGet/Home/issues/6722


natemcmaster:
Validated manually on templating\src\Microsoft.DotNet.Web.Client.ItemTemplates that `dotnet pack` works now.

natemcmaster:
Latest build tools is available now. Run `build.cmd -u` to get it. Let me know if you have any more issues.

PR #628, opened 2018-04-03
Update dependencies.props from other dependencies.props  file

pakrym:


pakrym:
Had to change the code to evaluate the props file passed in so we can get `NETCoreApp21PackageVersion` from universe `dependencies.props`

Issue #623, opened 2018-03-27
VSTest output not displayed on console output on Windows


natemcmaster:
After #621 , I'm not seeing console output from vstest

![image](https://user-images.githubusercontent.com/2696087/37999089-2259407a-31d6-11e8-99e5-d187e966caf7.png)

Doesn't appear to affect AppVeyor (Windows) or macOS/Linux though
![image](https://user-images.githubusercontent.com/2696087/37999166-7bbca418-31d6-11e8-8661-7c61daf2bc49.png)

natemcmaster:
Result of conversation with MSBuild guys:

This is basically the same problem as https://github.com/Microsoft/vstest/issues/1503. We are doing things with MSBuild that are apparently bad to do. We got away with it until now because MSBuild for .NET Core didn't support node reuse. Now that MSBuild is shipping node reuse for core, the bad things we did have bitten us. We can workaround for now with `$env:MSBUILDENSURESTDOUTFORTASKPROCESSES=1`, but this workaround may go away at some point. We should try to do the Right Thing so we can enable node reuse in the future.

ryanbrandenburg:
I'm un-blocking us by doing #TheWrongThing in #627, but I agree with @natemcmaster that we should change to doing the right thing once it's not blocking everyone else, so I propose leaving this issue open and just moving it to the RC1 milestone once my workaround merges.

For future reference, I believe that #TheRightThing is to redirect stdout and stderr in `RunBase` and attach those streams to events which write to MSBuild. It's my understanding that doing that will lose us some coloring and possibly cause difficulty with TC messages, which is why we don't want to do it right now.

Eilon:
https://github.com/aspnet/BuildTools/pull/629

Eilon:
Oops, bad copy/paste. Meant to say: @ryanbrandenburg - is there any further action required for this issue, or does https://github.com/aspnet/BuildTools/pull/629 cover it?

ryanbrandenburg:
Yes, #629 is #TheRightThing.

PR #629, opened 2018-04-04
Redirect output from tests

### LINKED ISSUE #624: Updating to new SDK causes TeamCity to not see test result files
After updating to 2.1.300-preview2-008407 #621, TeamCity builds started failing because it could not find .trx files. I had to revert because of this.

image

Possible related to #623 

FYI @jbagga @ajaybhargavb @pranavkm

ryanbrandenburg:
Tested against the BuildTools repo with and without error output, it seems to do the job (though we lose colorization as we expected).

Fixes #624 and #623.

### DIFF
--- a/modules/BuildTools.Tasks/Run.cs
+++ b/modules/BuildTools.Tasks/Run.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
 
 namespace Microsoft.AspNetCore.BuildTools
 {
@@ -24,6 +25,11 @@ namespace Microsoft.AspNetCore.BuildTools
         [Required]
         public string FileName { get; set; }
 
-        protected override string GetExecutable() => FileName;
+        protected override string ToolName => FileName;
+
+        protected override string GenerateFullPathToTool()
+        {
+            return FileName;
+        }
     }
 }
--- a/modules/BuildTools.Tasks/RunBase.cs
+++ b/modules/BuildTools.Tasks/RunBase.cs
@@ -2,11 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Runtime.InteropServices;
 using Microsoft.Build.Utilities;
 using Microsoft.Build.Framework;
 using Microsoft.Extensions.CommandLineUtils;
@@ -16,13 +13,11 @@ namespace Microsoft.AspNetCore.BuildTools
     /// <summary>
     /// A task that runs a process without piping output into the logger.
     /// </summary>
-    public abstract class RunBase : Task
+    public abstract class RunBase : ToolTask
     {
         private static readonly char[] EqualsArray = new[] { '=' };
 
         private const int OK = 0;
-
-        protected abstract string GetExecutable();
 
         /// <summary>
         /// A list of arguments to be passed to the executable. The task will escape them for spaces and quotes.
@@ -35,14 +30,6 @@ namespace Microsoft.AspNetCore.BuildTools
         /// Cannot be used with <see cref="Arguments"/>
         /// </summary>
         public string Command { get; set; }
-
-        /// <summary>
-        /// Environment variables to set on the process.
-        /// </summary>
-        /// <remarks>
-        /// The item spec will split on '='. Alternatively, it will look for the metadata name "Value".
-        /// </remarks>
-        public ITaskItem[] EnvironmentVariables { get; set; }
 
         // Additional options
         /// <summary>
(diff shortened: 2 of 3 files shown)


modules/BuildTools.Tasks/RunBase.cs
```diff
@@ -156,6 +159,29 @@ public override bool Execute()
...
+                process.ErrorDataReceived += (object sender, System.Diagnostics.DataReceivedEventArgs e) =>
+                {
+                    if(e.Data != null)
+                    {
+                        Log.LogError(e.Data);
+                    }
+                };
+                process.BeginErrorReadLine();
+
+                if(!process.WaitForExit(Int32.MaxValue))
```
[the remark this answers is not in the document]
ryanbrandenburg: This is following @natemcmaster's example [here](https://github.com/aspnet/DotNetTools/blob/5bec83c5c92a6df6adcaa2a6363d3857b2c35c1c/src/dotnet-watch/Internal/ProcessRunner.cs#L110). I'll add his explanatory comment for future confused people.

ryanbrandenburg:
:up::date:

modules/BuildTools.Tasks/RunBase.cs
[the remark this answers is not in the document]
Eilon: Weeeeeeeeeeird. But at least use `int.MaxValue` 😄

natemcmaster:
At this point, I'm not sure there is much value in managing the `Process` object ourselves. MSBuild provides an abstract class named `ToolTask`. All you have to implement are methods that provide the path to an executable + args to pass in. The original point of the `Run` and `RunDotNet` tasks was to solve the problem of argument escaping and finding the dotnet.exe file. If we use `ToolTask`, it will take care of everything else for us: redirecting output, managing cancellation, process disposal, logging levels, distributed loggers, etc.

See for examples [here](https://github.com/aspnet/BuildTools/blob/3e0fb6dc5243f9d7c5756c91c71ff3a30099b61d/src/ApiCheck.Task/ApiCheckTasksBase.cs) and [here](https://github.com/natemcmaster/Yarn.MSBuild/blob/master/src/Yarn.MSBuild/Yarn.cs)

modules/BuildTools.Tasks/RunDotNet.cs
```diff
@@ -16,9 +18,35 @@ public class RunDotNet
...
+
+        protected override string GenerateFullPathToTool()
+        {
+            return GetExecutable();
+        }
+
+        protected override string GenerateCommandLineCommands()
+        {
+            var arguments = string.Empty;
+            arguments += string.Join(" ", Arguments.Select(i => i.ItemSpec));
```
natemcmaster: Use ArgumentEscaper.EscapeAndConcatenate. String.Join(" ") doesn't work for arguments containing spaces or quotes.

files/KoreBuild/modules/vstest/module.targets
```diff
@@ -98,7 +98,7 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
 
     <Message Text="%0AStarting test group: $(TestGroupName)/$(TargetFramework)" Importance="High" />
 
-    <RunDotNet Arguments="@(VSTestArgs)" IgnoreExitCode="true">
```
natemcmaster: We still need IgnoreExitCode=true set. Otherwise, the task will exit before we can report the TRX file location to TeamCity.

modules/BuildTools.Tasks/Run.cs
```diff
@@ -16,14 +16,25 @@ public class Run
 #else
 #error This must be built either for an SDK or for BuildTools
 #endif
-        : RunBase
```
natemcmaster: This change drops some essential API such as `Arguments` and `Command`. It may be easier to instead change `RunBase` to inherit from `ToolTask`. There will be some overlap, such as EnvironmentVariables, and in those cases, you can remove the RunBase version.

ryanbrandenburg:
:up::date:

modules/BuildTools.Tasks/RunBase.cs
```diff
@@ -65,15 +51,24 @@ public abstract class RunBase : Task
         /// </summary>
         public bool UseShellExecute { get; set; }
```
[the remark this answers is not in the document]
ryanbrandenburg: npm publish uses MaxRetries and UseShellExecute, I assume because it's unreliable. Don't know why it's shell execute though.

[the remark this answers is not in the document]
natemcmaster: It uses shell execute because npm is actually npm.cmd. But this isn't an issue for ToolTask. It's already handled the complexity of launching/killing cmd.exe.

modules/BuildTools.Tasks/RunBase.cs
```diff
@@ -0,0 +1,128 @@
...
+        public int MaxRetries { get; set; }
+
+        /// <summary>
+        /// Set the value for UseShellExecute on the new process.
+        /// </summary>
+        public bool UseShellExecute { get; set; }
+
+        protected override bool HandleTaskExecutionErrors()
+        {
+            return IgnoreExitCode
```
Eilon: This looks like an `||` to me...

modules/BuildTools.Tasks/RunBase.cs
```diff
@@ -100,8 +88,28 @@ public override bool Execute()
                 return false;
             }
 
-            var arguments = string.Empty;
+            return base.ValidateParameters();
+        }
+
+        public override bool Execute()
+        {
+            var retries = MaxRetries == 0 ? 1 : MaxRetries;
```
Eilon: If you want: `Math.Max(1, MaxRetries)`

modules/BuildTools.Tasks/RunBase.cs
```diff
@@ -100,8 +88,28 @@ public override bool Execute()
...
+            for(int i = 0; i < retries; i++)
```
Eilon: Format document.

PR #630, opened 2018-04-04
Require SymbolsNuspecFile if a NuspecFile is specified

pranavkm:


src/Internal.AspNetCore.Sdk/build/Common.targets
```diff
@@ -10,16 +10,70 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
...
+
+    <!-- workaround https://github.com/NuGet/Home/issues/4726 -->
+    <IncludeSymbols Condition="'$(NuspecFile)'!=''">false</IncludeSymbols>
+    <IncludeSource Condition="'$(NuspecFile)'!=''">false</IncludeSource>
+
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);_EnsureDebugTypeIsPortable</TargetsForTfmSpecificBuildOutput>
   </PropertyGroup>
 
+  <Target Name="_GenerateSymbolsNupkg" >
+    <Error Condition="'$(NuspecFile)' != '' AND '$(SymbolsNuspecFile)' == ''"
```
pranavkm: Thoughts? We want to ensure we have symbols when we have a nuspec and we want to avoid putting pdbs in the main nupkg. This moves the symbol generation \ verifcation in to the Sdk.

pranavkm:
Ping

src/Internal.AspNetCore.Sdk/build/Common.targets
natemcmaster: Implementation looks fine, but I'm concerned about the longevity of this solution. The `PackTask` is "pubternal" API. Even though it's accessible, I've been told not to use it directly as the NuGet team regularly make changes to this task and targets.

pranavkm: Aww. We could write a task to call `PackageBuilder` ourselves, Cleans up some of the copying stuff

natemcmaster: We have the PackNuspec task in KoreBuild that uses PackageBuilder. This may become part of Arcade https://github.com/dotnet/arcade/pull/57, or part of NuGet itself if I had my wishes. But, it's probably easier to go with this for now. I'm okay with this change as long as you're aware it's a thing we'll likely have to service.

pranavkm: Let me have a look at `PackNuspec`. `PackTask` has some quirky behavior that would be nice to avoid.

ajaybhargavb:
Updated.

src/Internal.AspNetCore.Sdk/build/Common.targets
```diff
@@ -10,16 +10,72 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
...
+      Text="NuspecFile was specified but SymbolsNuspecFile was not specified."
+      Code="KRB5006" />
+
+    <ConvertToAbsolutePath Paths="$(SymbolsNuspecFile)">
+      <Output TaskParameter="AbsolutePaths" PropertyName="SymbolNuspecFileAbsolutePath" />
+    </ConvertToAbsolutePath>
+
+    <ConvertToAbsolutePath Paths="$(PackageOutputPath)symbols-nupkg\">
```
natemcmaster: I think it would be better to make this `$(IntermediateOutputPath)symbols-nupkg\`.

src/Internal.AspNetCore.Sdk/build/Common.targets
```diff
@@ -10,16 +10,72 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
...
+              NuspecOutputPath="$(SymbolNuspecFileAbsolutePath)"
+              IncludeBuildOutput="false"
+              RestoreOutputPath="$(RestoreOutputAbsolutePath)"
+              NuspecFile="$(SymbolNuspecFileAbsolutePath)"
+              NuspecBasePath="$(NuspecBasePath)"
+              NuspecProperties="$(NuspecProperties)"
+              AllowedOutputExtensionsInPackageBuildOutputFolder="$(AllowedOutputExtensionsInPackageBuildOutputFolder)"
+              AllowedOutputExtensionsInSymbolsPackageBuildOutputFolder="$(AllowedOutputExtensionsInSymbolsPackageBuildOutputFolder)"/>
+
+    <Copy
```
natemcmaster: Maybe use `<Move>` instead? This task doesn't use incrementalism, so there doesn't appear to be an advantage to `<Copy>`.

src/Internal.AspNetCore.Sdk/build/Common.targets
```diff
@@ -10,16 +10,72 @@ For single-tfm projects, this will be imported from build/Internal.AspNetCore.Sd
...
+      SourceFiles="$(SymbolsPackageOutputAbsolutePath)$(PackageId).$(PackageVersion).nupkg"
+      DestinationFiles="$(PackageOutputAbsolutePath)$(PackageId).$(PackageVersion).symbols.nupkg"
+      SkipUnchangedFiles="true" />
+
+    <RemoveDir Directories="$(SymbolsPackageOutputAbsolutePath);" />
```
natemcmaster: This will be racy. Could delete things currently being packed or copied on other, parallel builds.

ajaybhargavb: Agreed. I'll just use the Move task and put it Obj folder

ajaybhargavb:
Updated.

ajaybhargavb:
Closing as we don't need this anymore.

PR #632, opened 2018-04-08
Typo Fix

robrohan:
Typo: 'used to' vs. 'use to'


Eilon:
The current phrasing is correct. It's saying the option is the way "it used to be" with another tool.

robrohan:
Not an English major myself, and no one is going to be confused what you mean, but I think "use to" is actually correct: http://www.5minuteenglish.com/mar20.htm

PR #642, opened 2018-04-17
Prkrishn/2.2.0

pranavkm:


version.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
     <KoreBuildChannel>dev</KoreBuildChannel>
-    <VersionPrefix>2.1.0</VersionPrefix>
-    <VersionSuffix>preview3</VersionSuffix>
+    <VersionPrefix>2.2.0</VersionPrefix>
+    <VersionSuffix>preview1</VersionSuffix>
```
natemcmaster: Let's go with "alpha1" for now. Gives us more flexibility in the future. That's what we're doing for the runtime too, right?

pranavkm: Well, we went with preview1 there: https://github.com/aspnet/Mvc/blob/dev/version.props#L4, so it might make sense to be consistent.

natemcmaster: Oh. That's not what we discussed last week. Was there a thread on this that I missed?

pranavkm: cc @ryanbrandenburg

ryanbrandenburg: When I asked what version we were branching to I was told 2.2.0-preview1 by @DamianEdwards.

natemcmaster: Ok.

natemcmaster:
:shipit:

PR #645, opened 2018-04-18
Merge dev into release

ryanbrandenburg:
Specifically we want to bring @pakrym's fix for SiteExtensions back into release/2.1.


pakrym:
> Specifically we want to bring @pakrym's fix for SiteExtensions back into release/2.1.

We don't know if it works on CI yet :)

ryanbrandenburg:
Well, we're pausing dev builds for now (our build queue was clogged), so the only way we'll find out any time soon is if we brought it into release/2.1.

pakrym:
Oh, fine, I was hoping to see if it passes in dev.

Issue #634, opened 2018-04-10
Microbenchmark verification only runs for the last project in benchmarks directory


pakrym:

Eilon:
@pakrym - who should look at this?

pakrym:
I will.

pakrym:
Not required for 2.1, it's just our internal validation mechanism to make sure microbenchmarks are not broken by product changes.

Eilon:
OK parking in 2.2.0-mq then.

PR #654, opened 2018-04-24
Run validate on all benchmarks

pakrym:
https://github.com/aspnet/BuildTools/issues/634

### DIFF
--- a/files/KoreBuild/modules/vstest/module.targets
+++ b/files/KoreBuild/modules/vstest/module.targets
@@ -115,10 +115,10 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
       Condition="'@(BenchmarksToValidate)' != ''" />
 
     <MSBuild Projects="%(BenchmarksToValidate.Identity)" Targets="GetTargetPath" Properties="$(SolutionProperties)">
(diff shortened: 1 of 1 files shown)


files/KoreBuild/modules/vstest/module.targets
```diff
@@ -115,10 +115,10 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
       Condition="'@(BenchmarksToValidate)' != ''" />
 
     <MSBuild Projects="%(BenchmarksToValidate.Identity)" Targets="GetTargetPath" Properties="$(SolutionProperties)">
-      <Output TaskParameter="TargetOutputs" PropertyName="ExecPath" />
+      <Output TaskParameter="TargetOutputs" ItemName="BenchmarkExecPath" />
     </MSBuild>
 
-    <RunDotNet Command="$(ExecPath) * --validate-fast" />
+    <RunDotNet Command="&quot;%(BenchmarkExecPath.Identity)&quot; * --validate-fast" />
```
natemcmaster: Consider using the `Arguments` parameter instead, which takes `string[]`, so you don't have to manually do the escaping.
```xml
<RunDotNet Arguments="%(BenchmarkExecPath.Identity);*;--validate-fast" />
```

PR #657, opened 2018-04-26
Update to SDK 2.1.300-rc1-008669

natemcmaster:


natemcmaster:
Yeah, that's a good idea. https://github.com/aspnet/BuildTools/issues/658

PR #660, opened 2018-04-30
Add SDK package for generating global CLI tool shims

natemcmaster:
Infrastructure work required to enable https://github.com/aspnet/Universe/issues/1126.

This is a generalization of the work added by @prafullbhosale and @seancpeters in https://github.com/aspnet/Scaffolding/pull/757. It pre-generates the apphost shim for global CLI tools for win-x86, win-x64, and osx-64.

It is eventually meant to move to https://github.com/dotnet/arcade so other teams can use it.

## Usage
Projects that need to bundle and sign the global CLI tool shim should add this to their .csproj file. This will include files in the .nupkg

```xml
  <Sdk Name="Microsoft.DotNet.GlobalTools.Sdk" />
 
  <PropertyGroup>
    <PackAsTool>true</PackAsTool>
    <GenerateToolShims>true</GenerateToolShims>
  </PropertyGroup>
```

cc @wli3


natemcmaster:
@pranavkm can you take a look at this too?

modules/BundledPackages/BundledPackages.proj
```diff
@@ -13,7 +14,30 @@
...
+  <ItemGroup>
+    <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(Version)" />
+    <PackageReference Include="Microsoft.DotNet.GlobalTools.Sdk" Version="$(Version)" />
+  </ItemGroup>
+</Project>
+]]>
+      </BundledPackageRestorerContent>
+    </PropertyGroup>
+
+    <WriteLinesToFile File="$(PublishDir)BundledPackageRestorer.csproj" Lines="$(BundledPackageRestorerContent)" Overwrite="true" />
```
pranavkm: sad 👏

modules/BundledPackages/BundledPackages.proj
```diff
@@ -13,7 +14,30 @@
...
+        This is a workaround for https://github.com/Microsoft/msbuild/issues/2914.
+        Currently, the only way to configure the NuGetSdkResolver is with NuGet.config, which is not generally used in aspnet org projects.
+        This project is restored so that it pre-populates the NuGet cache with SDK packages.
+      -->
+      <BundledPackageRestorerContent>
+<![CDATA[
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
pranavkm: Does this need `RestoreSources` or does it pick the one via Directory.Build.props?

src/Microsoft.DotNet.GlobalTools.Sdk/Microsoft.DotNet.GlobalTools.Sdk.nuspec
```diff
@@ -0,0 +1,19 @@
...
+  <metadata>
+    <id>$id$</id>
+    <version>$version$</version>
+    <authors>Microsoft</authors>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Provides additional support to .NET Core teams producing global CLI tools. This package is only intended for internal Microsoft use.</description>
+    <copyright>$copyright$</copyright>
+    <packageTypes>
+      <!-- This project needs to be MSBuildSdk because its targets need to influence the way /t:Restore works -->
+      <packageType name="MSBuildSdk" />
```
pranavkm: Does Razor Sdk need one of these?

modules/BundledPackages/BundledPackages.proj
natemcmaster: It's set in the PreinstallBundledPackages target (see below). We can't set it here since the path will be machine-specific.

src/Microsoft.DotNet.GlobalTools.Sdk/Microsoft.DotNet.GlobalTools.Sdk.nuspec
natemcmaster: Depends. Do you expect users to reference razor as a PackageReference or only as an Sdk?

If you add this, UI tooling in VS for PackageReference will break.

pranavkm: Preferably both. We use it as a PackageReference in other repos in aspnet

natemcmaster: If NuGet UI tooling wasn't a concern, I would say you should add this to your .nuspec

```xml
  <packageTypes>
      <packageType name="Dependency" /> <!-- PackageReference usage -->
      <packageType name="MSBuildSdk" /> <!-- Sdk usage -->
  </packageTypes>
```

But that might be dangerous because the NuGet team is thinking of breaking packages with multiple types.

See https://github.com/NuGet/Home/issues/6298 and https://github.com/NuGet/Home/issues/6484 for context.

PR #661, opened 2018-05-02
Use global.json and switch to netcoreapp2.2

ryanbrandenburg:


files/KoreBuild/scripts/KoreBuild.psm1
```diff
@@ -431,7 +423,9 @@ function __get_dotnet_sdk_version {
         Write-Warning "dotnet SDK version overridden by KOREBUILD_DOTNET_VERSION"
         return $env:KOREBUILD_DOTNET_VERSION
     }
-    return Get-Content (Join-Paths $PSScriptRoot ('..', 'config', 'sdk.version'))
```
natemcmaster: Since we don't generate global.json anymore, let's also remove usages of KOREBUILD_DOTNET_VERSION, both in .ps1 and .sh scripts.

files/KoreBuild/scripts/common.sh
```diff
@@ -61,7 +61,8 @@ __ensure_macos_version() {
 
 __get_dotnet_sdk_version() {
     local src="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-    version=$(< "$src/../config/sdk.version" head -1 | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+
+    version=$(< "$src/../../../global.json" head -1 | grep -Po '"version":.*?[^\\]",' global.json)
```
natemcmaster: This only works when building inside aspnet/BuildTools. You'll need to look for the global.json file in the repo root, not a relative path to this script.

Also, it would be better to use `jq` (and fallback to `python`) to read the json file. See https://github.com/aspnet/BuildTools/blob/d9eb8a53eca84f61a7b0845c99fc43a3ec51e353/scripts/bootstrapper/run.sh#L219-L234 for example.

modules/KoreBuild.Tasks/UpgradeDependencies.cs
```diff
@@ -210,6 +215,12 @@ private int UpdateDependencies(DependencyVersionsFile localVersionsFile, Depende
                     newValue = KoreBuildVersion.Current;
                     Log.LogMessage(MessageImportance.Low, "Setting InternalAspNetCoreSdkPackageVersion to the current version of KoreBuild");
                 }
+                else if (var.Key == "SdkPackageVersion")
```
natemcmaster: Naming what about: DotNetCliVersion?

We should also add update logic for the "msbuild-sdks" section of the global.json file. See https://github.com/aspnet/BuildTools/tree/dev/src/Microsoft.DotNet.GlobalTools.Sdk#usage for example. We will need this more as we migrated Internal.AspNetCore.Sdk and others away from being a PackageReference and towards being proper MSBuild SDKs.

Also, maybe we should have this information stored separately. Dependencies.props was really only mean for PackageReference versions. Maybe use a new file named Tooling.props or something? Just a thought...haven't committed to the idea.

modules/KoreBuild.Tasks/Utilities/GlobalJsonFile.cs
```diff
@@ -0,0 +1,30 @@
...
+        public GlobalJsonFile(string globalPath)
+        {
+            _path = globalPath;
+        }
+
+        public void SetSdkVersion(string version)
+        {
+            var globalJson = JsonConvert.DeserializeObject<Dictionary<string, object>>(File.ReadAllText(_path));
+
+            var sdkDict = (Dictionary<string, object>)globalJson["sdk"];
```
natemcmaster: Double check this section of the file exists first. It may not.

modules/KoreBuild.Tasks/Utilities/GlobalJsonFile.cs
```diff
@@ -0,0 +1,30 @@
...
+        private string _path;
+
+        public GlobalJsonFile(string globalPath)
+        {
+            _path = globalPath;
+        }
+
+        public void SetSdkVersion(string version)
+        {
+            var globalJson = JsonConvert.DeserializeObject<Dictionary<string, object>>(File.ReadAllText(_path));
```
natemcmaster: Handle the case were the file may not exist at all.

modules/KoreBuild.Tasks/UpgradeDependencies.cs
ryanbrandenburg: 2 should be its own item right? Once I touch 50 repos this is already going to be a messy change, I'd really rather not complicate it further.

natemcmaster: If by 2 you mean updating the "msbuild-sdks", then no, we need that before we can take this change. The msbuild-sdks section is required in at least 4 repos for the global tool shim stuff I just added yesterday.

natemcmaster:
By the way, it might be good to take the netcoreapp2.2 update and global.json changes separately. I don't believe these need to be done at the same time. The global.json work is less urgent and has deeper implications on how KoreBuild functions.

Some unanswered questions we'll have to address:
 - How can we check for a min sdk version? If KoreBuild requires a minimum version of the CLI, how can we check and enforce that.
 - How can we isolate task dependencies? Today, KoreBuild's tasks are tied to the same version of dependencies that ship in MSBuild itself, like NuGet and JSON.NET. We need to break that dependency and isolate tasks in order to make the decoupling work. I suspect we can do this incrementally, but it's something we should have a plan for first.
 - How can we ensure we install required runtimes for tools? The console tools bundled in KoreBuild depend on a .NET Core runtime. How can we ensure this is installed? (Btw, this is addressed today by only using netcoreapp2.0 and the files/KoreBuild/config/runtime.version file. Updating to netcoreapp2.2 breaks that assumption)

natemcmaster:
Sorry to pile it on, but one more question to consider:

 - Does aspnet/Universe need to ensure the same CLI version is used in all repos? If so, how will we override global.json? We do this today be re-generating global.json on each build, and overriding the korebuild-lock.txt file in aspnet/Universe.

ryanbrandenburg:
Results from meeting: We're going to punt 1 and 2, our solution for 3 is to only have netcoreapp2.0 projects. As for the other comment we all agreed that Universe should enforce its SDK on all modules via replacing the global.json with its own (the same way we do for korebuild-lock.json)

PR #670, opened 2018-05-08
Include symbols output even when custom nuspec file is used

natemcmaster:
I believe this is the reason symbols are missing for packages produced with a custom .nuspec


ajaybhargavb:
cc @pranavkm

PR #673, opened 2018-05-15
SiteExtensions use netcoreapp2.2

ryanbrandenburg:
We get errors like `error NU1202: Package is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Package Microsoft.AspNetCore.App 2.2.0-preview1-34201 supports: netcoreapp2.2 (.NETCoreApp,Version=v2.2)` on the CI because our extensions are still trying to build against netcoreapp2.2. I've created this PR to change that, but I'm not 100% if it's the correct fix. Thoughts?

### DIFF
--- a/src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
+++ b/src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
@@ -17,8 +17,8 @@
     </PropertyGroup>
 
     <ItemGroup Condition="'@(HostingStartupRuntimeStoreFrameworks->Count())' == '0'">
-      <HostingStartupRuntimeStoreTargets Include="netcoreapp2.1" Runtime="win7-x64" />
-      <HostingStartupRuntimeStoreTargets Include="netcoreapp2.1" Runtime="win7-x86" />
+      <HostingStartupRuntimeStoreTargets Include="$(TargetFramework)" Runtime="win7-x64" />
+      <HostingStartupRuntimeStoreTargets Include="$(TargetFramework)" Runtime="win7-x86" />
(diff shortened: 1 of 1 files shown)


src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
```diff
@@ -17,8 +17,8 @@
     </PropertyGroup>
 
     <ItemGroup Condition="'@(HostingStartupRuntimeStoreFrameworks->Count())' == '0'">
-      <HostingStartupRuntimeStoreTargets Include="netcoreapp2.1" Runtime="win7-x64" />
```
natemcmaster: Maybe we should make this `$(TargetFramework)` so it updates automatically with the project. Thoughts @pakrym ?

pakrym: SiteExtension itself targets net461 because it never mattered:
https://github.com/aspnet/AzureIntegration/blob/dev/extensions/Microsoft.AspNetCore.AzureAppServices.SiteExtension/Microsoft.AspNetCore.AzureAppServices.SiteExtension.csproj#L6

I'm fine with changing it to target framework here but we need to fix consumption side too.

ryanbrandenburg: By comsumption side you mean in the csproj in AzureIntegration? What would happen if we did the change here but not there?

pakrym: It would be failing the same way it does now.

PR #675, opened 2018-05-16
Don't police tfm for extensions

ryanbrandenburg:
Let's not run the TFM check for extensions, the same way we do for other projects, to avoid problems.


src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
```diff
@@ -38,7 +38,10 @@
         <ManifestLines Include="&lt;PackageReference Remove=&quot;Internal.AspNetCore.Sdk&quot; /&gt;" />
         <ManifestLines Include="&lt;PackageReference Include=&quot;Microsoft.AspNetCore.App&quot; Version=&quot;$(MicrosoftAspNetCoreAppPackageVersion)&quot; IsImplicitlyDefined=&quot;true&quot;/&gt;" />
         <ManifestLines Include="&lt;PackageReference Include=&quot;%(HostingStartupPackageReference.Identity)&quot; Version=&quot;%(HostingStartupPackageReference.Version)&quot; /&gt;" />
-        <ManifestLines Include="&lt;/ItemGroup&gt;&lt;/Project&gt;" />
+        <ManifestLines Include="&lt;/ItemGroup&gt;" />
```
natemcmaster: Totally not important, feel free to merge, but I recently learned how to use CDATA. I think it's more readable than a bunch of escape characters. Example: https://github.com/aspnet/BuildTools/blob/3d2175ced2de3110d649c72703373cd76f98db92/test/KoreBuild.FunctionalTests/GenerateTestProps.targets#L4-L17

ryanbrandenburg: This is awesome, 100% more readable. I created aspnet/Universe#1166 and will likely do it as a "nothing better to do on the bus" project.

PR #676, opened 2018-05-17
SiteExtensions use netcoreapp2.2

ryanbrandenburg:


natemcmaster:
why?

pakrym:
We can't make the site extension project target netcoreapp because of other reasons. And this should always be latest netcoreapp anyway.

PR #681, opened 2018-05-24
Add YAML templates

natemcmaster:
**Update**: fixed the config, no more VSTS errors :)

----------------

@chcosta can you take a look?

Not sure what I'm doing wrong, but I can't queue a build with this change:

image


.vsts-pipelines/templates/phases/default-build.yml
```diff
@@ -0,0 +1,43 @@
...
+#       Additional steps to run after build.sh/cmd
+
+parameters:
+  agentOs: ''
+  buildArgs: ''
+  beforeBuild: []
+  afterBuild: []
+
+phases:
+- phase: ${{ parameters.os }}
```
chcosta: Parameter appears to be named `agentOs`

.vsts-pipelines/templates/phases/default-build.yml
```diff
@@ -0,0 +1,43 @@
...
+  queue:
+    ${{ if eq(parameters.os, 'macOS') }}:
```
chcosta: Maybe just name this `Mac`?  `parameter.agentOs = macOs` seems a little redundant

.vsts-pipelines/templates/phases/default-build.yml
```diff
@@ -0,0 +1,43 @@
...
+  - ${{ parameters.beforeBuild }}
+  - ${{ if eq(parameters.os, 'Windows_NT') }}:
+    - script: .\build.cmd -ci ${{ parameters.buildArgs }}
+      displayName: Run build.cmd
+  - ${{ if ne(parameters.os, 'Windows_NT') }}:
+    - script: ./build.sh -ci ${{ parameters.buildArgs }}
+      displayName: Run build.sh
+  - task: PublishTestResults@2
+    displayName: Publish test results
+    condition: always()
```
chcosta: Consider `succeededOrFailed()` vs `always()`.  The only difference being that the former doesn't execute if the build was cancelled.

chcosta:
parameter substitution works most places, but not in the place you've tried it there.  The `-script` object just gets interpreted as a string that it runs on the agent.  You can work around this by mapping a variable and then using the variable in your script.

Similar to https://github.com/dotnet/arcade/blob/master/eng/build.yml#L35, but you can use a parameter substitution.

natemcmaster:
Awesome, you're a lifesaver. CI builds pass now.

.vsts-pipelines/templates/phases/default-build.yml
natemcmaster: In general, yes `succeededOrFailed` is probably good, but even if a build is cancelled, I want to see test results for test projects that finished before cancellation. For example, if i have to cancel a build that is gets hung while running tests...which happens a lot to Kestrel.

.vsts-pipelines/templates/phases/default-build.yml
```diff
@@ -0,0 +1,45 @@
+# default-build.yml
+# Description: Defines a build phase for invoking build.sh/cmd
+# Parameters:
+#   agentOs: [Windows (default), Linux, macOS]
```
chcosta: Add `buildArgs`

.vsts-pipelines/templates/phases/default-build.yml
natemcmaster: I figured I'd stick with `macOS` since that is the name of the operating system, after all.

.vsts-pipelines/builds/ci.yml
```diff
@@ -3,41 +3,14 @@ trigger:
...
-  - script: ./build.sh --ci
-  - task: PublishTestResults@2
-    condition: always()
-    inputs:
-      testRunner: vstest
-      testResultsFiles: 'artifacts/logs/**/*.trx'
-
+- template: ../templates/phases/default-build.yml
+  parameters:
+    agentOs: Windows
```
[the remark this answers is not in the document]
natemcmaster: Is there any particular reason for keeping the `_NT` part? Or is this just convention?

.vsts-pipelines/templates/phases/default-build.yml
natemcmaster: Thanks. Updated.

.vsts-pipelines/builds/ci.yml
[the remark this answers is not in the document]
chcosta: I tend to align with the values that our VSTS agent pools provide via demands, so it's consistent between the demands I specify for my queue, and my parameters.

Windows -> Windows_NT
Linux -> Linux
Mac -> Darwin
 
Currently, for the hosted pools, the machine types are separated by OS, so that doesn't specifically apply and I don't even think the Agent.Os is listed in the dotnet.visualstudio.com hosted machine pools.  I'm not certain where the demand values come from or if they were just arbitrarily defined.

Historically, the demands were used to determine what machine OS you want and all OS legs are pointing to the same agent pool.

https://github.com/dotnet/arcade/blob/master/.vsts-dotnet.yml#L8-L9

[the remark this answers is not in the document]
natemcmaster: Ok, that sounds like sound reasoning. So, should agentOs be `{ Windows_NT | Linux | Darwin }`?

by the way, you may be interested in this: https://developercommunity.visualstudio.com/content/problem/251502/the-agentos-capability-is-missing-on-hosted-or-hos.html

natemcmaster:
Closing/re-opening to test triggers on VSTS.

.vsts-pipelines/builds/ci.yml
```diff
@@ -3,41 +3,14 @@ trigger:
...
-    name: Hosted VS2017
-    demands: cmd
-  steps:
-  - script: build.cmd -ci
-  - task: PublishTestResults@2
-    condition: always()
-    inputs:
-      testRunner: vstest
-      testResultsFiles: 'artifacts/logs/**/*.trx'
+- template: ../templates/phases/default-build.yml
```
ryanbrandenburg: I like how short this is!

PR #683, opened 2018-06-04
Create a class library for consumers to implement manifest-driven code signing

natemcmaster:
Currently, code signing is done by folder location and convention. This is fragile and causes tight coupling
between the code building outputs and the signing tools. This change introduces a set of abstractions for working with the signrequest.xml file. Sign tool implementers can read this file and use it to drive code signing implementations.

First step towards https://github.com/aspnet/Coherence-Signed/issues/784


natemcmaster:
@ryanbrandenburg for more context on this, see https://github.com/aspnet/Coherence-Signed/issues/683 and https://github.com/aspnet/Coherence-Signed/issues/784. To help us move to VSTS, we need to decouple Coherence-Signed from aspnet/Universe

docs/Signing.md
```diff
@@ -6,20 +6,23 @@ and information about the strongname or certificate that should be used.
...
   <File Path="MyAssembly.dll" Certificate="MyCert" StrongName="MyStrongName" />
   <File Path="build/Another.dll" Certificate="MyCert" />
-  <Container Path="MyLib.1.0.0.nupkg" Type="nupkg" Certificate="NuGetCert">
+  <Nupkg Path="MyLib.1.0.0.nupkg" Certificate="NuGetCert">
     <File Path="lib/netstandard2.0/MyLib.dll" Certificate="MyCert" />
-  </Container>
-  <Container Path="MyVSTool.vsix" Type="vsix" Certificate="VsixCert">
+  </Nupkg>
+  <Vsix Path="MyVSTool.vsix" Certificate="VsixCert">
     <File Path="MyVSTool.dll" Certificate="MyCert" />
```
ryanbrandenburg: Just so I'm clear, this is the dll inside the vsix right?

docs/Signing.md
```diff
@@ -6,20 +6,23 @@ and information about the strongname or certificate that should be used.
 
 ## Format
 
-The signing request manifest supports three element types. A minimal example looks like this. See [Elements](#Elements) below for details
+The signing request manifest supports multiple element types. Each element represents a specific kind of action to take on a file. A minimal example looks like this. See [Elements](#Elements) below for details
 
 ```xml
 <SigningRequest>
   <File Path="MyAssembly.dll" Certificate="MyCert" StrongName="MyStrongName" />
```
ryanbrandenburg: Where is the path rooted? `$(artifactsDir)/build`?

docs/Signing.md
```diff
@@ -154,28 +157,33 @@ Root element. No options.
 
 A file to be signed.
 
-**Path** - file path, relative to the file path. If nested in a `<Container>`, is relative to the organization within the container
+**Path** - file path, relative to the file path. If nested in a parent element, is relative to the organization within the containing package
```
ryanbrandenburg: `file path, relative to the file path`. Have I just not had enough coffee or is this wrong?

docs/Signing.md
```diff
@@ -154,28 +157,33 @@ Root element. No options.
...
 
 **Path** - file path to the container
 
 **Certificate** - the name of the certificate to use
 
-**Type** - The type of the container. Instructs the consumer how to extract the container. Example values:
+#### `Vsix`
 
-  - zip
-  - tar.gz
```
ryanbrandenburg: I noticed that `tar.gz` and `msi` went way as types, do we just not need them anymore?

modules/KoreBuild.Tasks/GenerateSignRequests.cs
```diff
@@ -106,21 +113,25 @@ internal bool Execute(Func<TextWriter> writerFactory)
...
+                            $"Signing request item '{item.ItemSpec}' specifies an unknown container '{containerPath}'.");
                         continue;
                     }
+
                     var packagePath = item.GetMetadata("PackagePath");
-                    normalizedPath = string.IsNullOrEmpty(packagePath) ? normalizedPath : packagePath.Replace('\\', '/');
-                    var file = new SignRequestItem.File(normalizedPath,
+                    normalizedPath = string.IsNullOrEmpty(packagePath)
+                        ? normalizedPath
+                        : packagePath.Replace('\\', '/');
```
ryanbrandenburg: Should we use the NormalizePath function here instead?

modules/KoreBuild.Tasks/GenerateSignRequests.cs
```diff
@@ -136,18 +147,21 @@ internal bool Execute(Func<TextWriter> writerFactory)
...
                             continue;
                         }
 
                         var packagePath = item.GetMetadata("PackagePath");
-                        normalizedPath = string.IsNullOrEmpty(packagePath) ? normalizedPath : packagePath.Replace('\\', '/');
-                        var file = new SignRequestItem.Exclusion(normalizedPath);
-                        container.AddItem(file);
+                        normalizedPath = string.IsNullOrEmpty(packagePath)
+                            ? normalizedPath
+                            : packagePath.Replace('\\', '/');
```
ryanbrandenburg: NormalizePath again.

modules/KoreBuild.Tasks/GenerateSignRequests.cs
```diff
@@ -136,18 +147,21 @@ internal bool Execute(Func<TextWriter> writerFactory)
                     {
                         if (!containers.TryGetValue(containerPath, out var container))
```
ryanbrandenburg: What happens if there's an authoring error and something is listed as both an exclusion and an explicit signing item?

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestCollection.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.BuildTools.CodeSign
+{
+    /// <summary>
+    /// A collection of sign request items.
+    /// </summary>
+    public class SignRequestCollection : IEnumerable<SignRequestItem>
+    {
+        private readonly SortedDictionary<string, SignRequestItem> _items = new SortedDictionary<string, SignRequestItem>(StringComparer.Ordinal);
```
ryanbrandenburg: What makes the ordering important here?

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestItem.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// </summary>
+        public bool CanContainChildren
+            => ItemType == SignRequestItemType.Nupkg
+               || ItemType == SignRequestItemType.Zip
+               || ItemType == SignRequestItemType.Vsix;
+
+        /// <summary>
+        /// Children of the sign request item.
+        /// </summary>
+        public IEnumerable<SignRequestItem> Children => _children;
```
ryanbrandenburg: Should this be a `SignRequestCollection`? Why have `_children` be read-only but not `Children`?

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestItemType.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        Nupkg,
+
+        /// <summary>
+        /// A Visual Studio extension installer file
+        /// </summary>
+        Vsix,
+
+        /// <summary>
+        /// A signable binary or powershell script.
```
ryanbrandenburg: Do we know that it's a binary? Should this be `Executable` or something then? I would have assumed that `File` was much more general than binary or powershell.

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestManifestXmlReader.cs
```diff
@@ -0,0 +1,118 @@
...
+            switch (element.Name.ToString())
+            {
+                case "Nupkg":
+                    return ReadNupkg(element);
+                case "Vsix":
+                    return ReadVsix(element);
+                case "Zip":
+                    return ReadZip(element);
+                case "File":
+                    return SignRequestItem.CreateFile(GetPath(element), GetCertificate(element), GetStrongName(element));
```
ryanbrandenburg: Nitpick: personally I would create a `ReadFile` function just to follow the pattern we did with Nupkg Vsix etc. Total preference though, I see that it's less complicated than the others.

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestManifestXmlReader.cs
```diff
@@ -0,0 +1,118 @@
...
+                case "ExcludedFile":
+                    return SignRequestItem.CreateExclusion(GetPath(element));
+                default:
+                    IXmlLineInfo lineInfo = element;
+                    throw new InvalidDataException($"Unrecognized element type {element.Name} on line {lineInfo.LineNumber}");
```
ryanbrandenburg: Is there worth in writing out the whole line here since the signing request files are generated? Maybe not if they'll be artifacts.

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestManifestXmlReader.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        private void AddChildren(XElement element, SignRequestItem package)
+        {
+            foreach (var child in element.Elements())
+            {
+                package.AddChild(ReadElement(child));
+            }
+        }
+
+        private static string GetStrongName(XElement element) => element.Attribute("StrongName")?.Value;
```
ryanbrandenburg: Maybe this is too abstract-y, but should these constants and those in `SignRequestManifestXmlWriter` be drawing from a base class/joint Constants class?

docs/Signing.md
natemcmaster: Yes.

docs/Signing.md
natemcmaster: Rooted next to the signrequest.xml file.

docs/Signing.md
natemcmaster: there were never implemented in the first place. I don't believe we actually need them, at least not yet.

modules/KoreBuild.Tasks/GenerateSignRequests.cs
natemcmaster: Good point. Multiple items with the same file path is currently undefined behavior. We should add errors for that.

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestCollection.cs
natemcmaster: Deterministic build output.

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestItem.cs
natemcmaster: I wanted to hide the implementation. It shouldn't matter to the user how this collection is implemented.

src/Internal.AspNetCore.CodeSign.Abstractions/SignRequestItemType.cs
natemcmaster: It's not necessarily a binary. e.g. a powershell script is not a binary, but it's still a signable file type. Technically, so is a .js file.

PR #684, opened 2018-06-05
Add support for Internal.AspNetCore.Sdk to be used as an msbuild-sdk package

natemcmaster:
We can define "SDKs" as packages. This gives us more flexibility and power than PackageReference. It also has the advantage of not putting a dummy package in the restore graph.


.gitignore
```diff
@@ -19,6 +19,7 @@ node_modules
 .nuget/NuGet.exe
 project.lock.json
 .build
+.idea/
```
ryanbrandenburg: Just curious, noticed this popping up in your PR's, what are you using IntelliJ for?

test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
```diff
@@ -74,16 +74,20 @@ public void FullBuildCompletes()
                     Assert.Collection(pkg.Descendants("File"),
                         a =>
                         {
-                            Assert.Equal("tools/any/any/Newtonsoft.Json.dll", a.Attribute("Path")?.Value);
+                            Assert.Equal("tools/any/netcoreapp2.1/Newtonsoft.Json.dll", a.Attribute("Path")?.Value);
```
ryanbrandenburg: Interesting that this should be a result, do we know why?

.gitignore
natemcmaster: Rider :)

test/KoreBuild.FunctionalTests/SimpleRepoTests.cs
natemcmaster: I switched the tests from using a .nuspec to using `PackAsTool`. When the tests were written, the SDK did not yet support packing tool projects.

PR #687, opened 2018-06-06
Fix #685 - fix incorrect reference to dotnet in powershell

### LINKED ISSUE #685: VSTS upgrade deps failing
I have a vsts task that invokes
`run.ps1 -d C:\.dotnet -Path $pwd.Path upgrade deps`
It fails on this error
```
Invoke-KoreBuildCommand : You cannot call a method on a null-valued expression.
At E:\A\_work\13\s\run.ps1:204 char:5
+     Invoke-KoreBuildCommand $Command @Arguments
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Invoke-KoreBuildCommand], 
    RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull,Invoke-KoreBuildCommand
```

Both powershell 1.* and 2.* results in the same issue. However, I cannot repo the issue at all outside vsts. I RDP'd into the agent that reports the error and run the exact same command. It runs successfully.

P.S.
1. `Invoke-KoreBuildCommand` exists, `Command` is `upgrade`, `arguments` is `deps`
2. Prior to running `run.ps1 upgrade deps` I ran `run.ps1 install-tools`, that command ran fine.

natemcmaster:
cc @shhsu

### DIFF
--- a/files/KoreBuild/KoreBuild.sh
+++ b/files/KoreBuild/KoreBuild.sh
@@ -109,6 +109,9 @@ __install_tools() {
     "$__korebuild_dir/scripts/get-dotnet.sh" $verbose_flag "$install_dir" \
         || return 1
 
+    # Set environment variables
+    export PATH="$install_dir:$PATH"
+
     # This is a workaround for https://github.com/Microsoft/msbuild/issues/2914.
     # Currently, the only way to configure the NuGetSdkResolver is with NuGet.config, which is not generally used in aspnet org projects.
     # This project is restored so that it pre-populates the NuGet cache with SDK packages.
@@ -120,9 +123,6 @@ __install_tools() {
         __exec dotnet msbuild -restore -t:noop -v:m "$restorerfile"
     fi
     # end workaround
-
-    # Set environment variables
-    export PATH="$install_dir:$PATH"
 }
 
 __show_version_info() {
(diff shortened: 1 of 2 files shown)


shhsu:
This is a valid change. But FYI it would not fix https://github.com/aspnet/BuildTools/issues/685#issuecomment-395113591 Though.

Because if dotnet is not in the path. Even this line would not be correct

https://github.com/aspnet/BuildTools/blob/76985d5cc0bf3f707570bdc7714a16f5ead4d256/files/KoreBuild/scripts/KoreBuild.psm1#L8

shhsu:
VSTS is an edge case but to cater to that edge case, the bootstrap script `run.ps1` must be able to communicate to `KoreBuild.psm1` where dotnet home is supposed to be and ensure the path contains dotnet executable

natemcmaster:
I believe it should still fix #685, but if I'm wrong, we can re-open to investigate more. `$global:dotnet` will be updated to a new value by the `Install-Tools` function if dotnet is not installed, but we never got that far because of this defect in detecting if dotnet is installed.

I'm going to merge this. Can you update your version of KoreBuild and let me know if it fixes the issue? This fix will be in korebuild >= 2.2.0-preview1-17078

PR #688, opened 2018-06-06
Add a task for merging XML files into one document

natemcmaster:
Part of decoupling signing from TeamCity: https://github.com/aspnet/Coherence-Signed/issues/784

I write this task intending to use it to merge multiple signrequest.xml files in an aspnet/Universe build.

cc @ryanbrandenburg

### DIFF
--- a/modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
+++ b/modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
@@ -1,4 +1,4 @@
-<Project>
+﻿<Project>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), version.props))\build\sdk.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
```diff
@@ -10,6 +10,7 @@
     <Content Include="*.targets" CopyToPublishDirectory="PreserveNewest" />
     <Compile Include="..\..\shared\Microsoft.Extensions.CommandLineUtils.Sources\Utilities\*.cs" />
     <Compile Include="..\..\shared\Utilities\MSBuildListSplitter.cs" />
+    <Compile Include="..\BuildTools.Tasks\Utilities\FileHelpers.cs" Link="Utilities\FileHelpers.cs" />
```
ryanbrandenburg: Curious what that link property is and why this one has it while others don't.

modules/KoreBuild.Tasks/MergeXmlFiles.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (Files.Length == 1)
+            {
+                File.Copy(Files[0].ItemSpec, OutputPath);
+                return true;
+            }
+
+            return Execute(() => File.CreateText(OutputPath));
+        }
+
+        protected internal bool Execute(Func<TextWriter> writerFactory)
```
ryanbrandenburg: I assume this task has a "Be it on your own head if this merge results in duplicate results." warning on it since we're not doing and explicit deduping.

modules/KoreBuild.Tasks/KoreBuild.Tasks.csproj
natemcmaster: It's a UI thing. It tells VS where to display this in the Solution Explorer

modules/KoreBuild.Tasks/MergeXmlFiles.cs
natemcmaster: correct. In aspnet/Universe, if we had conflicting files, we would have other issues, and not just in the xml.

Issue #445, opened 2017-10-28
Add install-vs command


natemcmaster:
In https://github.com/aspnet/BuildTools/issues/428, I added support to _validate_ that VS is installed and has required workloads. To take this one step further, it would be nice if users or CI agents could also start a VS installation from command-line.

Example:
korebuild.json
```json
{
  "toolsets": {
     "visualstudio": {
       "required": [ "windows" ],
       "minVersion": "15.4",
       "requiredWorkloads": [ "Microsoft.VisualStudio.Workload.VCTools" ]
     }
   }
}
```
```
run.ps1 install vs
```

Ideas:
We could generate an installation manifest and pass it to the VS installer.

 - Visual Studio supports defining installation parameters as a file. https://docs.microsoft.com/en-us/visualstudio/install/automated-installation-with-response-file

Example:
```
vs_enterprise.exe --in visualstudio.json
```
```json
{
  "installChannelUri": ".\\ChannelManifest.json",
  "channelUri": "https://aka.ms/vs/15/release/channel",
  "installCatalogUri": ".\\Catalog.json",
  "channelId": "VisualStudio.15.Release",
  "productId": "Microsoft.VisualStudio.Product.Enterprise",

  "installPath": "C:\\VS2017",
  "quiet": false,
  "passive": false,
  "includeRecommended": false,
  "norestart": false,

  "addProductLang": [
    "en-US"
    ],

    "add": [
        "Microsoft.VisualStudio.Workload.NetWeb",
        "Microsoft.VisualStudio.Workload.VCTools"
    ]
}
```

cc @jkotalik

jkotalik:
@muratg can we budget some of my time to do this? Numerous people have pinged me trying to figure out what tools are required to build IISIntegration. Also, on VMs, I don't want to require people installing VS or struggling to figure out what dependencies are needed. 
cc/ @mikeharder

natemcmaster:
By the way, checkout https://github.com/aspnet/Infrastructure/blob/master/BuildAgent/SetupScripts/BuildAgent_VisualStudio.ps1 for automating running the VS installer.

jkotalik:
Yep! We should be able to translate those from our CI to local machines.

muratg:
What's the estimated cost for this?

natemcmaster:
small, 1-2 days

muratg:
:+1:  

@jkotalik let's talk about this once you have time from the current ANCM work.

PR #690, opened 2018-06-11
Adds command to install VS components

jkotalik:
For #445 

- Adds InstallToolsets task in korebuild that installs all toolsets specified in korebuild.json
   - Can invoke in a project by calling run.ps1 install vs
   - Currently only implements installing vs toolsets, but can extend to node toolsets
- Moves DownloadFile task into a helper, which is shared with InstallToolsets
- Adds VsInstallerHelpers which contains methods to download vs_enterprise.exe and creating the vs.json file for specifying requirements to vs

Unresolved issues:
- Choose vs_exe to download? (vs community or pro can't be used today)
- Testing is tricky because actually invoking this test will start vs_enterprise.exe. Is there a way to invoke a test run with vs.json options.
- Decent chunk of code is shared with GetToolsets, can potentially refactor into one msbuild task.

cc/ @mikeharder

### DIFF
--- a/modules/KoreBuild.Tasks/GetToolsets.cs
+++ b/modules/KoreBuild.Tasks/GetToolsets.cs
@@ -98,7 +98,7 @@ namespace KoreBuild.Tasks
                 if (vsToolset.Required != KoreBuildSettings.RequiredPlatforms.None)
                 {
                     Log.LogError($"Could not find an installation of Visual Studio that satisifies the specified requirements in {ConfigFile}" +
-                        "See https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community for more details on any missing components.");
+                        "Execute run.ps1 install vs to update or install the current vs installation.");
                 }
                 return;
             }
--- a/modules/KoreBuild.Tasks/GetToolsets.cs
+++ b/modules/KoreBuild.Tasks/GetToolsets.cs
@@ -97,7 +97,7 @@ namespace KoreBuild.Tasks
             {
                 if (vsToolset.Required != KoreBuildSettings.RequiredPlatforms.None)
                 {
-                    Log.LogError($"Could not find an installation of Visual Studio that satisifies the specified requirements in {ConfigFile}. " + 
+                    Log.LogError($"Could not find an installation of Visual Studio that satisifies the specified requirements in {ConfigFile}" +
                         "See https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community for more details on any missing components.");
                 }
                 return;
--- a/modules/KoreBuild.Tasks/InstallToolsets.cs
+++ b/modules/KoreBuild.Tasks/InstallToolsets.cs
@@ -27,7 +27,7 @@ namespace KoreBuild.Tasks
         /// Whether to install toolsets with or without user interation.
         /// It will default prompting users to confirm and see installation steps.
         /// </summary>
-        public bool QuietInstallationOfToolsets { get; set; }
+        public bool QuietVSInstallation { get; set; }
 
         /// <summary>
         /// Whether to upgrade existing toolsets.
@@ -95,7 +95,7 @@ namespace KoreBuild.Tasks
             }
 
             var vs = VsWhere.FindLatestCompatibleInstallation(vsToolset, Log);
-            var vsExePath = await VsInstallerHelper.DownloadVsExe(Log);
+            var vsExePath = await VsInstallerHelper.DownloadVsExe(Log, VSProductVersionType);
             var vsJsonFilePath = VsInstallerHelper.CreateVsFileFromRequiredToolset(vsToolset, Log, VSProductVersionType);
 
             var args = GetVisualStudioArgs(vs, vsJsonFilePath);
@@ -138,7 +138,7 @@ namespace KoreBuild.Tasks
             args.Add("--wait");
             args.Add("--norestart");
 
-            if (QuietInstallationOfToolsets)
+            if (QuietVSInstallation)
             {
                 args.Add("--quiet");
             }
--- a/modules/KoreBuild.Tasks/InstallToolsets.cs
+++ b/modules/KoreBuild.Tasks/InstallToolsets.cs
@@ -1,0 +1,142 @@
+// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using KoreBuild.Tasks.Utilities;
+using Microsoft.Build.Framework;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace KoreBuild.Tasks
+{
+    /// <summary>
+    /// Installs toolset information as listed in korebuild.json
+    /// </summary>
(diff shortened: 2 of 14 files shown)


modules/KoreBuild.Tasks/InstallToolsets.cs
```diff
@@ -0,0 +1,142 @@
...
+                        await InstallVsComponents(vs);
+                        break;
+                    // TODO support NodeJSToolset
+                    default:
+                        Log.LogWarning("Toolset checks not implemented for " + toolset.GetType().Name);
+                        break;
+                }
+            }
+
+            return true;
```
natemcmaster: Best practice: `return !Log.HasLoggedErrors;`

modules/KoreBuild.Tasks/Utilities/DownloadFileHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Build.Utilities;
+using Task = System.Threading.Tasks.Task;
+
+namespace KoreBuild.Tasks.Utilities
+{
+    public class DownloadFileHelper
+    {
+        public static async Task<bool> DownloadFileAsync(string uri, string destinationPath, bool overwrite, CancellationTokenSource cts, int timeoutSeconds, TaskLoggingHelper log)
```
natemcmaster: Nit change: `CancellationTokenSource` to `CancellationToken` and pass in `cts.Token`.

modules/KoreBuild.Tasks/Utilities/VsInstallerHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace KoreBuild.Tasks.Utilities
+{
+    internal class VsInstallerHelper
+    { 
+        public static async Task<string> DownloadVsExe(TaskLoggingHelper log)
+        {
+            // TODO put this in an obj folder instead of temp?
+            var tempPath = Path.Combine(Path.GetTempPath(), "vs_enterprise.exe");
+
+            await DownloadFileHelper.DownloadFileAsync(uri: "https://aka.ms/vs/15/pre/vs_enterprise.exe",
```
natemcmaster: This is only available on corpnet. Switch this to https://aka.ms/vs/15/release/vs_enterprise.exe instead.

modules/KoreBuild.Tasks/Utilities/VsInstallerHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            public List<string> Add { get; set; }
+        }
+
+        public static string CreateVsFileFromRequiredToolset(KoreBuildSettings.VisualStudioToolset vsToolset, TaskLoggingHelper log)
+        {
+            var vsFile = new VsJsonFile
+            {
+                ChannelUri = "https://aka.ms/vs/15/release/channel",
+                ChannelId = "VisualStudio.15.Release",
+                ProductId = "Microsoft.VisualStudio.Product.Enterprise",
```
natemcmaster: Community? Many of our contributors will not have the Enterprise license.

modules/KoreBuild.Tasks/module.targets
```diff
@@ -107,6 +106,19 @@ dependencies.props file in the current project to use latest versions.
...
+####################################################################################
+Target: InstallToolsets
+
+Installs all VS components specified in the korebuild.json required components.
+####################################################################################
+-->
+  <Target Name="InstallToolsets">
+    <InstallToolsets ConfigFile="$(KoreBuildConfigFile)"
+      Quiet="$(Quiet)"
+      Upgade="$(Upgrade)" />
```
natemcmaster: Typo: Upgrade

modules/KoreBuild.Tasks/module.targets
```diff
@@ -107,6 +106,19 @@ dependencies.props file in the current project to use latest versions.
...
+<!--
+####################################################################################
+Target: InstallToolsets
+
+Installs all VS components specified in the korebuild.json required components.
+####################################################################################
+-->
+  <Target Name="InstallToolsets">
+    <InstallToolsets ConfigFile="$(KoreBuildConfigFile)"
+      Quiet="$(Quiet)"
```
natemcmaster: Consider giving these properties to something more specific like "ToolsetInstallerLogLevel" and "UpgradeToolsets". Because MSBuild is like one big C program, it's good to give properties specific names to reduce the chance of conflict with other targets.

tools/KoreBuild.Console/Commands/InstallToolsetsCommand.cs
```diff
@@ -0,0 +1,63 @@
...
+                CommandOptionType.NoValue);
+            _upgradeOpt = application.Option("-u|--upgrade",
+                "Upgrade existing toolsets.",
+                CommandOptionType.NoValue);
+            base.Configure(application);
+        }
+
+        protected override int Execute()
+        {
+            var args = new List<string>
```
natemcmaster: Any particular reason you chose to implement this as a two-process hop into MSBuild? Or just because this is what the other tasks are doing?

It seems like we don't really need really need MSBuild to run this command, so IMO this would benefit from the simplicity of running the vs installer directly from KoreBuild.Console.

modules/KoreBuild.Tasks/Utilities/VsInstallerHelper.cs
jkotalik: Mentioned that above. I'll add a flag to chose which product to install for.

tools/KoreBuild.Console/Commands/InstallToolsetsCommand.cs
jkotalik: Few reasons:
- Code sharing. We need to parse the korebuild.json file, find the vs installation (vs_where), downloading the vs file, etc were all part of the KoreBuild.Tasks project already. I didn't see a great way to share code between the Korebuild.Console and Korebuild.Tasks projects, so I found implement it this way was easier than moving everything around.
- Extensibility. Implementing things in Korebuild.Console seems to become more difficult the more expansive it gets. For example, if we wanted to install toolsets for nodejs, we would need to move even more things to a shared project.

I originally started with having everything in Korebuild.Console, but I quickly realized that if I wanted to parse the korebuild.json file and find vs, it would be simpler to move everything into Korebuild.Tasks.

natemcmaster: Ok, that's good enough justification for me.

jkotalik:
Investigating test failures. Besides that, 🆙 📅

modules/KoreBuild.Tasks/GetToolsets.cs
```diff
@@ -97,7 +97,7 @@ private void GetVisualStudio(KoreBuildSettings.VisualStudioToolset vsToolset)
             {
                 if (vsToolset.Required != KoreBuildSettings.RequiredPlatforms.None)
                 {
-                    Log.LogError($"Could not find an installation of Visual Studio that satisifies the specified requirements in {ConfigFile}. " + 
+                    Log.LogError($"Could not find an installation of Visual Studio that satisifies the specified requirements in {ConfigFile}" +
```
natemcmaster: After thought: to make this new command more visible, you might consider updating the error message to include instructions such as "Execute 'run.ps1 install vs' to update or install VS"

ryanbrandenburg:
Oh, also update the docs [here](https://github.com/aspnet/BuildTools/blob/dev/docs/KoreBuild.md#korebuild-commands) to include this new command.

PR #696, opened 2018-06-18
Use LocalPath

ryanbrandenburg:
Lets set the DOTNET_HOME enLets set the DOTNET_HOME env to LocalPath so that things get cleaned out on checkout, and also skip first time experience.v to LocalPath so that things get cleaned out on checkout, and also skip first time experience.


ryanbrandenburg:
Let's not complicate things then.

PR #712, opened 2018-07-24
Reduce PR noise by skipping PRs which only merge aspnetci commits

natemcmaster:
Don't make PRs on the low traffic repos where the only changes are made by the weekly run of aspnetci to update dependencies.props.


natemcmaster:
FYI @weshaggard. We can add more conditions to skip PR generation if you are finding the PRs are too noisy.

PR #716, opened 2018-07-30
Publish artifacts from VSTS builds

natemcmaster:
Changes:

* Template: Publish artifacts placed in artifacts/build by default, but not for PRs from external forks
* KoreBuild: make upload.ps1 part of the artifacts output and modify the format so it can be consumed by VSTS deployment pipelines.

Part of https://github.com/aspnet/Coherence-Signed/issues/694

### DIFF
--- a/.vsts-pipelines/templates/phases/default-build.yml
+++ b/.vsts-pipelines/templates/phases/default-build.yml
@@ -10,12 +10,16 @@
 #       Additional steps to run before build.sh/cmd
 #   afterBuild: [steps]
 #       Additional steps to run after build.sh/cmd
+#   artifactsPath: string
+#       The path to the folder of build output to capture. Defaults to artifacts/build/.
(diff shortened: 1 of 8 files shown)


natemcmaster:
cc @HaoK @ryanbrandenburg - can you take a look? This will be the first config to start capturing and publishing files from a VSTS feed.

.vsts-pipelines/templates/phases/default-build.yml
```diff
@@ -10,12 +10,15 @@
 #       Additional steps to run before build.sh/cmd
 #   afterBuild: [steps]
 #       Additional steps to run after build.sh/cmd
+#   artifactsPath: string
+#       The path to the folder of build output to capture. Defaults to artifacts/build/
```
HaoK: Worth mentioning that setting to empty is how to disable publish (assuming I am reading this PR properly that is)?

PR #726, opened 2018-08-09
fix #725 by making internal sdk detection more resilient

### LINKED ISSUE #725: MSB4100 error when a project has no PackageReference elements
When one of the project KoreBuild is trying to build has no PackageReference elements, you end up with this error:

```
error MSB4100: Expected "$(_ReferencesInternalAspNetCoreSdk)" to evaluate to a boolean instead of "", in condition "'$(IncludeSymbols)' == '' AND $(_ReferencesInternalAspNetCoreSdk)".
```

This occurs because `_ReferencesInternalAspNetCoreSdk` is defined by searching for a PackageReference with the Identity value `Internal.AspNetCore.Sdk` but when there are no PackageReference elements the MSBuild expression evaluates to `''` instead of `'false'`

The problematic expression in question:

https://github.com/aspnet/BuildTools/blob/ad09c8008e2f4ad63757131dd8c5a8e8a7653d60/files/KoreBuild/modules/solutionbuild/Project.Inspection.targets#L18-L19

Because `_ReferencesInternalAspNetCoreSdk` is `''`, it fails to function as a boolean.

PR to fix coming soon.

anurse:
Fixes #725


natemcmaster:
It appears to be flakiness in the network.

anurse:
Ok, just wondering if it was something you've seen before. I'll just respin it.

anurse:
Yeah, VSTS wasn't letting me re-queue it anyway.

PR #746, opened 2018-08-27
[WIP] Add support for setting env-vars in VSTS template

anurse:
Adds a new parameter 'env' to the default build template that sets environment variables in the `script` task that runs the build script. Useful for passing secret variables in as environment variables, because they don't flow unless explicitly specified.

We need this in SignalR to flow credentials for Sauce Labs down into the build script. Just putting the macros in the `buildArgs` parameter doesn't seem to work.

WIP because I'm still testing this. I'll come back and confirm when it's working, but I wanted to throw this out for review because I don't expect it to change if it works and it's pretty simple :).

### DIFF
--- a/.vsts-pipelines/templates/phases/default-build.yml
+++ b/.vsts-pipelines/templates/phases/default-build.yml
@@ -23,6 +23,8 @@
 #           The name of the artifact container
 #   variables: { string: string }
 #     A map of custom variables
+#   env: { string: string }
+#     A map of environment variables to set during the build script.
 #   matrix: { string: { string: string } }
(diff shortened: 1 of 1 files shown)


natemcmaster:
> Just putting the macros in the buildArgs parameter doesn't seem to work.

I'd like to know why this doesn't work. In theory, the `variables: {}` map should have the same effect, and has the added benefit of applying to all steps of the build.

anurse:
Guh, well now it's working perfectly fine. I don't know what I did differently, maybe some YAML goop somewhere.

PR #748, opened 2018-09-01
Call the right function in the bash script

davidfowl:
Found this while trying to make a minimal docker dev env for ASP.NET Core


natemcmaster:
Normally when we update this script we copy it to all repos except aspnet/Universe (it has customization for ProdCon).

davidfowl:
> Normally when we update this script we copy it to all repos except aspnet/Universe (it has customization for ProdCon).

ZOMG, I'll just merge this 😄

davidfowl:
@ryanbrandenburg I hope we have a bot for that.

natemcmaster:
We don’t have a bot for it because it happens rarely enough that automation isn’t worth the effort.

PR #750, opened 2018-09-05
dotnet-install use a lockfile

### LINKED ISSUE #730: Race condition when calling dotnet-install.ps1/sh twice in parallel threads
If multiple processes call dotnet-install.ps1/sh twice, the script may fail while attempting to write files into DOTNET_HOME. We should add a mechanism to prevent this race.

ryanbrandenburg:
Fixes #730.

There's definitely still potential for racing here, but 1) It's reduced significantly and 2) I'm not sure that we can do much better between independent powershell executions.

PR most easily viewed with `?w=1` to ignore whitespace only changes.

### DIFF
--- a/files/KoreBuild/scripts/dotnet-install.ps1
+++ b/files/KoreBuild/scripts/dotnet-install.ps1
@@ -192,7 +192,6 @@ function GetHTTPResponse([Uri] $Uri)
 {
     Invoke-With-Retry(
     {
-
         $HttpClient = $null
 
         try {
@@ -248,7 +247,6 @@ function GetHTTPResponse([Uri] $Uri)
     })
 }
 
-
 function Get-Latest-Version-Info([string]$AzureFeed, [string]$Channel, [bool]$Coherent) {
     Say-Invocation $MyInvocation
 
@@ -285,7 +283,6 @@ function Get-Latest-Version-Info([string]$AzureFeed, [string]$Channel, [bool]$Co
 
     return $VersionInfo
 }
-
 
 function Get-Specific-Version-From-Version([string]$AzureFeed, [string]$Channel, [string]$Version) {
     Say-Invocation $MyInvocation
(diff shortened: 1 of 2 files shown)


ryanbrandenburg:
Nevermind, the hook was just broken, I fixed it.

files/KoreBuild/scripts/dotnet-install.sh
```diff
@@ -959,31 +959,49 @@ do
...
+local lockFile=""
+local waitTime=0
+
+while [ ! -f $lockFile ] && [ $waitTime < 120 ]
+do
+    sleep 10
+    waitTime += 10
+done
+
+if [ $waitTime > 120]; then
```
natemcmaster: Also, this is a common pitfall of bash. Use `$waitTime -gt 120`. Likewise above, use `-lt`. See  http://tldp.org/LDP/abs/html/comparison-ops.html

I think this is a syntax error. You need a space between 120 and ].

natemcmaster:
We normally try to keep dotnet-install.sh/ps1 scripts inline with https://github.com/dotnet/cli/tree/master/scripts/obtain. They've been pretty receptive to us opening PRs to update their copy with our improvements.

@johnbeisner @livarcocc - would you like to review this? After this goes in, we'd probably open a similar PR on dotnet/cli to update that copy of the script.

files/KoreBuild/scripts/dotnet-install.ps1
```diff
@@ -501,89 +498,109 @@ function Prepend-Sdk-InstallRoot-To-Path([string]$InstallRoot, [string]$BinFolde
...
-    Say "Primary - $DownloadLink"
-    if ($LegacyDownloadLink) {
-        Say "Legacy - $LegacyDownloadLink"
-    }
-    Say "Repeatable invocation: .\$($MyInvocation.Line)"
-    exit 0
+$waitTime = 0
+while((Test-Path $lockFile) -and ($waitTime -lt 120))
+{
+    Start-Sleep -Seconds 10
```
natemcmaster: Before sleeping, it might be useful to log a message such as "Another installation of .NET Core is in process. Waiting for that installation to complete..."

files/KoreBuild/scripts/dotnet-install.ps1
```diff
@@ -501,89 +498,109 @@ function Prepend-Sdk-InstallRoot-To-Path([string]$InstallRoot, [string]$BinFolde
     }
 }
 
-$CLIArchitecture = Get-CLIArchitecture-From-Architecture $Architecture
-$SpecificVersion = Get-Specific-Version-From-Version -AzureFeed $AzureFeed -Channel $Channel -Version $Version
-$DownloadLink = Get-Download-Link -AzureFeed $AzureFeed -SpecificVersion $SpecificVersion -CLIArchitecture $CLIArchitecture
-$LegacyDownloadLink = Get-LegacyDownload-Link -AzureFeed $AzureFeed -SpecificVersion $SpecificVersion -CLIArchitecture $CLIArchitecture
+$lockFile = Join-Path $PSScriptRoot "dotnetinstall.lock"
```
natemcmaster: Putting this lockfile next to the dotnet-install.sh script doesn't solve the problem if two different copies of this script are invoked in parallel. What if the script placed the lockfile in the target install directory?

ryanbrandenburg:
:up::date:

PR #766, opened 2018-09-26
Implement MicroBuild code signing

natemcmaster:
This transforms MSBuild items (see docs/Signing.md) into a manifest which is then invoked by the Roslyn sign tool. In environments with MicroBuild enabled, this will run code signing. In local developer builds, this tool will do a dry run that validates the assemblies in .nupkgs are account file in the sign data.

### DIFF
--- a/modules/KoreBuild.Tasks/CodeSign.targets
+++ b/modules/KoreBuild.Tasks/CodeSign.targets
@@ -7,7 +7,12 @@
 
     <!-- Repos can explicitly define the signdata.json file if they want do. They must set SignToolDataFile to the path of the file. When set, file generation is skipped. -->
     <GenerateSignToolDataFile  Condition="'$(SignToolDataFile)' == '' ">true</GenerateSignToolDataFile>
+
+    <!-- The location of SignToolData.json. This can be handwritten if you don't want to use the generated version. -->
     <SignToolDataFile Condition="'$(GenerateSignToolDataFile)' == 'true'">$(_IntermediateSignToolDataFile)</SignToolDataFile>
+
+    <!-- Relative paths in SignToolData.json are relative to this path -->
+    <SignToolDataWorkingDir Condition=" '$(SignToolDataWorkingDir)' == '' ">$(RepositoryRoot)</SignToolDataWorkingDir>
   </PropertyGroup>
 
   <Target Name="CodeSign" Condition=" '$(SkipCodeSign)' != 'true' "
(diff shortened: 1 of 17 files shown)


modules/KoreBuild.Tasks/CodeSign.targets
```diff
@@ -0,0 +1,43 @@
...
+    </GetPathToFullMSBuild>
+
+    <GenerateSignToolDataFile Condition=" '$(GenerateSignToolDataFile)' != 'false' "
+      Files="@(FilesToSign)"
+      Exclusions="@(FilesToExcludeFromSigning)"
+      OutputPath="$(_IntermediateSignToolDataFile)" />
+
+    <ItemGroup>
+      <SignToolOptions Condition=" '$(MSBuildx86Path)' != '' " Include="-msbuildPath;$(MSBuildx86Path)" />
+      <SignToolOptions Condition=" '$(SignType)' != 'real' AND '$(SignType)' != 'test' "  Include="-test" />
```
HaoK: Just curious about when we use SignType != test or real, some kind of other special test case?

HaoK:
CI is failing tho

Eilon:
Does this essentially do signing on a per-repo basis, that is, each repo will get signed as its own build runs? (No problem with that, just trying to understand how this fits in.)

natemcmaster:
> Does this essentially do signing on a per-repo basis, that is, each repo will get signed as its own build runs?

Yes, that's the idea. It code-signs each repo in-place. 

For Universe builds, this is a big shift in the way we do things. Right now, we build most of the product unsigned, move it to a different machine for signing, and then move it back to other machines to finish up the rest of the build for things like installers. This should save our build the time it takes to copy between the machines.

There is another benefit here, too. Repos to aspnet/Blazor need code-signing too but don't participate in Universe builds. This makes code-signing more accessible for these repos. In fact, aspnet/Blazor is already using the Roslyn signtool. This PR essentially commonizes the work we did to get Blazor code signed.

natemcmaster:
I'm verifying all of our repos still build after this change to build tools. I found a few places where the signing config was incorrect. I'm submitting fixes, such as https://github.com/aspnet/EntityFrameworkCore/pull/13426 before this merges.

natemcmaster:
Alright, I've done a dry run on Universe and I believe all repos are updated to handle this change. Waiting on this PR https://github.com/aspnet/Scaffolding/pull/899. Once it's in, I'll merge this update.

natemcmaster:
Ok, here we go.

modules/KoreBuild.Tasks/CodeSign.targets
natemcmaster: SignType!= test or real for local dev builds. This instructs signtool to do a dry-run without actual codesigning. The tool does some validation of the signing configuration and should catch the most common errors.

PR #771, opened 2018-10-01
Remove MicroBuild.Core as a dependency of Internal.AspNetCore.Sdk

natemcmaster:
Most microbuild tasks do not support .NET Core, so this causes task assembly load issues. We don't really need MicroBuild.Core in the project context. We just need it restored to disk for CI builds. This should still put the package in the cache without causing task assembly load errors.

### DIFF
--- a/modules/BundledPackages/BundledPackages.proj
+++ b/modules/BundledPackages/BundledPackages.proj
@@ -27,7 +27,6 @@
   <Sdk Name="Internal.AspNetCore.Sdk" Version="$(Version)" />
 
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


Eilon:
Do we need to clean up these uses of `DisableMicroBuild`? https://github.com/search?q=org%3Aaspnet+DisableMicroBuild&type=Code

natemcmaster:
Yes. 

:up: :date:

natemcmaster:
We can keep the DisableMicroBuild variable in Universe. Same variable name, different context.

PR #790, opened 2018-10-10
Collect test logs in VSTS

BrennanConroy:
@natemcmaster ~~I don't remember how to access the build settings to put this somewhere useful in our builds~~ Nevermind, it's handled via the tasks in this file

### DIFF
--- a/.vsts-pipelines/templates/phases/default-build.yml
+++ b/.vsts-pipelines/templates/phases/default-build.yml
@@ -70,6 +70,7 @@ phases:
     demands: ${{ parameters.demands }}
   variables:
     AgentOsName: ${{ parameters.agentOs }}
+    ASPNETCORE_TEST_LOG_DIR: 'artifacts/testlogs'
     DOTNET_HOME: $(Agent.WorkFolder)/.dotnet
     BuildScriptArgs: ${{ parameters.buildArgs }}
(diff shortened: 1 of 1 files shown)


.vsts-pipelines/templates/phases/default-build.yml
```diff
@@ -70,6 +70,7 @@ phases:
     demands: ${{ parameters.demands }}
   variables:
     AgentOsName: ${{ parameters.agentOs }}
+    ASPNETCORE_TEST_LOG_DIR: ${{ parameters.artifacts.path }}/testlogs
```
natemcmaster: Can we put this in artifacts/logs/ instead? We already put other logging files into this folder.

BrennanConroy: If I understand correctly the publish task uses `pathtoPublish: ${{ parameters.artifacts.path }}` to determine what to publish as an artifact. So artifacts/logs wouldn't be picked up

natemcmaster: We can change artifact publishing to include other folders. I think using artifacts/build/testlogs/ may impact the way the "known artifacts" check works. KoreBuild scans artifacts/build/testlogs/ to make sure everything in it is accounted for.

PR #806, opened 2018-10-17
Update test log dir

JunTaoLuo:
I haven't seen any uses of `Build.*` variables so I'm not sure if it will be resolved correctly. How would I check anyway?

PR #810, opened 2018-10-18
Revert #806

BrennanConroy:
This is causing build failures because VSTS has bad (tm) logic for zipping artifacts. https://dnceng.visualstudio.com/public/_build/results?buildId=32061&view=logs

Who can we contact on VSTS to see if they can resolve it?


JunTaoLuo:
:shipit: to get the tests running, but we should figure out why vsts is running into long path issues

PR #812, opened 2018-10-18
Update YAML templates to use new DevOps features

natemcmaster:
Azure DevOps has made some syntax changes to YAML. This updates .azure/templates/project-ci.yml to the latest syntax and adds some features.

For compatibility, the old templates in .vsts-pipelines/templates/project-ci.yml remain untouched. Those will eventually be removed.

Changes:
* Add support for code signing tasks
* Update YAML syntax to 'jobs' instead of phases
* Update pool specification
* Add obsolete warning to old templates

### DIFF
--- a/.azure/templates/jobs/default-build.yml
+++ b/.azure/templates/jobs/default-build.yml
@@ -96,10 +96,6 @@ jobs:
   steps:
   - checkout: self
     clean: true
-  - ${{ if eq(parameters.artifacts.publish, 'true') }}:
-    # Workaround for PublishBuildArtifacts failing when this folder does not exist
-    - script: mkdir $(ASPNETCORE_TEST_LOG_DIR)
-      displayName: Make test log directory
   - ${{ if and(eq(parameters.agentOs, 'Windows'), eq(parameters.codeSign, 'true')) }}:
     - task: MicroBuildSigningPlugin@1
       displayName: Install MicroBuild Signing plugin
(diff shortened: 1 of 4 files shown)


.azure/templates/jobs/default-build.yml
```diff
@@ -29,55 +31,82 @@
...
+    ${{ if and(eq(parameters.poolName, ''), eq(parameters.agentOs, 'Windows')) }}:
+      ${{ if ne(parameters.codeSign, 'true') }}:
+        name: Hosted VS2017
+        vmImage: vs2017-win2016
+      ${{ if eq(parameters.codeSign, 'true') }}:
+        name: DotNetCore-Windows
   variables:
     AgentOsName: ${{ parameters.agentOs }}
-    ASPNETCORE_TEST_LOG_DIR: ${{ format('{0}/logs/testlogs', parameters.artifacts.path) }}
+    ASPNETCORE_TEST_LOG_DIR: $(Build.ArtifactStagingDirectory)/testlogs
```
BrennanConroy: Isn't this going to cause the long path issues again?

natemcmaster: I'm not sure I understood why it failed before, but thought we'd give this a try. The file path is at least 16 characters shorter (removes "artifacts/logs/" from the path)

BrennanConroy: It's not the location we put the logs that was the problem, it's the way VSTS zips them up that caused issues. They create a new subfolder and try to copy the logs into it which can increase the path past the limit.

natemcmaster: Thanks for clarifying. I'm going to remove ASPNETCORE_TEST_LOG_DIR while we investigate workarounds.

Btw, for faster iteration on this, you can set ASPNETCORE_TEST_LOG_DIR and add the upload tasks in your repo's .yml files instead of needing to update the template. When we find a solution that works, we can add that to the template.

BrennanConroy: 👍 I'll keep that in mind while we discuss with the VSTS folks

BrennanConroy:
Is the dotnet bot broken?

natemcmaster:
The PR generator one?

BrennanConroy:
Yeah!

natemcmaster:
It might be. Look at https://devdiv.visualstudio.com/DevDiv/_build?definitionId=9524 for failed builds.

I manually triggered the buildtools bot.

PR #814, opened 2018-10-18
Publish artifacts from forks

sebastienros:


sebastienros:
The previous PR was wrongly targeting master

natemcmaster:
Requesting @Eilon to take a look. This means artifacts are uploaded for all public PRs.

Eilon:
Hmm yes that is a no-go. We cannot publish artifacts for arbitrary PRs.

.azure/templates/phases/default-build.yml
```diff
@@ -94,7 +94,6 @@ phases:
   - ${{ if eq(parameters.artifacts.publish, 'true') }}:
     - task: PublishBuildArtifacts@1
       displayName: Upload artifacts
-      condition: eq(variables['system.pullrequest.isfork'], false)
```
natemcmaster: Suggest keeping this as:
```
condition: always()
```
This ensures artifacts are uploaded even if the tests failed. This is useful for debugging test failures.

sebastienros:
> Hmm yes that is a no-go. We cannot publish artifacts for arbitrary PRs.

@davidfowl so benchmark bot will be just for PR from local branches

sebastienros:
@Eilon would it be ok, if it's possible and if this is the concern, to publish them only in the internal CI?

Eilon:
Why would we want artifacts from random forks?

PR #819, opened 2018-10-19
Fix max path issue with testlog artifacts

BrennanConroy:
Should I update the deprecated template too?

This can go in before https://github.com/aspnet/Logging/pull/904 assuming we're happy with the variable


BrennanConroy:
:up: 📅

PR #825, opened 2018-10-24
Update SDK to 2.1.500-preview-009398

natemcmaster:


natemcmaster:
FYI @muratg @Eilon for 2.1.6

PR #837, opened 2018-10-29
Add KoreBuild targets to build a list of projects instead of solutions

natemcmaster:
Part of unifying to fewer repos means that we don't have one .sln file which covers everything which should be built. This introduces new targets which allow support for building a list of projects instead of .sln files.

The default behavior remains the same. To opt-in to this new behavior, repos can set `<BuildSolutions>false</BuildSolutions>`


natemcmaster:
This paves the way for https://github.com/aspnet/Extensions/pull/436

PR #849, opened 2018-11-02
Allow yml files to specify build directory

jkotalik:
When running CI builds in aspnetcore, we need a way to run build.cmd inside of a subfolder.

I am open for suggestions to make this more robust.

### DIFF
--- a/.azure/templates/jobs/default-build.yml
+++ b/.azure/templates/jobs/default-build.yml
@@ -129,7 +129,10 @@ jobs:
       displayName: Upload artifacts
       condition: eq(variables['system.pullrequest.isfork'], false)
       inputs:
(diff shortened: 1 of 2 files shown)


natemcmaster:
This solution is fine, but I'm wondering if we should take it a step further and allow callers to replace the build.sh/cmd step altogether. Thoughts?

jkotalik:
I'd be okay with that. However, I'm noticing we probably need to use the buildDirectory for uploading artifacts: https://dnceng.visualstudio.com/public/_build/results?buildId=38899&view=logs (or we can disable upload artifact for CI builds).

jkotalik:
Chatted offline, closing in favor of copying the yml files to mondo repo.

jkotalik:
Chatted offline, we are moving these files to aspnetcore.

PR #851, opened 2018-11-02
Add property to allow repos to control the NuGet package output path

natemcmaster:
KoreBuild currently enforces PackageOutputPath on all repos it packs. This allows repos to control the location of NuGet package output.

For example, this will be useful in 2.2 to allow us to organize packages into different folders based on package category.


jkotalik:
Seems to be some signing issues

PR #863, opened 2018-11-09
Exclude corehost from runtime store

pakrym:
Add a workaround to exclude hostpolicy.dll, apphost.exe, etc. from runtime store. 

Also, override global NoBuild when generating additionalDeps during the repositiory Package phase.


src/Internal.AspNetCore.SiteExtension.Sdk/build/Internal.AspNetCore.SiteExtension.Sdk.targets
```diff
@@ -36,6 +36,7 @@
 
         <ManifestLines Include="&lt;Project Sdk=&quot;Microsoft.NET.Sdk&quot;&gt;&lt;ItemGroup&gt;" />
         <ManifestLines Include="&lt;PackageReference Remove=&quot;Internal.AspNetCore.Sdk&quot; /&gt;" />
+        <ManifestLines Include="&lt;PackageReference Remove=&quot;Microsoft.NETCore.App&quot; /&gt;" />
```
natemcmaster: This looks like a good candidate for a CDATA block.

Example: https://github.com/aspnet/BuildTools/blob/1f1bbd1a38ec97acd322933661cf8ef257b4a496/modules/BundledPackages/BundledPackages.proj#L24-L48

PR #869, opened 2018-11-12
Support running KoreBuild on full msbuild

natemcmaster:
These changes support using "MSBuild.exe" instead of "dotnet msbuild" when building a repo with KoreBuild.  The default still remains `dotnet msbuild`, so this is an opt-in change for repos.

### Usage:
In korebuild.json, add the following:
```json
{
  "msbuildType": "full"
}
```
_This only applies to Windows builds._

### Changes in this PR:
* Remove unused tasks and targets (reduces what we had to port to .NET Framework)
* Invoke VSWhere to find MSBuild.exe
* Backport improvements to SourceLink and VSWhere tasks to reduce dependency on Json.Net

### Motivations
* One more step towards Arcade convergence (Arcade uses MSBuild.exe on Windows)
* Mondo repo - ASP.NET Core has many projects which required full MSBuild.exe. This simplifies the build system so that we don't have to use `dotnet msbuild` to find and invoke a second MSBuild process. It also allows the build to use ProjectReference correctly.


files/KoreBuild/msbuild/KoreBuild.RepoTasks.Sdk/Sdk/Sdk.targets
```diff
@@ -1,25 +1,40 @@
...
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="$(MicrosoftBuildVersion)" IsImplicitlyDefined="true" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildVersion)" IsImplicitlyDefined="true" PrivateAssets="All" />
-    <!-- because almost everyone ends up using JSON in repo tasks -->
-    <PackageReference Include="Newtonsoft.Json" Version="$(JsonInMSBuildVersion)" IsImplicitlyDefined="true" PrivateAssets="All" />
-  </ItemGroup>
+          Between MSBuild 15.6 and 15.8, the ref packages dropped .NET Standard 1.x suppport.
+          Use the netstandard2.0 packages if possible, otherwise fallback to 15.6.
+        -->
+        <MicrosoftBuildVersion Condition="$(TargetFramework.StartsWith('netstandard1'))">15.6.82</MicrosoftBuildVersion>
+        <MicrosoftBuildVersion Condition="'$(MicrosoftBuildVersion)' == ''">15.8.166</MicrosoftBuildVersion>
```
Eilon: These two versions are, like, super specific. Is this fragile?

modules/BuildTools.Tasks/CreateSourceLink.cs
```diff
@@ -51,20 +51,52 @@ public override bool Execute()
             SourceLinkRoot += '*';
 
             var codeSource = ConvertUrl();
-            var data = new JObject
+
+            // TODO allow multiple. See https://github.com/aspnet/Universe/issues/858
```
Eilon: Stale comment?

modules/BuildTools.Tasks/CreateSourceLink.cs
```diff
@@ -51,20 +51,52 @@ public override bool Execute()
...
             };
-            File.WriteAllText(DestinationFile, data.ToString(Formatting.None));
 
+            GenerateSourceLink(DestinationFile, mappings);
             SourceLinkFile = DestinationFile;
-
             return true;
         }
 
+        private static void GenerateSourceLink(string filePath, List<Mapping> mappings)
```
Eilon: Any reason not to just use Json.NET for this? Looks like it's available anyway, no?

natemcmaster: This dependency on Json.Net was causes build failures when opening solutions in VS 2019. We already fixed this in 2.2, and cherry-picked that change back to 2.1.

modules/BuildTools.Tasks/CreateSourceLink.cs
natemcmaster: I cherry picked this file completely without changes from release/2.2. But yes, this comment appears to be stale.

files/KoreBuild/msbuild/KoreBuild.RepoTasks.Sdk/Sdk/Sdk.targets
natemcmaster: There are two versions here because the MSBuild made breaking changes between 15.6 and 15.8. 15.6 only supports netstandard1.x, 15.8 only supports netstandard2.0. We prefer to use netstandard2.0 where possible (faster restores, fewer dependencies), but I kept the fallback to 15.6 for compatibility. In the master branch, I plan to remove netstandard1.x support.

Eilon: OK, but it's no problem that these are super exact versions? Like what if instead of 15.8.166 it's really 15.8.167? Is that even a thing?

natemcmaster: Right, not a problem. The very-specific versions are due to the way MSBuild versions the packages they release to nuget.org.

image

kp2017:
@natemcmaster Following the [directions](https://github.com/aspnet/AspNetCore/blob/master/docs/BuildFromSource.md) now produces an error:

'Could not find a version of MSBuild which satisfies the requirements specified in korebuild.json.
At C:\Users\username\\.dotnet\buildtools\korebuild\3.0.0-build-20181206.3\scripts\KoreBuild.psm1:511 char:5
+     throw 'Could not find a version of MSBuild which satisfies the re ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Could not find ...korebuild.json.:String) [], RuntimeException
    + FullyQualifiedErrorId : Could not find a version of MSBuild which satisfies the requirements specified in korebuild.json.`

muratg:
@kp2017 Our repos and code is currently being restructured, so anything can fail. If you insist on trying to get it done though, I recommend trying different builds of the SDK including the latest one.

PR #878, opened 2018-11-16
Updating SDK to 3.0.100-preview-009752

ryanbrandenburg:
I'm told this will fix https://github.com/dotnet/sdk/issues/2672 which should unblock aspnet/Templating.


natemcmaster:
You'll need to update the MSBuild version to consume this update. They're important.

ryanbrandenburg:
:up::date: looks like we cleaned up the worst of it, should we merge and see what shakes out or wait for mondo stuff to settle some?

PR #879, opened 2018-11-20
Move to latest reference assembly archive

dougbu:
- should get Razor building on Linux and MacOS again


dougbu:
I've created the requisite archive. That needs to be in blob storage before I merge this.

PR #893, opened 2018-12-07
Fix NU5123 when the package id for shared sources is long

natemcmaster:
Cherry-pick's https://github.com/aspnet/BuildTools/commit/84dcc6f0eb5455a3c0305d6d238926defb050889 back to release/2.1 to resolve NU5123

Resolves https://github.com/aspnet/AspNetCore-Internal/issues/1477


jkotalik:
Funny how that was giving long path issues 😄

PR #903, opened 2018-12-18
Fix powershell

ryanbrandenburg:
Some versions of powershell don't work with `& $vswherePath @vswhereArgs | ConvertFrom-Json`, but they do work with `& $vswherePath @vswhereArgs > ConvertFrom-Json`.

### DIFF
--- a/files/KoreBuild/scripts/KoreBuild.psm1
+++ b/files/KoreBuild/scripts/KoreBuild.psm1
@@ -125,7 +125,7 @@ function Invoke-RepositoryBuild(
         }
         else {
             [string[]]$repoTasksArgs = $MSBuildArgs | Where-Object { ($_ -like '-p:*') -or ($_ -like '/p:*') -or ($_ -like '-property:') -or ($_ -like '/property:') }
-            $repoTasksArgs += ,"@$msBuildLogRspFile"
+            $repoTasksArgs += , "@$msBuildLogRspFile"
             __build_task_project $Path $repoTasksArgs
         }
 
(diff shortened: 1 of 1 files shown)


jkotalik:
I'm glad this works, but this script will be deprecated soon once we move off korebuild. I added: https://github.com/aspnet/AspNetCore/blob/master/scripts/install_vs.ps1 as an alternative once we can't use this.

Also, why isn't your powershell version high enough? For a dev experience, why are you blocked on this?

ryanbrandenburg:
PS 4.0 is installed on our Win2008 CI machine and a broken version is also on our Win7 machine. I'm fine with any solution that [un-breaks our tests](http://aspnetci/viewType.html?buildTypeId=XPlat_Windows_Win7_Universe&tab=buildTypeHistoryList&branch_XPlat=__all_branches__). Is there a line than we can change that will install the proper version of Powershell?

jkotalik:
I'm confused about why this wasn't failing before then.

natemcmaster:
https://github.com/aspnet/BuildTools/blob/0872134136bf3d181e3ca3fa700b7ff392c055b3/scripts/bootstrapper/run.ps1#L2

This should have handled it.

natemcmaster:
I'd be happy to meet a PR with that fix.

poke:
@natemcmaster Oh sure, I was thinking that you didn’t want to change anything here anymore since you’re moving off KoreBuild anyway. I’ll send a PR in a moment.

PR #917, opened 2019-01-08
Support building projects which have custom implementations of 'Restore' and 'Test' targets

natemcmaster:
Support building projects which don't fit the vanilla C# template.

Examples:
 - C++
 - Wixproj
 - Custom projects, like shims projects that invoke another build system like npm


natemcmaster:
I'll add comments to clarify.

PR #927, opened 2019-01-18
Don't code-sign .symbols.nupkg files by default and allow signcheck to run on multiple directories

natemcmaster:


natemcmaster:
More preparation for moving aspnet/AspNetCore to Arcade. This will help me migrate the way we code sign to be closer to the way Arcade does it.

cc @ryanbrandenburg

modules/KoreBuild.Tasks/CodeSign.props
```diff
@@ -0,0 +1,29 @@
...
+    <SignCheckWorkingDir Condition=" '$(SignCheckWorkingDir)' == '' ">$(ArtifactsDir)</SignCheckWorkingDir>
+
+    <!-- Relative paths in SignToolData.json are relative to this path -->
+    <SignToolDataWorkingDir Condition=" '$(SignToolDataWorkingDir)' == '' ">$(RepositoryRoot)</SignToolDataWorkingDir>
+
+    <!-- Dry run checks signing config without code signing. -->
+    <SignToolDryRun Condition=" '$(SignType)' != 'real' AND '$(SignType)' != 'test' ">true</SignToolDryRun>
+    <SignToolDryRun Condition=" '$(SignToolDryRun)' == '' ">false</SignToolDryRun>
+
+    <!-- Test signing (not commonly used) -->
```
dougbu: Please clarify this comment to distinguish between `'%(SignToolDryRun)' == 'true'` and `'%(SignToolDryRun)' == 'true'` (aka `'$(SignType)' == 'test'`)

modules/KoreBuild.Tasks/CodeSign.targets
```diff
@@ -52,7 +30,7 @@
       DryRun="$(SignToolDryRun)"
       TestSign="$(SignToolTestSign)"
       ItemsToSign="@(_ItemsToSign)"
-      FileSignInfo="@(_FileSignInfo)"
+      FileSignInfo="@(_FileSignInfo);@(FileNamesToSign)"
```
dougbu: Where are `@(FileNamesToSign)` items created and why aren't they included in `@(_FileSignInfo)`?

natemcmaster: They are optionally specified by a repo. I'll be using this item group when I cleanup these files: https://github.com/aspnet/AspNetCore/blob/master/build/CodeSign.props, https://github.com/aspnet/AspNetCore/blob/master/build/CodeSign.targets

modules/KoreBuild.Tasks/CodeSign.props
natemcmaster: updated. let me know if it helps.

modules/KoreBuild.Tasks/CodeSign.targets
dougbu: Yup, that helped

modules/KoreBuild.Tasks/CodeSign.props
dougbu: And this helped too

PR #940, opened 2019-03-06
Fix Trx files for teamcity

ryanbrandenburg:


ryanbrandenburg:
We'll have to see it in action on TC before we know for sure, but the TC docs says it should work, and we can't get much more broken than not finding any of our tests, so we've got that going for us, which is nice.

PR #943, opened 2019-03-14
add support for handling flaky tests

anurse:
This configures KoreBuild to skip all tests marked with the relevant `[Flaky]` attribute when `RunFlakyTests` is not set. Similarly, it configures it to *only* run the relevant flaky tests when `RunFlakyTests` **is** set.

Once this is flowing through to AspNetCore, we'll configure the AzP yaml to do two test passes, one without `RunFlakyTest` that we care about and one with `RunFlakyTests` that we ignore failures in.

Separate TRX files are generated for each run and we'll still publish all the failing tests. If RaaS is configured to "triage" successful builds, it'll be able to see these failing flaky tests and update Issues as needed.

### DIFF
--- a/files/KoreBuild/modules/vstest/module.targets
+++ b/files/KoreBuild/modules/vstest/module.targets
@@ -94,9 +94,12 @@ Runs the VSTest on all projects in the ProjectToBuild itemgroup.
   <PropertyGroup>
     <TrxFileWithoutExtension>$(LogOutputDir)$(TestGroupName)-$(TargetFramework)-$(BuildNumber)</TrxFileWithoutExtension>
     <TrxFile>$(TrxFileWithoutExtension).trx</TrxFile>
+    <TrxFile Condition="'$(RunFlakyTests)' == 'true'">$(TrxFileWithoutExtension).flaky.trx</TrxFile>
     <TrxFileWildcard>$(TrxFileWithoutExtension)*.trx</TrxFileWildcard>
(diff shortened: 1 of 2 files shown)


files/KoreBuild/modules/vstest/module.targets
```diff
@@ -111,6 +114,12 @@ Runs the VSTest on all projects in the ProjectToBuild itemgroup.
 
     <PropertyGroup>
       <TestAdapterPath>%(TestAssemblies.RootDir)%(TestAssemblies.Directory)</TestAdapterPath>
+
+      <_TestFilter Condition="'$(RunFlakyTests)' != 'true'">Flaky:All!=true</_TestFilter>
+      <_TestFilter Condition="'$(RunFlakyTests)' == 'true'">Flaky:All=true</_TestFilter>
+
+      <_TestFilter Condition="'$(RunFlakyTests)' != 'true' And '$(AGENT_OS)' != ''">$(_TestFilter)&amp;Flaky:AzP:All!=true&amp;Flaky:AzP:OS:$(AGENT_OS)!=true</_TestFilter>
+      <_TestFilter Condition="'$(RunFlakyTests)' == 'true' And '$(AGENT_OS)' != ''">$(_TestFilter)|Flaky:AzP:All=true|Flaky:AzP:OS:$(AGENT_OS)=true</_TestFilter>
```
anurse: I'm going to refactor this slightly to ensure that a repo can configure it's own Flaky/Non-Flaky filter logic in `repo.props`, but that won't affect the bulk of this change so please give feedback on that :)

anurse: This is done now.

PR #942, opened 2019-03-07
Update sdk.version

sebastienros:

PR #944, opened 2019-03-14
Revert "Update sdk.version (#942)"

pakrym:
This reverts commit 321c5b9f50bd787e4cf2c66f6103d177b5d769aa.


dougbu:
Mostly curious: What's the background on upping then downing the SDK version?

pakrym:
We are not able to take new SDK into AspNetCore because of some regressions but we want to take other updates from buildtools

dougbu:
Thanks

PR #947, opened 2019-03-22
Update SDK

jkotalik:
Round 15 or so with this. A roslyn fix was merged and someone needs to get the ball rolling. Once this is merged, I'll try verifying on aspnetcore.


jkotalik:
I doubt it will work, but unfortunately we can't verify without updating build tools. I'll revert if it doesn't work.

PR #956, opened 2019-04-12
Add support for custom restore targets

javiercn:
* Allows projects to define custom targets that will run
  during the repository restore phase, for example; at the same time we restore npmproj files.
* These custom restore targets will run sequentially, so as to avoid issues with parallelism in the most common case (npm/yarn).
* Projects will define the custom restore targets to run by providing a target like the one below.
  ```
  <Target
    Name="_RequiresCustomRestore"
    Returns="@(CustomRestoreTargets)">

    <ItemGroup>
      <CustomRestoreTargets Include="$(MSbuildThisFile)">
        <Targets>YarnInstall</Targets>
      </CustomRestoreTargets>
    </ItemGroup>
  </Target>
  ```

### DIFF
--- a/files/KoreBuild/modules/projectbuild/module.targets
+++ b/files/KoreBuild/modules/projectbuild/module.targets
@@ -75,6 +75,19 @@ Executes /t:{Target} on all projects
           ItemName="_ProjectToRestoreWithNuGet" />
     </MSBuild>
 
+    <MSBuild Targets="_RequiresCustomRestore"
+             Projects="@(ProjectToBuild)"
+             BuildInParallel="true"
+             SkipNonexistentTargets="true"
+             SkipNonexistentProjects="true"
+             RemoveProperties="$(_BuildPropertiesToRemove);Platform"
+             Properties="$(BuildProperties);__BuildTarget=_RequiresCustomRestore">
+      <Output
+          TaskParameter="TargetOutputs"
+          ItemName="_ProjectWithCustomRestore" />
+    </MSBuild>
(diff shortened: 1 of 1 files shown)


files/KoreBuild/modules/projectbuild/module.targets
```diff
@@ -103,6 +125,7 @@ Executes /t:{Target} on all projects
              BuildInParallel="$(BuildInParallel)" />
   </Target>
 
+
```
ryanbrandenburg: Extra newline.

files/KoreBuild/modules/projectbuild/module.targets
```diff
@@ -75,6 +75,19 @@ Executes /t:{Target} on all projects
           ItemName="_ProjectToRestoreWithNuGet" />
     </MSBuild>
 
+    <MSBuild Targets="_RequiresCustomRestore"
```
natemcmaster: For naming consistency with how NuGet implemented this, could we name this target `_IsCustomRestoreTargetSupported`?

dougbu:
I'm unsure about taking this because it increases our dependence on BuildTools as we're deprecating it. The changes in aspnet/AspNetCore#9323 also seem to break aspnet/AspNetCore-Internal#2015 (which you fixed recently but haven't closed BTW).

What problem does moving the yarn restore step into our restore target solve?

javiercn:
@dougbu No concurrrent restores. By the time they happen they are noop

Issue #747, opened 2018-08-29
Convert from ApiCheck to ApiCompat


natemcmaster:
As a part of Arcade convergence, we plan to drop our ApiCheck tool and convert to ApiCompat, which is the tool other teams use to ensure public API remains compatible between releases.

PR #961, opened 2019-04-30
Remove API check

natemcmaster:
This tool has been made obsolete and will not be supported going forward. The recommended replacement is Microsoft.DotNet.ApiCompat (see https://github.com/dotnet/arcade).

Part of https://github.com/aspnet/BuildTools/issues/747


src/Internal.AspNetCore.Sdk/sdk/Sdk.props
```diff
@@ -3,20 +3,8 @@
...
     <_ReferencesInternalAspNetCoreSdk>true</_ReferencesInternalAspNetCoreSdk>
-
-    <!-- Suppress the message about using a preview version of the .NET Core SDK -->
-    <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
-
-    <!-- Make error messages clickable in VS Code's console -->
-    <GenerateFullPaths Condition="'$(VSCODE_PID)' != ''">true</GenerateFullPaths>
-
-    <!-- Default to setting warnings as errors -->
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
dougbu: Does this file pick up settings from files/KoreBuild/msbuild/KoreBuild.RepoTasks.Sdk/Sdk/Sdk.props or files/KoreBuild/KoreBuild.Common.props? If not, why are these no longer needed?

build/tasks/RepoTasks.csproj
```diff
@@ -1,5 +1,5 @@
-<Project>
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), version.props))\build\sdk.props" />
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$(RepositoryRoot)files\KoreBuild\msbuild\KoreBuild.RepoTasks.Sdk\Sdk\Sdk.props" />
```
dougbu: Can build/sdk.* be removed?

src/Internal.AspNetCore.Sdk/sdk/Sdk.targets
```diff
@@ -9,6 +9,4 @@
   <Import Project="..\buildMultiTargeting\Internal.AspNetCore.Sdk.targets" Condition="'$(TargetFrameworks)' != '' AND '$(TargetFramework)' == ''" />
   <Import Project="..\build\Internal.AspNetCore.Sdk.targets" Condition="'$(TargetFramework)' != ''" />
 
-  <Import Project="DefaultItems.targets" />
```
dougbu: Will we update RepoTasks.csproj in AspNetCore i.e. add the lost Microsoft.Build.* references when moving to the new BuildTools?

build/tasks/RepoTasks.csproj
natemcmaster: Not quite yet. I started doing that, but it required more changes to this repo than fit within the scope of "remove API check"

src/Internal.AspNetCore.Sdk/sdk/Sdk.targets
natemcmaster: I don't believe any changes will be needed. This code is obsolete in the 3.0 branch. Nothing calls into this targets file because we stopped using Internal.AspNetCore.Sdk as an "SDK".

src/Internal.AspNetCore.Sdk/sdk/Sdk.props
natemcmaster: No, it doesn't import from there. This can be removed because we stopped using Internal.AspNetCore.Sdk as an "SDK". This was originally added so, when used as an 'SDK', Internal.AspNetCore.Sdk could add implicit references to API check and other dependencies.

PR #963, opened 2019-05-01
[2.2] Update SDK version

JunTaoLuo:
This should update the nuget client to a version that will mitigate aspnet/AspNetCore-Internal#2279


JunTaoLuo:
Good bot :D

PR #964, opened 2019-05-01
[2.1] Update SDK version

JunTaoLuo:
This should update the nuget client to a version that will mitigate aspnet/AspNetCore-Internal#2279


files/KoreBuild/config/sdk.version
```diff
@@ -1 +1 @@
-2.1.503
+2.1.504
```
natemcmaster:
```suggestion
2.1.506
```

Might as well get the latest :)

PR #976, opened 2019-05-30
Remove Internal.AspNetCore.Sdk and unused build tasks

natemcmaster:
Part of https://github.com/aspnet/AspNetCore/issues/7280

This deletes Internal.AspNetCore.Sdk completely. We recommend anyone still using this convert to using Microsoft.DotNet.Arcade.Sdk (https://github.com/dotnet/arcade)

This change introduces a temporary package, Internal.AspNetCore.BuildTasks, which is necessary because there is one task still in use by aspnet/AspNetCore that doesn't have an adequate replacement in Arcade or MSBuild: the GenerateFileFromTemplate task.

Removing this altogether is tracked by https://github.com/dotnet/arcade/issues/204


JunTaoLuo:
:shipit: when tests pass. Curious if you tried it locally already?

natemcmaster:
> Curious if you tried it locally already?

Yup. Built locally and validated this will work. A reaction to this will be part of a PR I'm preparing to open soon on aspnet/AspNetCore which will replace Internal.AspNetCore.Sdk with Microsoft.DotNet.Arcade.Sdk.

PR #1000, opened 2021-04-07
Fix tests

akoeplinger:


al-cheb (approved):

dougbu:
Thanks very much @akoeplinger but I'm going to correct the locale instead. That'll be a single fix I can apply in all of the affected repos consistently. In addition, the current locale on the Ubuntu 18.04 agents isn't great for .NET Core 2.1.

PR #1011, opened 2022-04-22
[release/2.1] Fix build break

dougbu:
- correct `demand` for internal builds


dougbu:
Unbreak the build. Merging immediately since CI validation won't check this change and internal build might as well be for real

